PR #46, opened 2016-07-29
Npm registry internal

dborysiewicz:
Really not sure if this follows your pattern as it means to publish/deploy a new version you will have to be on the internal network. 

Alternatively we can remove it from the pre & post version hooks but it means it will have to be done as a manual step afterward npm version?


mjt01:
We can close this PR now the repo is public. @dborysiewicz happy?

PR #48, opened 2016-09-19
Darken footer and add accessible link colour

mjt01:
Makes the footer darker to add room for a gray panel that fits between our background colour and the footer colour. As a result of the darker colour we also need an accessible variant of our link blue.

Includes updates to the style guide app for layout of the Colour Palette preview.

### DIFF
--- a/theme/palette/accessible-variants.less
+++ b/theme/palette/accessible-variants.less
@@ -3,3 +3,4 @@
 @sk-pink-on-gray-light: #d4006b;
 @sk-learning-dark-on-gray-light: #157e00;
 @sk-business-on-gray-light: #0f749f;
+@sk-link-on-mid-gray-light: #005bb6;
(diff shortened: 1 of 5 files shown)


tcrook1:
Thanks Mike. Only point I’d add is that the need for a variant link blue existed prior to the footer bg colour update (as #0066CC was not accessible on #E1E1E1).

theme/palette/grays.less
```diff
@@ -5,6 +5,7 @@
 @sk-mid-gray-medium: #898989;
 @sk-mid-gray: #a8a8a8;
 @sk-mid-gray-light: #d6d6d6;
+@sk-gray-medium: #e1e1e1;
```
markdalgleish: Woo, another shade! :trollface:

theme/palette/accessible-variants.less
```diff
@@ -3,3 +3,4 @@
 @sk-pink-on-gray-light: #d4006b;
 @sk-learning-dark-on-gray-light: #157e00;
 @sk-business-on-gray-light: #0f749f;
+@sk-link-on-mid-gray-light: #005bb6;
```
markdalgleish: 👏

PR #49, opened 2016-10-02
Compile CSS classes to emoji

markdalgleish:
To avoid clashing with emoji classes being used by the consuming project, and to signify that it came from the style guide, we prefix all classes with a colour palette icon (🎨)


mjt01:
:art:

PR #52, opened 2016-10-05
Revert "Compile CSS classes to emoji"

mjt01:
Due to the non-deterministic order of emoji during webpack compilation, this results in mismatches of class names between server and client which to the user is a flash of unstyled content for server rendered apps.

Reverting until we dig a little deeper here.

/cc @markdalgleish


markdalgleish:
😭 🚀

PR #53, opened 2016-10-12
Added autosuggest field to style guide

mattsjones:


react/fields/Autosuggest/Autosuggest.less
```diff
@@ -0,0 +1,55 @@
...
+
+.suggestion {
+  .rawText();
+
+  line-height: (@grid-row-height * 4);
+  height: (@grid-row-height * 4);
+  padding: 0 (@grid-gutter-width * 0.75);
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
mjt01: only difference we have in chalice is `cursor: pointer;`, feel like either we should remove that or add it in here.

@tcrook1 @kaylajheffernan should we have the pointer (👆)  when hovering over suggestions in the auto suggest?

react/fields/Autosuggest/Autosuggest.less
```diff
@@ -0,0 +1,55 @@
...
+}
+
+.suggestionFocused {
+  background-color: @sk-gray-light;
```
mjt01: @tcrook1 also our focus/hover colour for suggestion differs, should it be `@sk-mid-gray-light` (#d6d6d6) or `@sk-gray-light` (#eeeeee)

tcrook1:
@mjt01 @mattsjones  Hover should be @sk-mid-gray-light (#d6d6d6), as we're using the @sk-gray-light (#eeeeee) to group elements within drop-downs. Hover needs to be darker than the grouping colour behind it. (Example is the logged-in user drop down, and the grouping of Saved Searches, Saved Jobs and Applied Jobs. This treatment is also planned for the visual update of the Classification drop-down). I think we should have the pointer on hover also. Thanks Matt and Mike.

mjt01:
The browser clear icon does not hide because we scope our override to `<input type="search">` ([see here](https://github.com/SEEK-Jobs/seek-style-guide/blob/master/react/fields/TextField/TextField.less#L46). Feel like this should be for inputs as we have custom clear icons.

app/components/Autosuggest/Autosuggest.js
```diff
@@ -0,0 +1,204 @@
...
+};
+
+function getSpec(specsObj) {
+  return Object.keys(specsObj).reduce((result, specName) => ({
+    ...result,
+    ...(specsObj[specName] ? specs[specName] : {})
+  }), {});
+}
+
+export default class TextFields extends Component {
```
mjt01: missed updating this component name

app/components/Autosuggest/Autosuggest.js
```diff
@@ -0,0 +1,204 @@
...
+    const className = classnames({
+      [textFieldStyles.rootFocus]: focus
+    });
+    const spec = getSpec({
+      default: true,
+      focus,
+      invalid
+    });
+    const autosuggestProps = {
+      suggestions: ['Developer'],
```
mjt01: maybe add a handful of roles in here, makes a better demo.

react/fields/Autosuggest/Autosuggest.less
```diff
@@ -0,0 +1,56 @@
...
+    padding: @grid-row-height 0;
+    border-radius: 2px;
+    min-width: 100%;
+    max-width: 100%;
+    max-height: 385px;
+    overflow-y: auto;
+    top: @grid-row-height * 6;
+
+    @media only screen and (min-width: 920px) {
+      max-width: initial;
```
mjt01: This rule looks like it's out of chalice, not sure the `920px` breakpoint makes sense here.

react/fields/Autosuggest/Autosuggest.less
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+}
+
+.sectionTitle {
```
mjt01: As per @tcrook1's comment in slack, let's update this rule to:
```
.sectionTitle {
  .rawText(1);
  line-height: (@grid-row-height * 4);
  height: (@grid-row-height * 4);
  margin: @grid-row-height (@grid-gutter-width * 0.75) 0;
  color: darken(@sk-highlight, 10%);
  font-weight: @sk-bold;

  .sectionContainer:first-child & {
    margin-top: 0;
  }
}
```

react/fields/Autosuggest/Autosuggest.less
mattsjones: 🤔 Maybe we should make it @grid-container-width? Or remove entirely?

mjt01: let's start with removing the media query and we'll deal with exceptions as they arise.

PR #55, opened 2016-10-20
Allow mutliple ref assignment in textfield to fix autosuggest behaviour

mattsjones:
- Also add route for github pages.

### DIFF
--- a/react/fields/Autosuggest/Autosuggest.test.js
+++ b/react/fields/Autosuggest/Autosuggest.test.js
@@ -34,12 +34,6 @@ const getAutosuggestProps = (suggestions = []) => ({
 describe('Autosuggest', () => {
   let element, autosuggest, errors, suggestions, suggestionsContainer, input;
 
-  const clickSuggestion = suggestion => {
-    Simulate.mouseDown(suggestion);
-    Simulate.mouseUp(suggestion);
-    Simulate.click(suggestion);
-  };
-
   beforeEach(() => {
     errors = [];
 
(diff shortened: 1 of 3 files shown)


react/fields/Autosuggest/Autosuggest.test.js
```diff
@@ -76,6 +85,11 @@ describe('Autosuggest', () => {
     expect(suggestionsContainer.className).to.not.contain('suggestionsContainer_withLabel');
   });
 
+  it('should render suggestions list below input field when a label is not supplied', () => {
```
mattsjones: This is weird, Github is lying. This code is already in master.

react/fields/Autosuggest/Autosuggest.test.js
```diff
@@ -30,7 +32,13 @@ const getAutosuggestProps = (suggestions = []) => ({
 });
 
 describe('Autosuggest', () => {
-  let element, autosuggest, errors, suggestion, suggestionsContainer;
+  let element, autosuggest, errors, suggestions, suggestionsContainer, input;
+
+  const clickSuggestion = suggestion => {
+    Simulate.mouseDown(suggestion);
+    Simulate.mouseUp(suggestion);
+    Simulate.click(suggestion);
```
mjt01: Did you need to do this? This complexity was only added for `ClearField` interaction, you should be able to simply use `Simulate.click`?

react/fields/Autosuggest/Autosuggest.test.js
mjt01: This test is duplicated now! Check line 83

mattsjones: Ahh... right

react/fields/Autosuggest/Autosuggest.test.js
mattsjones: I'll double check

mjt01:
:art:

PR #56, opened 2016-10-20
Fields subnavigation

mattsjones:


mjt01:
You forgot to update the static build routes in webpack.gh-pages.server.config.js

mattsjones:
This is what it looks like btw. 

image

mjt01:
Given where it's landed, i feel like this should be updated to: Home, Typography, Components — where Components is a list of Button, Icons, AutoSuggest, TextField etc

markdalgleish:
@mattsjones What's the status of this PR?

markdalgleish:
@mjt01 is this something you're still across, or do we need to wait for Matt?

mjt01:
@markdalgleish this should probably be closed, since then a lot of new components have been added. We need a better solution for the IA of the mini site, not just for fields. Arguably a better starting point would be:

"Home" | "Typography" | "Colours" | "Components"

Where "Components" is a drop down menu. The site is pretty broken right now. Thoughts?

markdalgleish:
@mjt01 agreed that this is a much larger problem that needs a more holistic solution.

mengtzu:
@mjt01 +1 to that suggestion.  A drop-down might become unwieldy with enough components (or even categories of components), but we could deal with that when actually reaching that scale.

markdalgleish:
Spoke with @mattsjones, closing this in favour or more large-scale changes which will hopefully be coming soon.

PR #57, opened 2016-10-20
Improved Code section formatting and added horizontal scrolling

mattsjones:
Turns this: 
image

Into this:
image

### DIFF
--- a/app/components/Code/Code.less
+++ b/app/components/Code/Code.less
@@ -15,7 +15,7 @@
   font-family: Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace;
   overflow-x: auto;
   overflow-y: hidden;
-  padding: 0 (@grid-gutter-width * 2);
+  padding: 0 (@grid-gutter-width / 2);
   &::-webkit-scrollbar {
     width: 0px;
     background: transparent;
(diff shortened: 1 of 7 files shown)


app/components/Code/Code.less
```diff
@@ -3,14 +3,23 @@
 .root {
   .rawText(@standard-type-scale);
   position: relative;
-  padding: (@grid-row-height * 2) (@grid-gutter-width * 2);
+  padding: (@grid-row-height * 2) 0;
```
mjt01: This will render hard up against the blue border on the left. Maybe more suited to:
```
padding: (@grid-row-height * 2) 0 (@grid-row-height * 2) (@grid-gutter-width / 2)
```
screen shot 2016-10-21 at 9 06 43 am
Thoughts?

mjt01: Ah I see you moved the padding to `.card`. Still, I think the 2 gutter spacing is too much, cuts off more code than it needs to. See [next comment](https://github.com/SEEK-Jobs/seek-style-guide/pull/57/files#r84386417)

mjt01:
Other than squeezing more horizontal space out of `Code` block, lgtm :art:

PR #59, opened 2016-10-25
Fix peer dependency requirements

mjt01:
Any module imported by a runtime module/component (e.g. everything in the `react` directory) should be specified as a `peerDependency` in `package.json`. The style guide is consumed as a `devDependency` and the required modules are bundled as part of the consuming projects build artefact. This means and `dependencies` in this project are treated as `devDependencies` and dont end up in the bundle 😢 

There's room to discuss how the style guide handles dependencies, but this is at least a shorter term clean up.

/cc @mattsjones


mjt01:
Should mention I identified the peerDependency packages by looking through all `import` statements in the `react` directory 🕵️

mattsjones:
Nice job mate. Good sorting this out. Think this approach is fine, at least for the mean time. 🎨

PR #60, opened 2016-10-27
Export the icon component

georgespyropoulos:
I'm working on the Comm games project and we're in the process of creating a few icons for our app that (I'm told) are not appropriate / useful to have in the style guide.

We've started a new style guide project and we are adding all comm games specific styling to the project.  This project extends the seek style guide.

It'd be great for us to be able to use your `Icon` component as a base for building on top of as we create new icons for our project.  Grab me on slack (@george) if you'd like to discuss this further.


georgespyropoulos:
I've discussed this with @mjt01 and we have decided that it'd be better to not export this component as it could encourage people to create icon sets outside of the styleguide rather than contribute to it.  Closing this PR

PR #61, opened 2016-10-27
Dropdown

mattsjones:
This is preceding PR for the month picker. Dropdown is not exported publically and is not linked to  in head as it may require some more visual work before it is used for more common purposes. However, the component is in a fully working and tested state with support for all basic style guide field elements (invalid, message, label, etc).


mjt01:
You haven't included the update for [static pages config](https://github.com/SEEK-Jobs/seek-style-guide/blob/master/webpack.gh-pages.server.config.js)

app/components/Dropdown/Dropdown.js
```diff
@@ -0,0 +1,205 @@
...
+        label: 'Product Manager'
+      }
+    ];
+    const dropdown = (
+      <SeekDropDown
+        id="jobTitles"
+        className={className}
+        invalid={invalid}
+        label="Job Titles"
+        focus={focus}
```
mjt01: `focus` does not appear necessary to pass as a prop (also it's not validated or consumed internally to `Dropdown`)

react/fields/Dropdown/Dropdown.js
```diff
@@ -0,0 +1,239 @@
...
+
+    return (
+      <label {...allLabelProps}>
+        {label}
+      </label>
+    );
+  }
+
+  renderSelect() {
+    const { inputProps = {}, id, options = [], placeholder } = this.props;
```
mjt01: `options` is a required prop so defaulting it here seems like it's redundant.

react/fields/Dropdown/Dropdown.js
```diff
@@ -0,0 +1,239 @@
...
+      <select {...allInputProps}>
+        <option
+          className={styles.blankOption}
+          value=""
+          disabled={true}>
+          { placeholder }
+        </option>
+        {
+          options.map(({ value, label }) => (
+            <option value={value} key={value}>{ label }</option>
```
mjt01: you may want to use the a combination of the value and index for the `key` as the value can be blank

react/fields/Dropdown/Dropdown.less
```diff
@@ -0,0 +1,111 @@
...
+    color: @sk-mid-gray;
+  }
+
+  &:first-child {
+    margin-right: @grid-gutter-width / 2;
+  }
+
+  &:last-child {
+    margin-bottom: (@grid-row-height * 3);
+  }
```
mjt01: these `first-child` and `last-child` rules look like implementation detail rather than `Dropdown` styles, is that right?

react/fields/Dropdown/Dropdown.js
```diff
@@ -0,0 +1,239 @@
...
+      [styles.root]: true,
+      [styles.invalid]: invalid,
+      [className]: className
+    });
+
+    return (
+      <div className={classNames}>
+        {this.renderLabel()}
+        {this.renderChevron()}
+        {this.renderSelect()}
```
mjt01: Just confirming that rendering the select after the chevron makes it clickable in IE 10 as `pointer-events:none` wont be recognised. Worth double checking.

react/fields/Dropdown/Dropdown.less
```diff
@@ -0,0 +1,111 @@
...
+    box-shadow: 0 0 0 @field-border-width @sk-pink;
+
+    &:not([readonly]):not([disabled]):focus,
+    .rootFocus & {
+      border-color: transparent;
+      box-shadow: 0 0 0 @field-border-width @sk-pink;
+    }
+  }
+
+  &.blankValueSelected {
```
mjt01: does `placeholder` make more sense?

mjt01:
Would be nice to have a test for `placeholder` if you get a chance

react/fields/Dropdown/Dropdown.js
mattsjones: Pretty sure it won't be clickable in IE10 but honestly my care factor is pretty low. Couldn't think of a nice solution for this. Thoughts?

react/fields/Dropdown/Dropdown.less
mattsjones: 👍

react/fields/Dropdown/Dropdown.js
mattsjones: This is interesting as a value of empty string is used in the placeholder option so that would cause bigger problems then react warnings. Maybe I should add custom prop validation for the options array?

react/fields/Dropdown/Dropdown.js
```diff
@@ -0,0 +1,239 @@
...
+    const allInputProps = {
+      ...combineClassNames(inputProps, inputStyles),
+      ...(id ? { id } : {}),
+      ref: this.storeInputReference,
+      value: inputProps.value || ''
+    };
+
+    return (
+      <select {...allInputProps}>
+        <option
```
mattsjones: Discussed this with Steph. We're going to always render as otherwise, the first value from your options becomes the default value.

react/fields/Dropdown/Dropdown.less
mattsjones: `first-child` is my bad, forgot to remove from something I was trying. `last-child` however supplies the margin-bottom for when there is no message or help text.

app/components/Dropdown/Dropdown.js
mattsjones: Nice pickup

mattsjones:
@mjt01 Resolved or responded to all your comments. Also cleaned up a couple of styling things. `inputProps` is now required along with a `value`.

react/fields/Dropdown/Dropdown.js
mjt01: Everything in the style guide should be compatible in my mind. I would encourage to check this before merging.

mjt01:
Im happy apart from the last comment, everything in the style guide should work in supported browsers. I would encourage you to test this in IE10 please. Other than that :art:

react/fields/Dropdown/Dropdown.js
mattsjones: IE10 needed a few style tweaks but interactions were good 👍

PR #62, opened 2016-11-03
Add Checkbox

shawn-dsz:


react/fields/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,77 @@
...
+      checked: PropTypes.bool
+    })
+  }
+
+  static defaultProps = {
+    inputProps: {}
+  }
+
+  renderLabel() {
+    const { inputProps: { checked = false }, label, id } = this.props;
```
mattsjones: You don't need to default `checked` to false as it will be falsy if it doesn't exist anyway. However, this expression will throw if `inputProps` is not supplied. I think we should make `inputProps` required and potentially `checked` as well.

react/fields/Checkbox/Checkbox.less
```diff
@@ -0,0 +1,47 @@
...
+.checkMark {
+  display: block;
+  width: initial;
+  height: initial;
+  fill: @sk-pink;
+  opacity: 0;
+}
+
+.checked {
+  opacity: 1;
```
mattsjones: Instead of adding an extra class here you use the `:checked` selector. 
https://css-tricks.com/almanac/selectors/c/checked/

react/fields/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,68 @@
...
+  };
+}
+
+export default class Checkbox extends Component {
+  static displayName = 'Checkbox';
+
+  static propTypes = {
+    id: PropTypes.string.isRequired,
+    label: PropTypes.string.isRequired,
+    className: PropTypes.string,
```
mjt01: For props that are not required can you please include a `defaultProp`, makes the component's api more self-documenting

react/fields/Checkbox/Checkbox.less
```diff
@@ -0,0 +1,45 @@
...
+.label {
+  .touchableText(@standard-type-scale);
+  cursor: pointer;
+  display: flex;
+  align-items: center;
+}
+
+.checkBox {
+  width: @grid-row-height * 3;
+  height: @grid-row-height * 3;
```
mjt01: should the height/width take into account the border size so that the component stays on grid?

react/fields/Checkbox/Checkbox.less
```diff
@@ -0,0 +1,45 @@
...
+    outline: none;
+    border-color: transparent;
+    .focus();
+  }
+}
+
+.checkMark {
+  display: block;
+  width: initial;
+  height: initial;
```
mjt01: what are these values overriding?

react/icons/CheckMarkIcon/CheckMarkIcon.svg
```diff
@@ -0,0 +1 @@
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="90.471 302.809 823 823"><g></g><g> <g><polygon points="365.407,1022.185 90.677,747.457 158.56,679.575 365.406,886.421 845.383,406.432 913.266,474.314"/></g></g> </svg>
```
mjt01: Has this asset been optimised? Seems to have some redundant group elements. @stephaniemoss has previously tidied up assets for me before adding. Not a big deal, just noticed it

mjt01:
Couple of tidy up comments, but looking really good!

react/fields/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,68 @@
...
+    inputProps: PropTypes.shape({
+      onChange: PropTypes.func,
+      checked: PropTypes.bool
```
mattsjones: Personally, I'm in favour of forcing required if it makes the component logic easier. I don't see any use for these fields in a non-controlled scenario anyway.

react/fields/Checkbox/Checkbox.less
shawn-dsz: good point, I'll minus the border widths

react/fields/Checkbox/Checkbox.less
shawn-dsz: This is to override the icon width and height [set by the Icon component](https://github.com/SEEK-Jobs/seek-style-guide/blob/master/react/icons/icon.less#L9-L10). I would like the icon to fill up the parent checkbox, setting it to ```initial``` will do that without magic numbers.

react/fields/Checkbox/Checkbox.js
shawn-dsz: True, I don't expect there to be a scenario where you would include a style guide component  and not have full control of its events and values.

PR #63, opened 2016-11-08
Added the Helmet meta to SeekApp

tamoore:
- Also added accessibility component for the `ScreenReaderOnly` h1 (both accessibility and seo implications I imagine).


react/SeekApp/SeekApp.js
```diff
@@ -2,23 +2,46 @@ import styles from './SeekApp.less';
...
+import Helmet from 'react-helmet';
+import ScreenReaderOnly from 'Accessibility/ScreenReaderOnly';
 
-export default function SeekApp({ fullScreen, children }) {
+const defaultPageTitleAU = 'SEEK - Australia\'s no. 1 jobs, employment, career and recruitment site';
+const defaultPageTitleNZ = 'Jobs on SEEK - New Zealand\'s no. 1 Employment, Career and Recruitment site';
+
+const getLocalisedPageTitle = locale => locale === 'AU' ? defaultPageTitleAU : defaultPageTitleNZ;
+
+export default function SeekApp({ fullScreen, children, meta, link, title, locale }) {
```
tamoore: This could just be an object for `Helmet` but I thought making these explicit might be better?

package.json
```diff
@@ -44,6 +44,7 @@
     "less-loader": "2.2.3",
     "postcss-loader": "0.9.1",
     "raw-loader": "0.5.1",
+    "react-helmet": "3.1.0",
```
mjt01: This should be listed as a `peerDependency` for consuming apps to have installed. Also anything required for the demo app (this is) should be a `devDependency` not a `dependency`. Only thing in `dependencies` should be required webpack plugins for the config that gets decorated.

react/SeekApp/SeekApp.js
```diff
@@ -2,23 +2,46 @@ import styles from './SeekApp.less';
...
+
   return (
     <div className={className}>
+      <Helmet
+        title={pageTitle}
+        meta={meta}
+        link={link}
+      />
+
+      <ScreenReaderOnly>
```
mjt01: If we hoist this up to the style guide is should probably be top level in the `react` directory and exported as part of the style guide package. Consuming apps should then refactor away their internal usage of `<ScreenReaderOnly />` in favour of this. We'll cover this in the release notes

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "seek-style-guide",
-  "version": "11.2.0",
+  "version": "11.3.0",
```
mjt01: As with the other modules, let's keep the version bump as a separate PR once this functionality is merged

react/SeekApp/SeekApp.js
```diff
@@ -2,23 +2,46 @@ import styles from './SeekApp.less';
...
+import Helmet from 'react-helmet';
+import ScreenReaderOnly from 'Accessibility/ScreenReaderOnly';
 
-export default function SeekApp({ fullScreen, children }) {
+const defaultPageTitleAU = 'SEEK - Australia\'s no. 1 jobs, employment, career and recruitment site';
+const defaultPageTitleNZ = 'Jobs on SEEK - New Zealand\'s no. 1 Employment, Career and Recruitment site';
+
+const getLocalisedPageTitle = locale => locale === 'AU' ? defaultPageTitleAU : defaultPageTitleNZ;
+
+export default function SeekApp({ fullScreen, children, meta = [], link = [], title, locale = 'AU' }) {
```
mjt01: I think it's cleaner to move these defaults to `defaultProps`, w're already doing this for `fullScreen`

mjt01:
lgtm :art:

PR #64, opened 2016-11-08
Month picker

mattsjones:
The long awaited month picker! 

This PR also updates SVGO to add `focusable="true"` to all SVG elements. This solves a problem in IE where SVG's are focusable unless this attribute is present.


package.json
```diff
@@ -78,10 +80,12 @@
     "less-vars-to-js": "^1.1.1",
     "lodash.debounce": "^4.0.6",
     "lodash.omit": "^4.5.0",
+    "lodash.range": "^3.2.0",
     "mocha": "^2.5.3",
     "mock-css-modules": "^1.0.0",
     "opn": "^4.0.1",
     "opn-cli": "^3.1.0",
+    "pad-left": "^2.1.0",
```
tamoore: LOL 😹

react/fields/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -0,0 +1,168 @@
...
+          this.blurIfNotFocussed(activeElement);
+        });
+      }
+    }
+  }
+
+  render() {
+    const { value, className, invalid, id } = this.props;
+    const { month, year } = value;
+    const monthValue = month ? `${month}` : '';
```
mjt01: for the same reason [as above](https://github.com/SEEK-Jobs/seek-style-guide/pull/64/files#r87528592), `String(month || '')`? Depending on your thoughts on the previous comment

react/fields/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -0,0 +1,168 @@
...
+    const yearValue = year ? `${year}` : '';
+
+    const rootClasses = classnames({
+      [className]: className,
+      [styles.root]: true
```
mjt01: normally i would flip these that the optional class is applied on top of the default styles. in this case the `root` properties would override the custom properties

react/fields/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -0,0 +1,168 @@
...
+import range from 'lodash.range';
+
+import Dropdown from '../../Dropdown/Dropdown';
+
+const getYearOptions = () => {
+  const maxYear = new Date().getFullYear();
+  const minYear = maxYear - 100;
+
+  return range(maxYear, minYear - 1).map(value => {
+    const stringValue = `${value}`;
```
mattsjones: Yeah that's probably better as it's more declarative what's it's intending to do

react/fields/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
mattsjones: 👍

react/fields/MonthPicker/MonthPicker.js
```diff
@@ -0,0 +1,175 @@
...
+  };
+}
+
+export default class MonthPicker extends Component {
+
+  static displayName = 'MonthPicker';
+
+  static propTypes = {
+    /* eslint-disable consistent-return */
+    id: (props, propName, componentName) => {
```
tamoore: I noticed that `Dropdown` uses the same propType validator. I wonder if it's worth extracting these into a shared file that contains common `PropTypes`.

react/fields/MonthPicker/MonthPicker.js
```diff
@@ -0,0 +1,175 @@
...
+      if (label && !id) {
+        return new Error(`When ${componentName} has a \`label\`, it should also have an \`id\`.`);
+      }
+    },
+    className: PropTypes.string,
+    /* eslint-enable consistent-return */
+    invalid: PropTypes.bool,
+    label: PropTypes.string,
+    /* eslint-disable consistent-return */
+    labelProps: (props, propName, componentName) => {
```
tamoore: This is also in `Dropdown`

react/fields/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -0,0 +1,168 @@
...
+
+  static propTypes = {
+    onChange: PropTypes.func,
+    onBlur: PropTypes.func,
+    value: PropTypes.shape({
+      month: PropTypes.number,
+      year: PropTypes.number
+    }),
+    invalid: PropTypes.bool,
+    className: PropTypes.string,
```
mjt01: missing default prop for this

react/fields/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -0,0 +1,168 @@
...
+
+  static displayName = 'CustomMonthPicker';
+
+  static propTypes = {
+    onChange: PropTypes.func,
+    onBlur: PropTypes.func,
+    value: PropTypes.shape({
+      month: PropTypes.number,
+      year: PropTypes.number
+    }),
```
mattsjones: No

react/fields/MonthPicker/MonthPicker.js
```diff
@@ -0,0 +1,175 @@
...
+      ...(id ? { id } : {}),
+      value,
+      onChange,
+      onBlur,
+      invalid
+    };
+
+    return (
+      native ?
+        <NativeMonthPicker {...monthPickerProps} /> :
```
tamoore: I got a little confused by the `Native` name, for some reason I thought that meant `ReactNative`. Take from that what you will 😉

react/fields/MonthPicker/MonthPicker.test.js
```diff
@@ -0,0 +1,175 @@
...
+  createRenderer
+} from 'react-addons-test-utils';
+import { findAllWithClass } from 'react-shallow-testutils';
+import MonthPicker from './MonthPicker';
+
+chai.use(sinonChai);
+
+const renderer = createRenderer();
+
+describe('MonthPicker', () => {
```
mjt01: This test suite doesnt test anything about the value of the monthpicker being set. Any thoughts on getting coverage of this?

react/fields/MonthPicker/MonthPicker.js
```diff
@@ -0,0 +1,175 @@
...
+function combineClassNames(props = {}, ...classNames) {
+  const { className, ...restProps } = props;
+
+  return {
+    className: classnames.apply(null, [...classNames, className]), // eslint-disable-line no-useless-call
+    ...restProps
+  };
+}
+
+export default class MonthPicker extends Component {
```
tamoore: Having looked at this component I wonder whether it needs to be a `composite`. It seems to just be a proxy for the `custom` and the `native` component. What are your thoughts? It could be a functional component? Maybe I'm missing something but it looks like it only renders.

react/fields/MonthPicker/NativeMonthPicker/NativeMonthPicker.js
```diff
@@ -0,0 +1,88 @@
...
+
+  render() {
+    const { value, className, invalid, id, onBlur } = this.props;
+
+    const inputValue = makeMonthString(value);
+
+    const rootClasses = classnames({
+      [styles.root]: true,
+      [className]: className,
+      [styles.invalid]: invalid
```
mjt01: again, `className` should be last here (and have a `defaultProp`)

react/fields/MonthPicker/MonthPicker.test.js
mattsjones: 🤔  `MonthPicker` simply passed through a `value` & `onChange` function. All we can really test is that it forwards those props. I've done that for onChange & onBlur so I can add one for value as well.

react/fields/MonthPicker/NativeMonthPicker/NativeMonthPicker.less
```diff
@@ -0,0 +1,59 @@
...
+  border-radius: @field-border-radius;
+  color: @sk-charcoal;
+  background-color: @sk-white;
+
+  &::-webkit-inner-spin-button,
+  &::-webkit-calendar-picker-indicator,
+  &::-webkit-clear-button {
+    display: none;
+    -webkit-appearance: none;
+    -moz-appearance: none;
```
mjt01: is this valid that we are styling webkit pseudo elements with firefox prefix? should we leave this to autoprefixer and make it `appearance: none`?

webpack/index.js
```diff
@@ -92,6 +92,16 @@ const getLocalIdentName = () => isProduction() ?
   '[hash:base64:7]' :
   '__STYLE_GUIDE__[name]__[local]___[hash:base64:7]';
 
+const getSvgoLoader = () => {
+  const svgoConfig = JSON.stringify({
+    plugins: [
+      { addAttributesToSVGElement: { attribute: 'focusable="false"' } }
```
mjt01: 👏

react/fields/MonthPicker/NativeMonthPicker/NativeMonthPicker.less
mattsjones: I'm checking on autoprefixer, I was under the impression appearance was more specific and wouldn't be auto-prefixed. However, I do need to move the `-moz-apearance` out of this selector.

react/fields/MonthPicker/MonthPicker.js
mattsjones: 🤔  Interesting. The only thing is there is a nice consistency between all the fields in style guide atm in the way they're set out. Personally I found this way more readable for a component as large as this (although not that complex). Value in both approaches to me. @mjt01 thoughts?

react/fields/MonthPicker/MonthPicker.js
mattsjones: You're right. There are a lot of duplicates in style guide atm. Also we want to pull label and message (and any other common patterns) into reusable components so we're not duplicating so much. Feel both of these would be better off having a dedicated PR though.

react/fields/MonthPicker/MonthPicker.js
tamoore: I'm terrible at naming! It's like my endless achilles heel and "just when I think I'm out, it pulls me back in".  Soooo....maybe it's better to name it for what the target is current, i.e. mobile devices. So something like, `StandardMobileMonthPicker`. Just a thought.

tamoore: Or `TouchDeviceMonthPicker`

mjt01: Yeah as im trying to think of a better name i lean more to `Native` as it really is cross platform/device etc.

mattsjones: I WIN 🏆

react/fields/MonthPicker/NativeMonthPicker/NativeMonthPicker.js
```diff
@@ -0,0 +1,89 @@
...
+          className={styles.input}
+          type="month"
+          value={inputValue}
+          onChange={this.handleChange}
+          onBlur={onBlur}
+        />
+      </div>
+    );
+  }
+}
```
tamoore: 👍

react/fields/MonthPicker/NativeMonthPicker/NativeMonthPicker.less
```diff
@@ -0,0 +1,59 @@
+@import (reference) "~seek-style-guide/theme";
```
tamoore: I'm not sure if this is mean to be there but I get a strange chevron on iPad

tamoore: screen shot 2016-11-11 at 3 32 43 pm

react/fields/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -0,0 +1,169 @@
...
+    onChange({
+      month,
+      year: parseInt(value, 10)
+    });
+  }
+
+  blurIfNotFocussed(active) {
+    const { onBlur, value } = this.props;
+
+    if (active !== this.monthInput && active !== this.yearInput) {
```
mjt01: `&& typeof onBlur === 'function'`

react/fields/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -0,0 +1,169 @@
...
+      onBlur(value);
+    }
+  }
+
+  handleBlur({ relatedTarget }) {
+    const { onBlur } = this.props;
+
+    if (onBlur) {
```
mjt01: ah gotcha, maybe check its a function too

mattsjones: 👍

react/fields/MonthPicker/NativeMonthPicker/NativeMonthPicker.less
```diff
@@ -16,13 +16,13 @@
   border-radius: @field-border-radius;
   color: @sk-charcoal;
   background-color: @sk-white;
+  appearance: none;
 
   &::-webkit-inner-spin-button,
   &::-webkit-calendar-picker-indicator,
   &::-webkit-clear-button {
     display: none;
     -webkit-appearance: none;
```
mjt01: this is redundant with `appearance: none` being outside this pseudo selector. confirming that it still works as expected if its on the class not the pseudo element?

mattsjones: It actually needed to be moved outside to deal with a separate issue on iOS. This combination of styles should be good now. The `appearance:none` takes care of the normal scenarios and the pseudo selectors deal with the extra webkit specific parts. Both are required though.

mjt01:
:art:

PR #69, opened 2016-11-17
Fix Native Month Picker bugs

mattsjones:
This PR resolves two issue with the native month picker. 

- Creates custom onBlur so correct input value is passed back to application rather than default string representation. 
- Upgrades React peerDependency due to React issue https://github.com/facebook/react/issues/7233 causing prefill to be broken on some mobile devices.


react/fields/MonthPicker/NativeMonthPicker/NativeMonthPicker.js
```diff
@@ -49,6 +49,7 @@ export default class NativeMonthPicker extends Component {
     super();
 
     this.handleChange = this.handleChange.bind(this);
+    this.handleBlur = this.handleBlur.bind(this);
```
tamoore: Assuming the addition of the `blur` event handler is for the resolution of the prefill bug? Do you think it's worth adding a descriptive comment as to why both `blur` and `change` are used? The main reason I say that is that they both have the same implementation.

react/fields/MonthPicker/NativeMonthPicker/NativeMonthPicker.test.js
```diff
@@ -76,9 +76,16 @@ describe('NativeMonthPicker', () => {
     });
   });
 
-  it('should pass onBlur prop through to input', () => {
-    const onBlur = () => {};
-    render(<NativeMonthPicker onBlur={onBlur} />);
-    expect(input.props.onBlur).to.equal(onBlur);
+  it('should send correct month year format in onBlur handler', () => {
```
tamoore: Again, maybe it's worth mentioning why this is needed.

tamoore:
🎨

react/fields/MonthPicker/NativeMonthPicker/NativeMonthPicker.js
tamoore: Ignore me, I just read the PR comment...

mattsjones: It's actually not for the prefill bug. The reason the onBlur was added is because the default onBlur implementation was passing back a string value rather than our custom object. This was playing havoc with our forms library so I added a new onBlur that ensures our custom value.

tamoore: I  realised that about 20 minutes after writing it. <Delayed>

PR #70, opened 2016-12-12
Adds values for responsive types as verified by Scott

mengtzu:
I've verified with Scott that this is the only remaining place where the desired small-screen and large-screen type sizes should differ.  If we can get this in, we can call this our baseline, and discuss adjustments to it case by case going forward, rather than thrashing while trying to re-re-re-confirm our commitment to them as a whole.

As far as I know nobody is consuming these sizes.  This PR should unblock CR from using them (it currently uses Standard but not the others).  From this point on, deviation would be our fault :)

### DIFF
--- a/theme/type/type.less
+++ b/theme/type/type.less
@@ -11,8 +11,8 @@
 // Hierarchy
 @hero-type-row-span: 5;
 @hero-type-scale: 4.2;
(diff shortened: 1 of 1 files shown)


mjt01:
Great! Looks good to me, the missing piece is probably around weights. This will come as we move forward with performant loading of web fonts (hoping to get back to this soon)

:art:

PR #72, opened 2016-12-13
dropdown supports hierarchy

LeviSeek:
this feature would enable the dropdown to display nested information.
optionProps supports hierarchy structure, here is one example of options:
```
const options = [
      {
        label: 'Really really long job title that is not gonna fit',
        value: '3'
      },
      {
        label: 'Major Cities',
        value: [
          {
            label: 'Melbourne',
            value: '3004'
          },
          {
            label: 'Sydney',
            value: '3002'
          }
        ]
      },
      {
        label: 'Balarat',
        value: '3005'
      }
    ]
```
screen shot 2016-12-14 at 11 10 32 am

### DIFF
--- a/app/components/Dropdown/Dropdown.js
+++ b/app/components/Dropdown/Dropdown.js
@@ -121,7 +121,7 @@ export default class DropDownDemo extends Component {
         ]
       },
       {
-        label: 'Balarat',
+        label: 'Ballarat',
         value: '3005'
       }
     ];
--- a/app/components/Dropdown/Dropdown.js
+++ b/app/components/Dropdown/Dropdown.js
@@ -101,20 +101,31 @@ export default class DropDownDemo extends Component {
       focus,
       invalid
     });
+
     const options = [
       {
-        value: '1',
-        label: 'Developer'
+        label: 'Major Cities',
+        value: [
+          {
+            label: 'Melbourne',
+            value: '3004'
+          },
+          {
+            label: 'Sydney',
+            value: '3002'
+          }
+        ]
       },
       {
-        value: '2',
-        label: 'Tester'
+        label: 'Balarat',
+        value: '3005'
       },
       {
-        value: '3',
-        label: 'Really really long job title that is not gonna fit'
+        label: 'Really really long job title that is not gonna fit',
+        value: '3'
       }
     ];
+
     const dropdown = (
       <SeekDropDown
         id="jobTitles"
--- a/app/components/Dropdown/Dropdown.js
+++ b/app/components/Dropdown/Dropdown.js
@@ -104,6 +104,10 @@ export default class DropDownDemo extends Component {
 
     const options = [
       {
+        label: 'Really really long job title that is not gonna fit',
+        value: '3'
+      },
+      {
         label: 'Major Cities',
         value: [
           {
@@ -119,10 +123,6 @@ export default class DropDownDemo extends Component {
       {
         label: 'Balarat',
         value: '3005'
-      },
-      {
-        label: 'Really really long job title that is not gonna fit',
-        value: '3'
       }
     ];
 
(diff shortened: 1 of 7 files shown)


react/fields/Dropdown/Dropdown.js
```diff
@@ -77,10 +77,6 @@ export default class Dropdown extends Component {
     options: PropTypes.arrayOf((propValue, key, componentName) => {
       const { value, label } = propValue[key];
 
-      if (typeof value !== 'string') {
-        return new Error(`Invalid prop \`options[${key}].value\` of type \`${typeof value}\` supplied to \`${componentName}\`, expected \`string\`.`);
-      }
-
```
mjt01: Rather than removing this check we should redefine our expectations here to something like:
```js
options: PropTypes.arrayOf(
  PropTypes.shape({
    value: PropTypes.oneOfType([
      PropTypes.arrayOf(PropTypes.object),
      PropTypes.string
    ]).isRequired,
    label: PropTypes.string
  })
)
```

react/fields/Dropdown/Dropdown.js
```diff
@@ -152,14 +156,12 @@ export default class Dropdown extends Component {
...
-            <option
-              value={value}
-              key={value}
-              className={styles.option}>
-              { label }
-            </option>
-          ))
+          options.map(({ value, label }) => {
+            if (Array.isArray(value)) {
+              return (<optgroup value="" label={label} key={label}>{value.map(this.renderOption)}</optgroup>);
```
mjt01: Do we have any default styling for our option group names drop downs? @tcrook1 @stephaniemoss?

mjt01:
Adding Dropdown to the navigation will require adding it to the route list in the [webpack config for ghpages](https://github.com/SEEK-Jobs/seek-style-guide/blob/master/webpack.gh-pages.server.config.js#L18). This will ensure the static rendered app has a html snapshot to load when that route is requested.

react/fields/Dropdown/Dropdown.test.js
```diff
@@ -93,6 +90,27 @@ describe('Dropdown', () => {
...
+      render(<Dropdown inputProps={{ value: '' }} options={opts} />);
+      expect(optionGroup.props.label).to.equal('suburbs');
+    });
+    it('should render optgroup children correctly', () => {
+      render(<Dropdown inputProps={{ value: '' }} options={opts} />);
+      expect(optionGroup.props.children.length).to.equal(2);
+    });
+    it('should render 1st optgroup child value correctly', () => {
+      render(<Dropdown inputProps={{ value: '' }} options={opts} />);
+      console.log(childOptions);
```
mjt01: this can be removed

react/fields/Dropdown/Dropdown.test.js
```diff
@@ -93,6 +90,27 @@ describe('Dropdown', () => {
     });
   });
 
+  describe('Option Group', () => {
+    const opts = [{ label: 'suburbs', value: [{ label: 'truganina', value: '3029' }, { label: 'wl', value: '3029' }] }];
+    it('should render optgroup correctly', () => {
+      render(<Dropdown inputProps={{ value: '' }} options={opts} />);
```
mjt01: given all the tests are rendering the same thing, you can optimise this be calling `render` in a `before` within this `describe`. Then it'll just be firing off all the expectations nice and fast after a single render.

LeviSeek:
@mjt01

app/components/Dropdown/Dropdown.js
```diff
@@ -101,20 +101,31 @@ export default class DropDownDemo extends Component {
...
+          {
+            label: 'Sydney',
+            value: '3002'
+          }
+        ]
       },
       {
-        value: '3',
-        label: 'Really really long job title that is not gonna fit'
+        label: 'Balarat',
```
mengtzu: Ballarat :)

LeviSeek: typo gone now.

mjt01:
Sorry I havent been able to get to this sooner.

:art:

PR #73, opened 2017-01-09
Upgrade react-autosuggest to 8.0.0

moroshko:
[`react-autosuggest` 8.0.0](https://github.com/moroshko/react-autosuggest/releases/tag/v8.0.0) removes `redux` and `react-redux` from its dependencies. This should reduce the size of the production js bundle.
Hello from Silicon Valley ✋


mjt01:
Open Source Level: Boss 🏆 

Upgrading your package and sending PRs to consumers! You're the best. I'll give it a once over. Thanks mate 🙌

moroshko:
I'm curious to know the effect it had on the production js bundle 🤔

mjt01:
@moroshko the chalice production bundle dropped from 1654kb to 1649kb. It would probably have a bigger affect on projects that did not use redux/react-redux.

moroshko:
Did you mean `mb` instead of `kb`?

mjt01:
No, its not 1654mb! That's in kilobytes

moroshko:
Oh, sorry, not sure what I was thinking 🤔

markdalgleish:
@moroshko hello from SEEK! 👋 😄

mengtzu:
Any reason not to merge this?  I installed react-autocomplete 8 and now I'm getting peer dependency complaints >_<

markdalgleish:
@mengtzu could you post the exact error you're seeing?

mengtzu:
`npm WARN EPEERINVALID seek-style-guide@13.1.1 requires a peer of react-autosuggest@^7.0.1 but none was installed.`

markdalgleish:
I'd say it seems like this needs to be a regular dependency, not a peer. @mjt01 thoughts?

mjt01:
Reason for it being a peer is otherwise every app would include `react-autosuggest` even if they weren't using that component. We definitely need to look at dependency management as a whole, but that should not be a blocker for this PR.

markdalgleish:
Cool, sounds good to me. Given it's a major version bump for react-autosuggest, do we consider it a major version for the style guide too? Seems like the safest bet, and we're not scared of major versions over here 😄

markdalgleish:
👍

PR #74, opened 2017-01-10
New tick icon

georgespyropoulos:
Added a new tick icon to the style.  Renamed the old `TickIcon` to `TickInCircleIcon`. 

We realise this will break existing usage of `TickIcon` but the tick icon is only used in one place in Huston so this would be an easy fix.  I'm happy to put together a PR for the chalice repo to update the styleguide version and provide a new name for the tick icon.  

I'm open to different names for the new icon as well if you have any suggestions? 

screen shot 2017-01-10 at 4 02 40 pm


mjt01:
@georgespyropoulos I'm happy enough with the naming, the question I would have is around context of use and sizing. For example we have rolled two separate icons together in `<StarIcon />` and we use the `filled` prop to control which icon is used.

Should this be a `<TickIcon />` with a `circle` property? Do both icons take up the same space in the `viewbox` of the `svg` so that toggling between them the only change is the border?

Sorry for the 20 questions, just want to make sure we represent it correctly. Feel free to pull me into any discussion about this.

markdalgleish:
>Should this be a <TickIcon /> with a circle property?

I'd probably prefer this from a dev consumer perspective, but I'm concerned that we should probably maintain a 1:1 relationship between the design assets and the components here. Designers are going to treat these as two different shapes, and merging them into one would probably end up being a manual dev step.

mjt01:
I hear that. We are already maintaining multiple assets for things like the heart and star, so definitely a discussion to be had. I like one representation with multiple states, but we would need to have that conversation

markdalgleish:
I think merging multiple shapes into a single component should probably only be done if we can automate it somehow, which probably isn't worth the effort right now.

georgespyropoulos:
Right, I wasn't aware that we had filled / not filled states for the star and heart.  Merging Tick and TickInCircle into one component would be consistent with the star and heart however I think the tick is a more complicated icon.

This is an interesting question:

> Do both icons take up the same space in the viewbox of the svg so that toggling between them the only change is the border?

This is where the tick is different to the heart and star - it's not as simple as turning fill on and off, there are other decisions to be made.  Also, the tick in the circle has round corners but the other tick has square edges.  If you made a tick in a square, would it have square or round corners?

georgespyropoulos:
@markdalgleish @mjt01 are you both happy to merge this PR?

markdalgleish:
Just to make it a bit easier to say out loud, can we rename it to `TickCircleIcon`?

georgespyropoulos:
Yes that does sounds better.  I've pushed up the change

markdalgleish:
👍  Nice one. @mjt01?

mjt01:
:art:

georgespyropoulos:
Yep will do.  I don't have write access to the repo... can someone either give me access or merge this PR and do the release please?

PR #76, opened 2017-01-11
fix textarea help style

LeviSeek:
fixes in this PR:
- fix to avoid the count number being wrapped to a new line.

issue:
image

- add more space between the help text and the count, since given some specific resolution, the help text and count will be joined together.

issue:
image

### DIFF
--- a/react/fields/Textarea/Textarea.less
+++ b/react/fields/Textarea/Textarea.less
@@ -57,7 +57,6 @@
 .footer {
   display: flex;
   justify-content: space-between;
-  align-items: center;
   min-height: (@grid-row-height * @standard-type-row-span);
   margin-bottom: @grid-row-height;
   align-items: flex-start;
--- a/react/fields/Textarea/Textarea.less
+++ b/react/fields/Textarea/Textarea.less
@@ -60,6 +60,7 @@
   align-items: center;
   min-height: (@grid-row-height * @standard-type-row-span);
   margin-bottom: @grid-row-height;
+  align-items: flex-start;
 }
 
 .help {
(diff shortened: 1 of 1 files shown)


react/fields/Textarea/Textarea.less
```diff
@@ -60,6 +60,7 @@
   align-items: center;
   min-height: (@grid-row-height * @standard-type-row-span);
   margin-bottom: @grid-row-height;
+  align-items: flex-start;
```
markdalgleish: This is a duplicate rule (see line 60 above)

LeviSeek: yep, will remove the duplicated rule

mjt01:
Should the count align on the same baseline as the first line of text rather than vertically centred?

markdalgleish:
@mjt01 just checked this out for myself, seems like it's already the case:

screen shot 2017-01-11 at 2 09 44 pm

markdalgleish:
Note that this PR triggered the creation of #77 since it's a bit tedious right now to check out a branch and build it locally just so you can play around with it.

LeviSeek:
yep, aligned on the same baseline @mjt01 

image

markdalgleish:
Seems good to me. Although this definitely highlights that the QA process needs to be improved a bit.

markdalgleish:
To clarify, I mean that we're missing some automated linting and tests, and verifying cross browser is quite time consuming, needs a bit more streamlining.

markdalgleish:
@LeviSeek thanks so much for your PR. Love your work 🎨 ❤️

LeviSeek:
@markdalgleish

PR #80, opened 2017-01-13
Add MIT license

markdalgleish:


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2016 SEEK
```
mengtzu: Might want to bump that to 2017 actually

markdalgleish: Nice catch. And this right here is why I have code reviews, even for simple changes 😄

PR #82, opened 2017-01-13
Fix documentation layout

markdalgleish:
We've added too many navigation items for the existing layout, causing it to wrap unexpectedly.

At some point we'll need a more holistic change to the documentation layout, but this at least gets us back to a non-broken state.

Preview:
screen shot 2017-01-13 at 3 54 09 pm

### DIFF
--- a/app/components/App/components/Header/Header.less
+++ b/app/components/App/components/Header/Header.less
@@ -1,15 +1,14 @@
 @import (reference) "~seek-style-guide/theme";
 
 .root {
-  height: (@grid-row-height * 14);
-  display: flex;
-  align-items: center;
-  justify-content: space-between;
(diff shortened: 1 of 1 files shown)


app/components/App/components/Header/Header.less
```diff
@@ -1,15 +1,14 @@
...
-  align-items: center;
-  justify-content: space-between;
+  height: (@grid-row-height * 16);
+  padding-top: (@grid-row-height * 2);
 }
 
 .logo {
   display: flex;
   align-items: center;
+  margin-left: -(@grid-gutter-width / 2);
```
mengtzu: I still instinctively twitch when I see a negative margin.  Fine to get it unbroken though!

PR #83, opened 2017-01-15
Upgrade to Node v6

markdalgleish:
Since I'm running the latest version of Node, I kept getting warnings about the `--harmony-proxies` CLI flag whenever running the tests. Figured it'd be worth upgrading to the latest version, where proxies are enabled by default, rather than working around it locally.


mengtzu:
🚀

mjt01:
@markdalgleish interested in the choice between Node 7 and 6 with regard to LTS?

markdalgleish:
@mjt01 interesting—I just went with the latest stable version, but it makes sense to only use even-numbered releases. Should I roll back to v6?

markdalgleish:
@mjt01 decided to roll back to v6 based on LTS, like you said.

markdalgleish:
@mjt01 Unfortunately this means the branch name is incorrect. Do we care, or should I close this and open a new PR?

jahredhope:
Would it be worth adding Node engine requirements to package.json as well?
https://docs.npmjs.com/files/package.json#engines

and/or mention the engine requirements in a `README.md` or `CONTRIBUTING.md`?

markdalgleish:
@jahredhope good point, I think we should add a `CONTRIBUTING.md`. Might raise an issue for this, don't want this PR to grow too might right now.

markdalgleish:
Branch name is wrong, so closing in favour of #86.

PR #87, opened 2017-01-16
Run ESLint during test script

markdalgleish:
I couldn't release the latest version of the style guide because of a linting error. Turns out we run ESLint before release but not as part of the test script, which is why it also wasn't being run in CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,10 +6,9 @@
   "scripts": {
     "start": "BASE_HREF=/ node server",
     "lint": "eslint app react test ./*.js",
-    "test": "BABEL_ENV=test mocha 'react/**/*.test.js' --compilers js:babel-register --require test/setup.js",
-    "prebuild": "npm run lint && npm test",
-    "local-build": "npm run prebuild && rm -rf dist/ && mkdir dist && NODE_ENV=production BASE_HREF=/ webpack --config webpack.gh-pages.server.config.js && rm dist/render.js && NODE_ENV=production webpack --config webpack.gh-pages.client.config.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -6,10 +6,9 @@
   "scripts": {
     "start": "BASE_HREF=/ node server",
     "lint": "eslint app react test ./*.js",
-    "test": "BABEL_ENV=test mocha 'react/**/*.test.js' --compilers js:babel-register --require test/setup.js",
-    "prebuild": "npm run lint && npm test",
-    "local-build": "npm run prebuild && rm -rf dist/ && mkdir dist && NODE_ENV=production BASE_HREF=/ webpack --config webpack.gh-pages.server.config.js && rm dist/render.js && NODE_ENV=production webpack --config webpack.gh-pages.client.config.js",
-    "gh-pages-build": "npm run prebuild && rm -rf dist/ && mkdir dist && NODE_ENV=production BASE_HREF=/seek-style-guide/ webpack --config webpack.gh-pages.server.config.js && rm dist/render.js && NODE_ENV=production webpack --config webpack.gh-pages.client.config.js",
+    "test": "BABEL_ENV=test mocha 'react/**/*.test.js' --compilers js:babel-register --require test/setup.js && npm run lint",
```
mjt01: I would recommend linting first as a fast fail. What do you think??

markdalgleish: I actually find it to be more frustrating this way, because I see linting as more of a polish phase at the end, rather than a blocker for testing actual functionality.

markdalgleish: You can still run `npm run lint` if you need to quickly iterate on lint fixes.

mjt01: I would agree if this you were in watch mode and iterating on tests, but for the `npm test` command it feels right. But I don't want to frustrate you unnecessarily 🤡

markdalgleish: Nothing here is set in stone. Leave it and debate later over ☕️ ? 😄

jahredhope: I'd vote linting at the end like it is.
Hopefully your linting is done in your IDE, when you're running this step linting is just a formality to make sure.

PR #90, opened 2017-01-17
fix(test): Test semantic-release

markdalgleish:
This is just a single character readme change, but the commit message should trigger a patch release of `v15.1.1`.


mjt01:
:art:

markdalgleish:
Here goes! 📦 🚀

PR #93, opened 2017-01-18
build(npm): Clean up npm scripts

markdalgleish:
Was about to start work on automatic GitHub Pages deployments, but found myself having to work amongst some outdated scripts, so decided I'd first clean things up a bit.

I've made the following changes:
- Removed the `preversion` and `postversion` scripts, since we're now using [semantic-release](https://github.com/semantic-release/semantic-release).
- Renamed `deploy` to `deploy-gh-pages` to be consistent with `deploy-surge`.
- Since `build` and `build-gh-pages` were mostly identical, I refactored `build-gh-pages` to call `build` with an override to the `BASE_HREF` environment variable, which was the only difference.
- Since the `build` script was getting way too long, I split out two smaller `compile-render` and `compile-client` scripts
- Split out a separate `unit` script, so now the `test` script only runs other, more focused scripts.
- Removed `install-git-hooks` and accompanying bash script since we're now covered by [Travis CI](https://travis-ci.org/seek-oss/seek-style-guide).
- Generally re-ordered scripts to improve readability, although this is subjective.


package.json
```diff
@@ -5,16 +5,16 @@
   "main": "index.js",
   "scripts": {
     "start": "BASE_HREF=/ node server",
+    "test": "npm run unit && npm run lint && npm run build",
```
mengtzu: In the description you said you split out `unit` so that `test` only runs other, more focussed tests. I actually prefer unit to still be in here because I think that matches the expectation, but it does contradict your comment!

package.json
```diff
@@ -5,16 +5,16 @@
...
     "lint": "eslint app react test ./*.js",
-    "test": "BABEL_ENV=test mocha 'react/**/*.test.js' --compilers js:babel-register --require test/setup.js && npm run lint && npm run build",
-    "build": "rm -rf dist/ && mkdir dist && NODE_ENV=production BASE_HREF=/ webpack --config webpack.static.render.config.js && rm dist/render.js && NODE_ENV=production webpack --config webpack.static.client.config.js",
-    "build-gh-pages": "rm -rf dist/ && mkdir dist && NODE_ENV=production BASE_HREF=/seek-style-guide/ webpack --config webpack.static.render.config.js && rm dist/render.js && NODE_ENV=production webpack --config webpack.static.client.config.js",
-    "preversion": "npm run build-gh-pages",
-    "postversion": "npm publish && git push && git push --tags && npm run deploy",
-    "deploy": "gh-pages -d dist && printf '\nURL: https://seek-oss.github.io/seek-style-guide\n\n'",
-    "deploy-surge": "surge -p ./dist",
+    "unit": "BABEL_ENV=test mocha 'react/**/*.test.js' --compilers js:babel-register --require test/setup.js",
```
mengtzu: Very mildly prefer the Company Reviews (and others?) convention of calling this `test-unit`.  Probably less important if you don't have too many other types of tests.

markdalgleish: 👍

PR #99, opened 2017-01-20
ci(gh-pages): Log GitHub Pages deployment errors

markdalgleish:
Deployments are still failing for some reason, but it's not reflected in the build logs. I realised I forgot to log deployment errors, so I've added some extra logging, but a refactor was required to ensure that all console logs are filtered.

### DIFF
--- a/scripts/deploy-gh-pages.js
+++ b/scripts/deploy-gh-pages.js
@@ -3,27 +3,32 @@ var ghpages = require('gh-pages');
 var basePath = path.join(__dirname, '../dist');
 var repoUrl = require('../package.json').repository.url;
 
-var makeConfig = function(token) {
-  if (token) {
-    // CI deployment config
-    return {
-      repo: repoUrl.replace(/^https?:\/\//, 'http://' + token + '@'),
-      logger: function(message) {
-        // Hide token from build logs
-        message = message.replace(new RegExp(token, 'g'), '[TOKEN]');
-        console.log(message);
-      }
-    };
+var GH_PAGES_TOKEN = process.env.GH_PAGES_TOKEN;
+var tokenRegex = GH_PAGES_TOKEN ? new RegExp(GH_PAGES_TOKEN, 'g') : null;
+
+var log = function(message) {
+  if (tokenRegex) {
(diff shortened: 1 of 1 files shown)


scripts/deploy-gh-pages.js
```diff
@@ -3,27 +3,38 @@ var ghpages = require('gh-pages');
 var basePath = path.join(__dirname, '../dist');
 var repoUrl = require('../package.json').repository.url;
 
-var makeConfig = function(token) {
-  if (token) {
+var GH_TOKEN = process.env.GH_TOKEN;
+var tokenRegex = GH_TOKEN ? new RegExp(GH_TOKEN, 'g') : null;
```
jahredhope: Does the GH_TOKEN ever contain Regex special characters?

scripts/deploy-gh-pages.js
```diff
@@ -3,27 +3,38 @@ var ghpages = require('gh-pages');
...
+var log = function(message) {
+  if (tokenRegex) {
+    // Hide token from build logs
+    message = message.replace(tokenRegex, '[GH_TOKEN]');
+  }
+
+  console.log(message);
+};
+
+var makeConfig = function() {
```
mjt01: This function could be simplified now:
```
return {
  repo: GH_TOKEN ? repoUrl.replace(/^https?:\/\//, 'http://' + GH_TOKEN + '@') : repoUrl,
  logger: log
};
```

scripts/deploy-gh-pages.js
markdalgleish: It shouldn't, it's just alphanumeric.

scripts/deploy-gh-pages.js
markdalgleish: I've been deliberately optimising for readable config without forcing you to execute the code in your head, hence why I'm maintaining two separate objects. Does that sound okay?

mjt01: I hear what you're saying but not sure it's clearer having nested if statements that all hinge on the presence of `GH_TOKEN`.

PR #103, opened 2017-02-02
feat(components): Add initial layout and typographical components

markdalgleish:
Layout components:
- PageBlock
- Section
- AsidedLayout
- Card

Typographical components:
- Text
- Strong
- Secondary

These components are still experimental, so they are currently undocumented.

This also introduces the “playground” section of the style guide site as a dedicated space to experiment and collaborate on existing and upcoming components.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -191,7 +191,6 @@
     "react/sort-prop-types": 0,
     "react/wrap-multilines": 2,
 
-    "react/jsx-boolean-value": [2, "always"],
     "react/jsx-closing-bracket-location": [2, { "selfClosing": "tag-aligned", "nonEmpty": "after-props" }],
     "react/jsx-curly-spacing": [2, "never", { "allowMultiline": true }],
     "react/jsx-equals-spacing": [2, "never"],
--- a/react/AsidedLayout/AsidedLayout.js
+++ b/react/AsidedLayout/AsidedLayout.js
@@ -1,0 +1,38 @@
+import styles from './AsidedLayout.less';
+import React, { PropTypes } from 'react';
+import classnames from 'classnames';
+
+const defaultRenderAside = () => null;
+
+const conditionallyRenderAside = (condition, renderAside, size) => (
+  condition ?
+    <div className={styles.aside} style={{ flexBasis: size }}>
+      {renderAside()}
+    </div> :
+    null
+);
+
+export default function AsidedLayout({ className, children, renderAside = defaultRenderAside, size, reverse }) {
+  return (
+    <div
+      className={classnames({
+        [className]: className,
+        [styles.root]: true,
+        [styles.reverse]: reverse
+      })}>
+      { conditionallyRenderAside(reverse, renderAside, size) }
+      <div className={styles.content}>
+        {children}
+      </div>
+      { conditionallyRenderAside(!reverse, renderAside, size) }
+    </div>
+  );
+}
+
+AsidedLayout.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  renderAside: PropTypes.func,
+  size: PropTypes.string,
+  reverse: PropTypes.bool
+};
--- a/react/AsidedLayout/AsidedLayout.less
+++ b/react/AsidedLayout/AsidedLayout.less
@@ -1,0 +1,21 @@
+@import (reference) "~seek-style-guide/theme";
+
+@media only screen and (min-width: unit(@responsive-type-breakpoint, px)) {
+  .root {
+    display: flex;
+  }
+
+  .reverse {
+    flex-direction: row-reverse;
+  }
+
+  .content {
+    flex: 1 1;
+  }
+
+  .aside {
+    padding-left: @grid-gutter-width;
+    flex-grow: 0;
+    flex-shrink: 0;
+  }
+}
--- a/react/Card/Card.js
+++ b/react/Card/Card.js
@@ -1,0 +1,22 @@
+import styles from './Card.less';
+import React, { PropTypes } from 'react';
+import classnames from 'classnames';
+
+export default function Card({ className, children, transparent }) {
+  return (
+    <div
+      className={classnames({
+        [className]: className,
+        [styles.root]: true,
+        [styles.transparent]: transparent
+      })}>
+      {children}
+    </div>
+  );
+}
+
+Card.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  transparent: PropTypes.bool
+};
--- a/react/PageBlock/PageBlock.js
+++ b/react/PageBlock/PageBlock.js
@@ -1,0 +1,18 @@
+import styles from './PageBlock.less';
+import React, { PropTypes } from 'react';
+import classnames from 'classnames';
+
+export default function PageBlock({ children, className }) {
+  return (
+    <div className={classnames(className)}>
+      <div className={styles.content}>
+        {children}
+      </div>
+    </div>
+  );
+}
+
+PageBlock.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.node.isRequired
+};
--- a/react/PageBlock/PageBlock.less
+++ b/react/PageBlock/PageBlock.less
@@ -1,0 +1,11 @@
+@import (reference) "~seek-style-guide/theme";
+
+.content {
+  max-width: @grid-container-width + (@grid-gutter-width * 2);
+  margin: 0 auto;
+
+  @media only screen and (min-width: unit(@responsive-type-breakpoint, px)) {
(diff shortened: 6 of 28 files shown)


react/PageBlock/PageBlock.less
```diff
@@ -0,0 +1,11 @@
+@import (reference) "~seek-style-guide/theme";
+
+.content {
+  max-width: 940px;
```
mjt01: Can you use `@grid-container-width` for this instead.

react/AsidedLayout/AsidedLayout.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default function AsidedLayout({ className, children, renderAside = defaultRenderAside, size, reverse }) {
+  return (
+    <div
+      className={classnames({
+        [className]: className,
+        [styles.root]: true,
+        [styles.reverse]: reverse
+      })}>
+      { reverse ?
```
mjt01: You could achieve the same result with `flex-direction` right?

react/PageBlock/PageBlock.less
```diff
@@ -0,0 +1,11 @@
...
+  margin: 0 auto;
+
+  @media only screen and (min-width: 740px) and (max-width: 1000px) {
```
mjt01: We introduced `@responsive-type-breakpoint` for the min breakpoint

react/Text/Text.js
```diff
@@ -0,0 +1,29 @@
...
+  subheading,
+  headline,
+  heading,
+  hero,
+  raw
+}) {
+  return (
+    <div className={classnames({
+      [styles.root]: true,
+      [className]: className,
```
mjt01: Would it be more correct for `className` to be last so it's an override?

react/Section/Section.less
```diff
@@ -0,0 +1,19 @@
+@import (reference) "~seek-style-guide/theme";
+
+.root {
+  padding: (@grid-row-height * 2) @grid-gutter-width;
+  @media only screen and (min-width: 980px) {
```
mjt01: We have `980px` here and `1000px` for `PageBlock`, should these have a common value?

react/AsidedLayout/AsidedLayout.js
```diff
@@ -0,0 +1,30 @@
...
+import React from 'react';
+import classnames from 'classnames';
+
+const defaultRenderAside = () => null;
+
+export default function AsidedLayout({ className, children, renderAside = defaultRenderAside, size, reverse }) {
+  return (
+    <div
+      className={classnames({
+        [className]: className,
```
mjt01: As per the other, `className` should probably be last as an override

react/AsidedLayout/AsidedLayout.less
```diff
@@ -0,0 +1,21 @@
...
+  .reverse {
+    flex-direction: row-reverse;
+  }
+
+  .content {
+    flex: 1 1;
+  }
+
+  .aside {
+    padding: 0 0 0 @grid-gutter-width;
```
mjt01: Any reason for not expressing this as `padding-left`?

react/Card/Card.js
```diff
@@ -0,0 +1,15 @@
+import styles from './Card.less';
+import React from 'react';
+import classnames from 'classnames';
+
+export default function Card({ className, children, transparent }) {
+  return (
+    <div className={classnames({
+      [className]: className,
```
mjt01: ☝️

react/PageBlock/PageBlock.js
```diff
@@ -0,0 +1,13 @@
+import styles from './PageBlock.less';
+import React from 'react';
+import classnames from 'classnames';
+
+export default function PageBlock({ children, className }) {
+  return (
+    <div className={classnames(className)}>
```
mjt01: Did we see this return "false", should this be `classnames({[className]: className})`?

react/AsidedLayout/AsidedLayout.js
mengtzu: Given we're passing a reverse class anyway, seems like it'd be neater

react/AsidedLayout/AsidedLayout.less
```diff
@@ -0,0 +1,21 @@
+@import (reference) "~seek-style-guide/theme";
+
+@media only screen and (min-width:740px) {
```
mengtzu: Another magic number here.

In CR we use mixins to generate our media queries, probably not a huge difference but looks a little neater.

react/Text/Text.less
```diff
@@ -0,0 +1,31 @@
+@import (reference) "~seek-style-guide/theme";
+
+.root {
+  .standardTextResponsive();
+  padding-bottom: @grid-row-height;
+
+  .root {
+    position: relative;
```
mengtzu: Is this highlighting of nesting a common pattern?  Could be a mixin candidate

react/AsidedLayout/AsidedLayout.less
mjt01: `740px` is not magic, it's the `@responsive-type-breakpoint` as specified by the design team. (should probably be renamed as it's more than type)

mengtzu: Yeah, unpacking my comment would be "probably meaningful number that *looks* like magic and should use an appropriately named variable"

react/AsidedLayout/AsidedLayout.js
markdalgleish: @mjt01 doing it in the markup means you're presenting content to screen readers in the same order as the mobile view, which is much better IMO.

react/PageBlock/PageBlock.js
markdalgleish: Either works, but this form is a lot simpler.

react/AsidedLayout/AsidedLayout.js
markdalgleish: Just to be clear, the order of classes in a `className` prop has no impact on specificity. I decided to put `className` first to make it obvious that styles can be overridden, but happy to settle on the opposite convention. What do you think?

react/Text/Text.less
markdalgleish: Nope, this is the first instance of it, and may be the last. In this case, it was introduced because nesting Text components is an easy mistake to make, with very subtle side effects, and debugging it is quite tricky.

react/PageBlock/PageBlock.js
mjt01: I agree this is simpler, but was this resulting in `class="false"` on the element?

markdalgleish: No, it results in a blank class attribute.

react/AsidedLayout/AsidedLayout.js
mjt01: I'm not sure you would necessarily want asides content announced first. We can revisit once we have some example implementations

markdalgleish: The use case is when the sidebar contains important contextual information, but is much, much smaller than the main content. If you're reading the content top-to-bottom, it's important to read it first, but it can happily sit on the right-hand side if your screen is large enough. Make sense?

markdalgleish:
Since we're using the unlinked "playground" page to validate our changes as we iterate on this PR, @dborysiewicz and I figured it makes sense to include it in this PR too so we can standardise its usage and officially share it with each other, even if we're likely to make big changes to it soon. @mjt01 sound okay?

markdalgleish:
@mjt01 I've updated the PR based on your feedback, addressed most of it. Let me know what you think.

.eslintrc
```diff
@@ -191,7 +191,6 @@
     "react/sort-prop-types": 0,
     "react/wrap-multilines": 2,
 
-    "react/jsx-boolean-value": [2, "always"],
```
markdalgleish: I had to remove this line to support the `<Text heading>...` pattern.

react/PageBlock/PageBlock.less
```diff
@@ -0,0 +1,11 @@
+@import (reference) "~seek-style-guide/theme";
+
+.content {
+  max-width: @grid-container-width + (@grid-gutter-width * 2);
```
markdalgleish: @mjt01 since you last reviewed this, we noticed that this was incorrectly setting the page width at `940px`, when the page actually needs to be `980px` including gutters.

react/AsidedLayout/AsidedLayout.js
mjt01: Personally i prefer after, but i am happy if it's consistent.

mjt01:
Looks good to me. As discussed we are holding off reviewing the "supporting components" that are left in the `Atoms` folder. We can look at those if/when they get promoted out.

Be good to start experimenting with these in the real products. Super excited.

:art:

PR #105, opened 2017-02-03
fix(checkbox): Fix checkbox size

vedranio:
Checkbox was 25px instead of 27px (3 grid rows) because the calc was subtracting border widths


markdalgleish:
🎨 👍

PR #106, opened 2017-02-03
fix(components): Spread props on presentational components

markdalgleish:
Any extra props passed to presentation components were being ignored, forcing you to add redundant wrapper elements if you needed to add aria roles, data attributes etc.


mjt01:
:art:

PR #107, opened 2017-02-03
Fix checkbox colors

vedranio:


markdalgleish:
Looks like this branch includes the size fix too for some reason :thinking:

PR #108, opened 2017-02-05
fix(Checkbox): Fix checkbox colours

vedranio:
Fixed checkbox selected state background colour and tick colour so that they are in line with our
styles


markdalgleish:
The tick is still pink, is that correct?

markdalgleish:
🎨 👍

PR #109, opened 2017-02-06
feat(text): Add strong & secondary modifiers to Text component

mjt01:
This PR should clean up the implementation code around using the `Text` component by providing helper props that wire up sub-components internally:

**Before:**
```jsx
<Text><Secondary><Strong>My text</Strong></Secondary></Text>
```
**After:**
```jsx
<Text secondary strong>My text</Text>
```

Makes the application code much more readable

### DIFF
--- a/react/Text/Text.js
+++ b/react/Text/Text.js
@@ -11,7 +11,7 @@ const renderContent = ({ children, secondary, strong }) => {
     strong ? Strong : null
   ].filter(x => x);
 
-  return modifiers.reduce((content, mod) => mod({ children: content }), children);
+  return modifiers.reduce((content, Modifier) => <Modifier>{content}</Modifier>, children);
 };
 
 export default function Text({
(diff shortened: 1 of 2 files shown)


react/Text/Text.js
```diff
@@ -2,6 +2,18 @@ import styles from './Text.less';
...
+import Secondary from '../Secondary/Secondary';
+import Strong from '../Strong/Strong';
+
+const renderContent = ({ children, secondary, strong }) => {
+  const modifiers = [
+    secondary ? Secondary : null,
+    strong ? Strong : null
+  ].filter(x => x);
+
+  return modifiers.reduce((content, mod) => mod({ children: content }), children);
```
markdalgleish: Could we call it `Modifier` instead of `mod`, and use JSX in the reducer to make it a bit more obvious that we're building up a component tree?

e.g.
```js
return modifiers.reduce((content, Modifier) => <Modifier>{content}</Modifier>, children);
```

mjt01: Yep, i like it 👏

PR #110, opened 2017-02-06
fix(button): Prevent wrapping of button text

mjt01:
Fix for the following situation where the external layout forces the button text to wrap which should never be the case.

**Before:**
wrapping button labels

**After:**
truncated button labels

It makes it a less broken experience, but our implementations should hopefully never hit up against this anyway.

### DIFF
--- a/react/Button/Button.less
+++ b/react/Button/Button.less
@@ -24,6 +24,9 @@
   border-radius: 2px;
   transition: background-color 0.2s, transform .2s cubic-bezier(0.020, 1.505, 0.745, 1.235);
   cursor: pointer;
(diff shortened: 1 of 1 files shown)


markdalgleish:
Wondering if we should also add title text just in case?

mjt01:
Interesting, you could default it to the content/children if the `title` prop was not specified, but that could be something other than text, eg. an `<Icon />` component. What are your thoughts?

markdalgleish:
Yeah, good point, it's not as straightforward as I thought. We can leave that for later.

markdalgleish:
🎨 👍

PR #111, opened 2017-02-06
docs(contributing): Improve contributing guide

markdalgleish:
After working with @vedranio on a couple of pull requests, it became clear that our contribution docs needed a lot more work, particularly for those less familiar with the traditional workflow, so this is my first attempt at fleshing things out.


mengtzu:
🎨   Probably more iteration to be done in future but definitely a big step in the right direction.

PR #112, opened 2017-02-07
feat(columns): Add columns component

markdalgleish:
Note that this PR is more of a strawman proposal so that we can work together to get an initial version we're happy with. Keen to hear your thoughts.

### DIFF
--- a/app/components/Playground/Playground.js
+++ b/app/components/Playground/Playground.js
@@ -10,6 +10,7 @@ import {
   PageBlock,
   Section,
   AsidedLayout,
+  Columns,
   Card,
   Text
 } from 'seek-style-guide/react';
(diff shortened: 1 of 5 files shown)


markdalgleish:
The biggest questions I have are—as usual—around the management of whitespace. The most obvious problem is that sometimes you want whitespace between columns when they're stacked vertically on mobile, and sometimes you don't, e.g. compare the usage of columns to wrap the action buttons vs. wrapping card components.

markdalgleish:
There's also a question here of whether `<AsidedLayout>` could just be a variant of `<Columns>`, which is worth investigating, although this doesn't have to be a blocker for the MVP.

mjt01:
Im happy with this, as a first pass. Let's keep an open mind around new layouts (eg. masonry) and consolidating layouts (eg. AsidedLayout). Happy to iterate on this stuff.

:art:

mengtzu:
I am also happy with this, assuming we're OK with the default being no vertical spacing below the breakpoint (we can come back and adds a prop that does that later).  That's probably easier since figuring out what the vertical spacing should be might be mild tarpit.

🎨

PR #116, opened 2017-02-13
refactor(*): refactors responsive breakpoint, removes large-screen mixin

mengtzu:
BREAKING CHANGE: anyone consuming breakpoint externally will need to refactor

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,33 @@ All type should use the standard typographic mixins, e.g.
 
 Under the hood, these mixins leverage [basekick](https://github.com/mjt01/basekick) to ensure that typography sits correctly on the baseline grid.
 
+
+###Responsive Breakpoint and Type
+
+The style guide exposes one responsive breakpoint:
+
+```less
+@responsive-breakpoint: 740px;
+```
+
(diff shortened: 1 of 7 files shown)


markdalgleish:
Could we change the commit message + PR title to something like:
>refactor(less): Standardise responsive breakpoint, remove large-screen mixin

Also note that the message has been switched to imperative tense ("remove" instead of "removes").

The description should also be more explicit about what exactly is breaking, e.g.
>The `@responsive-type-breakpoint` variable has been renamed to `@responsive-breakpoint`. The `.large-screen()` mixin has also been removed, since it was undocumented and non-standard.

README.md
```diff
@@ -190,6 +190,33 @@ All type should use the standard typographic mixins, e.g.
 
 Under the hood, these mixins leverage [basekick](https://github.com/mjt01/basekick) to ensure that typography sits correctly on the baseline grid.
 
+
+###Responsive Breakpoint and Type
```
markdalgleish: There should be a space here after the hashes :female_detective:

markdalgleish:
:art: :+1: (just remember to clean up the commit message and description when merging)

PR #118, opened 2017-02-13
ci(deployment): Skip deployment step on fork builds

markdalgleish:
The Travis CI build step for PRs originating from forks are currently unable to pass because they
don't have access to secure environment variables, causing the deployment step to fail. Deployment
of static previews for PRs is a useful part of our workflow, but not something that should prevent
contribution from those who don't currently have access to the main repo.

I'm issuing this PR from my own personal fork to ensure this logic works correctly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,21 @@ notifications:
 before_script:
   - npm prune
 after_success:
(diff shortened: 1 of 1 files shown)


ryardley:
I didn't think I had review permissions. :)

markdalgleish:
That's so weird—you can approve the changes, but it doesn't affect the PR status.

PR #121, opened 2017-02-14
fix(babel): Remove babelrc

markdalgleish:
We've experienced issues with some tooling where the style guide's babelrc is being picked up in the context of an application or its test suite. In order to ensure this can't happen, we're moving away from babelrc in favour of passing config directly to babel-loader and babel-register.


markdalgleish:
It's worth noting that, while we could add `.babelrc` to an `.npmignore` file, we also want to allow people to work on the style guide in the context of their app (sourcing the code from GitHub rather than npm) without running into the same issue.

PR #124, opened 2017-02-15
feat(typography): Add new tiny text size

ryardley:
Added a smaller typographical style to text to accomodate copy that sits exactly within a single
grid row.


markdalgleish:
I'm not sure about this, it seems more likely to be a sign that our "small" type actually isn't correct, and we might need to discuss this further with some designers 🤔

ryardley:
Happy to close this until we get more consensus.

PR #125, opened 2017-02-15
fix(icons): Enlargen Sharing Icons

ryardley:
Sharing Icons were mistakenly supplied smaller than their viewBox which means sizing them correctly is hard with basekick.

Icons are now larger than they were and the size will need to be adjusted to be smaller if they already have been used.


markdalgleish:
🎨 👍

PR #126, opened 2017-02-16
docs(row): Proposal for a new Row component to handle vertical spacing

tamoore:


tamoore:
This could possibly replace the Separator proposal as the main problem I ran into was really around vertical spacing.

PR #127, opened 2017-02-16
feat(multiple): Changes to TextField, StyleGuideProvider & TickCircle

dborysiewicz:
BREAKING CHANGE:
**TextField**:
Rewrite on message interface to use new **Text** component which also using messageProps to switch colour & icons on messageProps.critical, messageProps.positive & messageProps.valid.
Default message style is **Text secondary**. 
Removed help in favour using message. 
Switched root div to block over inline-block.

**StyleGuideProvider**:
Removed global input margin-bottom 0px rule.

**TickCircle**:
Added filled version of icon.


react/fields/TextField/TextField.js
```diff
@@ -163,63 +168,57 @@ export default class TextField extends Component {
...
   }
 
   render() {
-    const { className, invalid, onClear, inputProps = {} } = this.props;
+    const { className, invalid, message, onClear, inputProps = {} } = this.props;
     const hasValue = (inputProps.value && inputProps.value.length > 0);
     const canClear = hasValue && (typeof onClear === 'function');
     const classNames = classnames({
       [styles.root]: true,
+      [styles.hasMessage]: message || false,
```
markdalgleish: The use of `|| false` here is redundant, I believe.

react/StyleGuideProvider/StyleGuideProvider.less
```diff
@@ -27,10 +27,6 @@
       -webkit-tap-highlight-color: transparent; /* For some Androids */
     }
 
-    input {
```
mengtzu: 👍

mengtzu:
🎨

markdalgleish:
Just a heads up, this PR has been close as we're iterating on the API. This is a tricky one to get right and we don't want too much churn, so please bear with us 😄

PR #128, opened 2017-02-16
feat(textfield): Improve TextField validity options

dborysiewicz:
BREAKING CHANGE:

FieldMessage:
Rewrite of message/help interface.
Message uses Text component in which text colours change on prop valid (true,false,null/undefined). Message default style is Text secondary.
Message icon also uses prop valid (true,false).
Message colour can be overridden using messageProps which as passed directly to Text. Removal of help in favour using message.
Message supports value false which turns off margin-bottom for when you don't want to reserve white space.

 TextField:
Rewrite to use new FieldMessage 
Removal of inline-block on root div.

MonthPicker:
Rewrite to use new FieldMessage

Dropdown:
Rewrite to use new FieldMessage

Textarea:
Rewrite to use new FieldMessage

StyleGuideProvider: 
Removed global input margin-bottom 0px rule.

TickCircle:
 Added filled version of icon.


react/fields/TextField/TextField.js
```diff
@@ -163,64 +167,58 @@ export default class TextField extends Component {
...
     }
 
-    return (
-      <ErrorIcon
-        filled={true}
-        className={styles.messageIcon}
-        svgClassName={styles.messageIconSvg}
-      />
-    );
+    return;
```
markdalgleish: This `return` is redundant.

dborysiewicz: @mjt01 ?

react/fields/TextField/TextField.js
```diff
@@ -163,64 +167,58 @@ export default class TextField extends Component {
...
 
-    const { messageProps } = this.props;
-    const allMessageProps = combineClassNames(messageProps, styles.message);
+    const { critical, positive, secondary, ...restMessageProps } = this.props.messageProps;
+    const allMessageProps = combineClassNames(restMessageProps, styles.message);
 
     return (
-      <p {...allMessageProps}>
-        {this.renderIcon()}
+      <Text {...allMessageProps} critical={(valid === false && !secondary) || critical} positive={(valid === true && !secondary) || positive} secondary={typeof valid === 'undefined' || secondary}>
```
mjt01: Can i suggest from a legibility perspective putting these props on indented new lines? Just makes it a lot easier to scan, sorry minor i know

react/fields/TextField/TextField.js
mjt01: I suggested `return null;` given this function is being called from a template, and there is a possibility of not entering either of the if statements.

react/fields/TextField/TextField.less
```diff
@@ -2,7 +2,7 @@
 @import (reference) "ClearField/ClearFieldVariables";
 
 .root {
-  display: inline-block;
+  display: block;
```
mjt01: You can remove `display` all together if you like.

mjt01:
:art:

markdalgleish:
🎨 👍 🚀

react/fields/FieldMessage/FieldMessage.js
```diff
@@ -0,0 +1,119 @@
...
+          {message}
+        </Text>
+      );
+    }
+
+    return null;
+  }
+
+  renderMessageIcon() {
+    const { valid } = this.props;
```
mjt01: You have a mix of `valid` and `invalid` in this component, but only one present in the `propTypes`

mjt01: Take that back, you have both, but should this not be a single boolean?

react/fields/FieldMessage/FieldMessage.js
```diff
@@ -0,0 +1,119 @@
...
+      );
+    }
+
+    return null;
+  }
+
+  render() {
+    const { invalid, help, helpProps, message } = this.props;
+
+    if (invalid || help || helpProps) {
```
mjt01: i see now! (reviews are hard)

react/fields/FieldMessage/FieldMessage.less
```diff
@@ -0,0 +1,21 @@
...
+.messageIcon {
+  position: relative;
+  top: 2px;
+  margin-right: 4px;
+}
+
+.messageIconSvg {
+  width: 14px;
+  height: 14px;
+  fill: currentColor;
```
mjt01: All Icon components set the `fill: currentColor` out of the box.

react/fields/FieldMessage/FieldMessage.js
```diff
@@ -0,0 +1,119 @@
...
+    this.renderMessage = this.renderMessage.bind(this);
+    this.renderMessageIcon = this.renderMessageIcon.bind(this);
+  }
+
+  renderMessage() {
+    const { message, valid } = this.props;
+
+    if (message) {
+      const { critical, positive, secondary, ...restMessageProps } = this.props.messageProps;
+      const allMessageProps = combineClassNames(restMessageProps, styles.message);
```
mjt01: This style does not exist anymore.

react/fields/Dropdown/Dropdown.js
```diff
@@ -242,8 +181,7 @@ export default class Dropdown extends Component {
         {this.renderLabel()}
         {this.renderChevron()}
         {this.renderSelect()}
-        {this.renderHelp()}
-        {this.renderMessage()}
+        <FieldMessage {...this.props} />
```
markdalgleish: 👏 👏 👏

markdalgleish:
I'm super excited about where this has ended up. Extracting out that common `<FieldMessage>` component was a much needed refactor, and it's done wonders for cleaning up this PR 👏

react/fields/FieldMessage/FieldMessage.js
```diff
@@ -54,11 +54,10 @@ export default class FieldMessage extends Component {
 
     if (message) {
       const { critical, positive, secondary, ...restMessageProps } = this.props.messageProps;
-      const allMessageProps = combineClassNames(restMessageProps, styles.message);
```
mjt01: The `combineClassNames` function is no longer used in here now. Nuke it 💣

PR #129, opened 2017-02-17
feat(layout): Add Card and Section groups

markdalgleish:
This PR is a proposal to introduce a `group` concept for layout components. When using a series of layout components that form a group, the current implementation adds a lot of unnecessary whitespace. Sometimes we want to visually separate a single `<Card>` or `<Section>` with minimum visual disruption. In the case of closely related cards, we add a single grey separator between them, and in the case of closely related sections, we only use 2 grid rows to separate them.

@tamoore I'd love your input on this, since this is primarily aimed to address the gaps you've raised.

### DIFF
--- a/app/components/Playground/Playground.js
+++ b/app/components/Playground/Playground.js
@@ -93,17 +93,19 @@ const renderJobDetailDate = () => (
 const renderJobDetailMetadata = () => (
   <div>
     <Card>
-      <Section>
-        <Button color="pink" className={styles.fullWidthTextField}>Apply for this job</Button>
-      </Section>
-      <Section>
-        <Text secondary>{'Applications for this role will take you to the advertiser\'s site.'}</Text>
-      </Section>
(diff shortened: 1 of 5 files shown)


tamoore:
screen shot 2017-02-20 at 9 49 35 am

☝️ I think that's a good way to handle that component. It adds the separation without introducing anything new. Bonus.

The `group` property makes sense when I read your description of the PR, but it wasn't intuitive to me when I first looked at it. Is it possible that we're trying to push functionality onto groups and cards that might be better placed with a Separate `<Group>` component? What are your thoughts?

markdalgleish:
@tamoore because of the way we scope CSS, we need a specific group per component, so our options are pretty much either `<Card group>` or <CardGroup>`. Personally, I kinda like the former because it doesn't require extra imports if you're iterating on a card layout and want to group them, for example.

PR #130, opened 2017-02-19
feat(rating): Add Ratings component

shawn-dsz:
**BREAKING CHANGE**: `@sk-yellow` has been renamed to `@sk-yellow-light`.

screen shot 2017-02-19 at 9 51 16 pm


theme/palette/elements.less
```diff
@@ -7,3 +7,4 @@
 @sk-footer: @sk-mid-gray-light;
 @sk-background: @sk-gray-light;
 @sk-yellow: #fffce3;
+@sk-gold: #ffc600;
```
mengtzu: I'm assuming this is one of the canonical colours!

stephaniemoss: Yes, #ffc600 is in the brand guidelines, generally used for Company Reviews (stars) :)

stephaniemoss: Actually, that @sk-yellow on the other hand... where is it used? Is @sk-gold replacing it?

mengtzu:
🎨

theme/palette/elements.less
mjt01: As discussed on slack, should we refactor `@sk-yellow` to be new colour documented in the guidelines, and assign the old value to `@sk-premium-job-highlight`?

react/Rating/Rating.js
```diff
@@ -0,0 +1,52 @@
...
+    </div>
+  );
+};
+
+Rating.displayName = 'Rating';
+
+Rating.propTypes = {
+  rating: PropTypes.number.isRequired,
+  className: PropTypes.string,
+  starClassName: PropTypes.string
```
mjt01: What are the overriding classes (`starClassName` & `className`) being used for? Ideally we are trying to reduce the need for these style of props going forward.

mengtzu: In CompanyReviews they're used to control the heights of the stars in various places.  We're happy to try life without them though!

markdalgleish: Are we adjusting the styling of the stars individually, rather than the entire rating?

I think the thinking here is that we should have predefined icon sizes that are officially supported, rather than every app manually overriding styles, but that's obviously a much larger discussion. We can certainly deprecate the need for class overrides in the future, once we've figured out what the supported subset looks like, but we're not quite there yet.

shawn-dsz:
Is there anything preventing this PR from being merged?

theme/palette/elements.less
shawn-dsz: `@sk-premium-job-highlight` is too specific to SERP I think, what if we want to use the same colour scheme in company reviews for example. Should the color name define its use?

mjt01: That's exactly my proposition. `@sk-yellow` should be the colour defined in the brand guidelines (currently it's not), everything in `elements.less` is largely that, element colours that are not captured in the core brand colours. @markdalgleish @stephaniemoss?

markdalgleish: @stephaniemoss do we use that light yellow colour (#fffce3) anywhere else? If not, do we plan to? Wondering whether this even belongs in the canonical list of colours, or whether it's actually just a one-off that we're trying to move away from.

mengtzu: IIRC correctly that yellow was used for notification popups in AdCentre, we referred to it as "Scott's favourite colour" owing to his bitterness at needing to use it.  It probably is used elsewhere but we probably should get away from it.

stephaniemoss: I recommend that @sk-yellow is #FFC600, as our primary yellow, and #FFFCE3 should be @sk-yellow-light, since that is the naming pattern we've used for the greens.

react/Rating/Rating.test.js
```diff
@@ -0,0 +1,84 @@
...
+    render(<Rating rating={3.3} />);
+    expect(element.type.displayName).to.equal('Rating');
+  });
+
+  it('should render 5 stars', () => {
+    render(<Rating rating={3.3} />);
+    expect(findStars().length).to.equal(5);
+  });
+
+  describe('should render correct start rating', () => {
```
ryardley: Should this be `should render correct star rating`?

theme/palette/elements.less
```diff
@@ -4,6 +4,7 @@
 @sk-focus: #1e90ff;
 @sk-highlight: #1d9ede;
 @sk-green-light: #5bc252;
+@sk-yellow: #ffc600;
+@sk-yellow-light: #fffce3;
```
markdalgleish: :+1:

markdalgleish: Could you also add this to the list of colours in the readme?

react/Rating/Rating.js
```diff
@@ -0,0 +1,52 @@
...
+import StarIcon from '../icons/StarIcon/StarIcon';
+import HalfStarIcon from './HalfStarIcon.svg';
+import Icon from '../icons/icon';
+import ScreenReaderOnly from '../Accessibility/ScreenReaderOnly';
+
+const getPercent = (rating, position) => Math.round(Math.min(Math.max(rating - position, 0), 1) * 100);
+
+const getStar = (percent, key, starClassName) => {
+  const props = {
+    key: `star-${key}`,
```
markdalgleish: The key doesn't have to be prefixed, it only has the be unique amongst its siblings.

react/Rating/Rating.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return <StarIcon {...props} />;
+};
+
+const Rating = ({ rating, starClassName, ...restProps }) => {
+  return (
+    <div {...restProps}>
+      <ScreenReaderOnly>
+        {rating} overall rating
```
markdalgleish: Should this say "out of 5"? If you can't see the screen, you can't tell that there are 5 stars, right?

markdalgleish:
About to approve this, but just a reminder to ensure that, when merging, the description is `BREAKING CHANGE: @sk-yellow has been renamed to @sk-yellow-light`.

markdalgleish:
🎨 📦 🚀

PR #131, opened 2017-02-20
fix(less): add px unit to @responsive-breakpoint, try to pretend it w…

mengtzu:
…as always there

### DIFF
--- a/theme/layout/breakpoints.less
+++ b/theme/layout/breakpoints.less
@@ -1,2 +1,2 @@
 // Breakpoint
-@responsive-breakpoint: 740;
+@responsive-breakpoint: 740px;


markdalgleish:
Can you rename the commit to `fix(less): Add px unit to @responsive-breakpoint` so it doesn't get clipped?

markdalgleish:
Does this mean that we can remove all the instances of `unit(...)` throughout the style guide?

mengtzu:
@markdalgleish is `unit()` the convention for the styleguide?

markdalgleish:
AFAIK we only use it because we were operating on a numeric variable that didn't have a corresponding unit, wanting to turn it into a pixel value. @mjt01, do you know more about this?

mjt01:
I think @markdalgleish is talking about removing all of these:
```css
@media only screen and (min-width: unit(@responsive-breakpoint, px))
```
The initial intent was modelling these as values and converting them when required, but the logic behind this breakpoint is not a calculation so making it `px` seems fair.

I agree with mark the clean up should be changing all the `unit()` calls to simply use the value now.

mengtzu:
@markdalgleish @mjt01 Sure, I can do that, I'll add it to this PR

mengtzu:
I'll open a new PR with a single commit

PR #132, opened 2017-02-20
refactor(less): add px to @responsive-breakpoint

mengtzu:
add px to @responsive-breakpoint
remove unit() call from uses of @responsive-breakpoint


mjt01:
Looks good to me, just not sure on the `refactor` vs `fix` commit type? Do you know what resultant version will be? Any breaking changes you can think of?

mengtzu:
I had `fix` because it was broken in Company Reviews and I thought I'd just omitted the `px`.  Now I understand I could have just used `unit` to solve our problem, this is really a refactor.

There *shouldn't* be any breaking changes (not for 16.x which already broke `@responsive-type-breakpoint`)

mjt01:
:art:

PR #133, opened 2017-02-20
feat(button): Allow Buttons to be built on custom components

ryardley:
Provide a "component" property to Button in order to pass in a custom html component name or a class reference.

### DIFF
--- a/react/Button/Button.less
+++ b/react/Button/Button.less
@@ -17,7 +17,7 @@
 
 .root {
   .touchableText();
-  display: block;
+  display: inline-block;
   text-decoration: none;
   text-align: center;
   padding: 0 @grid-gutter-width;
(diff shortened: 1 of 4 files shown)


markdalgleish:
👏  This is fantastic.

Given how critical this functionality is, it's probably worth adding a test that ensures extraneous props are carried through, e.g. that you can render an `href` when using `component="a"`.

react/Button/Button.less
```diff
@@ -17,6 +17,8 @@
 
 .root {
   .touchableText();
+  display: block;
```
markdalgleish: IIRC buttons are `inline-block` by default. I'm guessing we should probably change this?

ryardley:
In testing I noticed that text was being highlighted on anchor buttons when being clicked. So added `user-select: none;` to Button.

markdalgleish:
Awesome work, this was a much needed addition to the style guide 👍 

🎨 📦 🚀

ryardley:
Thanks! :)

PR #136, opened 2017-02-24
Add candidate header & footer

michaeltaranto:
Promoting the candidate header and footer. Special thanks to our working group @markdalgleish, @mattsjones & Birgitta.


README.md
```diff
@@ -76,6 +76,27 @@ render() {
 
 `StyleGuideProvider`'s props are used to set the page head properties using [Helmet](https://github.com/nfl/react-helmet).
 
+### Applying the Standard Header and Footer
```
mengtzu: Good job including a decent chunk of doco :1st_place_medal:

michaeltaranto: @markdalgleish 👏

react/Footer/Footer.common.less
```diff
@@ -0,0 +1,25 @@
+@import (reference) "~seek-style-guide/theme";
+
+@footer-vertical: ~"and (max-width: 767px)";
```
mengtzu: I assume the plan is to `740` this stuff at a later time?

markdalgleish: Yep. Phase one is to extract the current header and footer, cleaning up the code to an acceptable level, but otherwise leaving it alone.

michaeltaranto: Correct, this is centralising what already exists for all the candidate products (excluding CR). Iterating on the responsive design once in the style guide will address these break points.

mengtzu:
🎨 🚀 👣

PR #137, opened 2017-02-24
feat(header): Release header components

markdalgleish:
The previous release failed due to an invalid scope in the commit message. The previous commit also accidentally included a yarn lockfile, so figured this is a good opportunity to clean it up at the same time.


mengtzu:
🎨 :village burger:

PR #138, opened 2017-02-27
fix(header): Fix username class & remove divider responsively

michaeltaranto:
Responsively we drop off the employer site link, but we were leaving the divider behind.
screen shot 2017-02-27 at 11 41 45 am

Also noticed that the user name wasnt being truncated due to a class name not matching.

### DIFF
--- a/react/Header/UserAccount/UserAccount.less
+++ b/react/Header/UserAccount/UserAccount.less
@@ -65,15 +65,16 @@
   bottom: 0;
 }
 
-.userDisplayName {
+.userName {
   pointer-events: none;
   overflow: hidden;
   text-overflow: ellipsis;
   display: inline-block;
-  max-width: 101px; // ~ 10 characters
+  white-space: nowrap;
+  max-width: 100px; // ~ 10 characters
 
   @media only screen and (min-width: 400px) {
-    max-width: 201px; // ~ 20 characters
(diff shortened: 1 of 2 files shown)


react/Header/UserAccount/UserAccount.less
```diff
@@ -65,15 +65,16 @@
...
+.userName {
   pointer-events: none;
   overflow: hidden;
   text-overflow: ellipsis;
   display: inline-block;
-  max-width: 101px; // ~ 10 characters
+  white-space: nowrap;
+  max-width: 100px; // ~ 10 characters
 
   @media only screen and (min-width: 400px) {
```
mengtzu: Might be preferable to use `@responsive-breakpoint` and then a scaling width below that (ie you get somewhere between 10 and 20 characters as you approach 740px)?

michaeltaranto: Going forward we will be using standardised breakpoints for these components. This is simply correcting some oversights from the last PR. I would prefer to leave the breakpoint changes to the redesign of the small device header which is upcoming.

mengtzu: That is totally OK so long as you ritually apologise for incurring design debt :trollface:

mengtzu:
🎨

PR #139, opened 2017-02-27
feat(button): Add transparent style for Button component

georgespyropoulos:
The new seek profile page has some instances of buttons that look like links.  Rather than using the touchableText mixin to achieve this style, we thought it'd be great if we could have a button that looks like a link instead.  Here's an example of the desired style (clear and cancel buttons):

screen shot 2017-02-27 at 3 23 13 pm


mengtzu:
🎨 

Great!  We need this in CR too

georgespyropoulos:
@markdalgleish I've pushed up the changes we discussed earlier

markdalgleish:
LGTM—awesome work.

👏 🎨 📦 🚀

PR #140, opened 2017-02-27
fix(navigation): Pluralise 150k navigation label

michaeltaranto:
A small oversight in the extraction of the `<Header />`, we accidentally made the "$150k Jobs" navigation item singular.


mengtzu:
🎨

Issue #142, opened 2017-02-28
Button groups


markdalgleish:
We don't currently have a defined way of grouping a row of buttons together.

markdalgleish:
Closing for now.

PR #143, opened 2017-02-28
feat(InputGroup): New component for grouping and spacing a set of inputs

georgespyropoulos:
ButtonGroup component for adding spacing between two or more buttons that sit in a row.

https://github.com/seek-oss/seek-style-guide/issues/142


react/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,17 @@
+import styles from './ButtonGroup.less';
+import React, { PropTypes } from 'react';
+
+export default function ButtonGroup({ children }) {
+  return (
+    <div>{children.map(child => React.cloneElement(child, { className: styles.button }))}</div>
```
markdalgleish: Why are we cloning the element rather than wrapping it?

markdalgleish:
Admittedly the title of my issue was misleading here, but we need to be careful about our assumptions—we often want to group more than just buttons. For example, I've seen us place a checkbox next to a submit button. It'd be worth running this pattern past some designers first.

react/ButtonGroup/ButtonGroup.js
georgespyropoulos: I did initially wrap it but adding an extra element to achieve some extra margin felt a bit wrong.  I am happy to change it back if that's everyone's preferred approach?

markdalgleish: The issue with this implementation is that you might potentially clash with margins on child elements, whereas container elements don't have to worry about this.

georgespyropoulos: Yes true but the only thing this component does right now is add margin so is this necessarily a bad thing?

markdalgleish: Not right now, but it increases the likelihood of a bug creeping in as we iterate on this component.

georgespyropoulos:
Hmmmm... something like a checkbox next to a button is a common pattern however I am not sure we can predict all the combinations of elements that may need to sit next to each other.  Perhaps we better off allowing people to set their own spacing?

markdalgleish:
@georgespyropoulos sorry for letting this PR collect dust. We should schedule some time to catch up with @stephaniemoss to talk through this further.

georgespyropoulos:
Yes would be good to chat about this.  Let me know what time works for you.

stephaniemoss:
We could do it between 2-3 today at VDC if that suits?

markdalgleish:
@georgespyropoulos is it okay if we close this and revisit it more holistically later?

georgespyropoulos:
Yeah I think that would be best.  The requirements really got complicated out after we had a meeting on this.

Here's some notes I took at that meeting about what the InputGroup component should do:

- Space any input with 20px horizontally
- Collapse down to a vertical stack on mobile… 9px gap.
- Next role mobile has some good examples of edge cases we need to cover
- Need to be able to configure for many different options - stack, don’t stack.  
- General approach should be to provide defaults and allow overrides for edge cases

markdalgleish:
Cool. I'm keen to address this, but it's probably worth trialling it in a real app before getting it into the style guide.

PR #144, opened 2017-03-01
feat(FieldLabel): Added concept of FieldLabel

dborysiewicz:
Added concept the of FieldLabel and switched all inputs (besides Checkbox) to utilize this, behind the scenes it uses Text (which is responsive at this time 14px -> 18px <740px) but also sets text to bold <740px. 

One downside was this is not usable in Checkbox however the same rules (14px -> 18px & bold for <740px) has been applied for consistency.

### DIFF
--- a/react/fields/Checkbox/Checkbox.less
+++ b/react/fields/Checkbox/Checkbox.less
@@ -16,11 +16,15 @@
 }
 
 .standard {
-  .touchableText(@standard-type-scale);
+  .touchableText(@standard-type-scale-mobile);
 
   cursor: pointer;
   display: flex;
   align-items: center;
+
+  @media only screen and (min-width: (@responsive-breakpoint)) {
+    .touchableText(@standard-type-scale);
+  }
 }
 
 .checkBox {
@@ -56,7 +60,7 @@
 }
 
 .button {
-  .touchableText(@standard-type-scale);
+  .touchableText(@standard-type-scale-mobile);
 
   display: flex;
   align-items: center;
@@ -66,6 +70,10 @@
   border: 1px solid @sk-mid-gray-light;
   border-radius: 2px;
   cursor: pointer;
+
+  @media only screen and (min-width: (@responsive-breakpoint)) {
+    .touchableText(@standard-type-scale);
+  }
 
   .input:checked + .label & {
     font-weight: @sk-bold;
--- a/react/fields/Dropdown/Dropdown.js
+++ b/react/fields/Dropdown/Dropdown.js
@@ -6,6 +6,7 @@ import classnames from 'classnames';
 import ChevronIcon from '../../icons/ChevronIcon/ChevronIcon';
 
 import FieldMessage from '../FieldMessage/FieldMessage';
+import FieldLabel from '../FieldLabel/FieldLabel';
 
 function combineClassNames(props = {}, ...classNames) {
   const { className, ...restProps } = props;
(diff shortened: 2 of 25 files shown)


react/fields/Checkbox/Checkbox.less
```diff
@@ -16,11 +16,15 @@
 }
 
 .standard {
-  .touchableText(@standard-type-scale);
+  .touchableText(@standard-type-scale-mobile);
```
michaeltaranto: 🙌

react/fields/Dropdown/Dropdown.js
```diff
@@ -70,6 +50,7 @@ export default class Dropdown extends Component {
         return new Error(`\`inputProps.id\` will be overridden by \`id\` in ${componentName}. Please remove it.`);
       }
     },
+    /* eslint-enable consistent-return */
```
michaeltaranto: Nice catch!

react/fields/Dropdown/Dropdown.js
```diff
@@ -178,7 +137,7 @@ export default class Dropdown extends Component {
 
     return (
       <div className={classNames}>
-        {this.renderLabel()}
+        <FieldLabel {...this.props} />
```
michaeltaranto: I think in the case of `FieldLabel` and `FieldMessage` we shouldnt be blindly spreading props, rather cherry picking what should flow through to them. Same probably goes for the input/select renderer, we should only be spreading the not-message and not-label related props. Thoughts?

dborysiewicz: The idea behind this is that the prop validation (special rules around label, id & labelProps, etc) can live inside FieldLabel. Happy to look at alternatives if this is still possible?

react/fields/FieldLabel/FieldLabel.less
```diff
@@ -0,0 +1,11 @@
+@import (reference) "~seek-style-guide/theme";
+
+.label {
+  @media only screen and (max-width: (@responsive-breakpoint - 1)) {
+    font-weight: @sk-bold;
+  }
+}
+
+.secondaryLabel {
+  font-weight: 300;
```
mengtzu: Shouldn't this either be an existing weight, or we should be creating a new variable for it?

react/fields/Dropdown/Dropdown.js
michaeltaranto: Sure, the validation I understand living in the component that cares. I'm not sure about blindly passing everything though. @markdalgleish you have any thoughts on this?

react/fields/FieldLabel/FieldLabel.less
```diff
@@ -0,0 +1,11 @@
+@import (reference) "~seek-style-guide/theme";
+
+.label {
+  @media only screen and (max-width: (@responsive-breakpoint - 1)) {
+    font-weight: @sk-bold;
+  }
+}
+
+.secondaryLabel {
+  font-weight: @sk-light;
```
michaeltaranto: Has this been validated by the design team? Is a `secondaryLabel` the same weight regardless of screen size, even though the `label` is changing weight?

react/fields/TextField/TextField.less
```diff
@@ -5,12 +5,6 @@
   position: relative;
 }
 
-.label {
-  .standardText();
-  display: inline-block;
-  margin-bottom: @grid-row-height;
-}
```
michaeltaranto: 👋

react/fields/Dropdown/Dropdown.js
markdalgleish: I guess the middle ground here is we pluck the props that the `FieldLabel` cares about and it should still work, right?

markdalgleish: Sorry, that was a little unclear, I mean like this:
```js
const { foo, bar, baz } = this.props;
<FieldLabel {…{ foo, bar, baz }} />
```

react/fields/FieldLabel/FieldLabel.less
dborysiewicz: secondaryLabel is really a internal/private implementation used for the max character wording in the TextArea however I will run it by Steph.

react/fields/Dropdown/Dropdown.js
dborysiewicz: If you explicitly grab the props required for FieldLabel you have to have the matching propType validation in every single component that is using FieldLabel :(

react/fields/FieldLabel/FieldLabel.less
dborysiewicz: @stephaniemoss when you get a chance can you let us know about this? In this context secondaryLabel is used in TextArea when `Max Character` options is passed in, example at: http://seek-style-guide--53438087bf8ed17f0a3d3cbaacc79035f261ebf2.surge.sh/textarea

stephaniemoss: Let's stick with the default way it currently is now, I don't want to be kind of half tweaking this change on individual components. It really should be formed as patterns on a higher level.

Thanks!

react/fields/Dropdown/Dropdown.js
```diff
@@ -135,12 +135,15 @@ export default class Dropdown extends Component {
       [className]: className
     });
 
+    // eslint-disable-next-line react/prop-types
+    const { id, label, labelProps, secondaryLabel, invalid, help, helpProps, message, messageProps } = this.props;
+
     return (
       <div className={classNames}>
-        <FieldLabel {...this.props} />
+        <FieldLabel id={id} label={label} labelProps={labelProps} secondaryLabel={secondaryLabel} />
```
markdalgleish: I'd write this as `<FieldLabel {...{ id, label, labelProps, secondaryLabel }} />`

theme/type/type.less
```diff
@@ -44,5 +44,6 @@
 @interaction-type-scale: 1.8;
 
 // Text variants
+@sk-light: 300;
```
markdalgleish: Should we be documenting this?

react/fields/FieldLabel/FieldLabel.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    const { labelProps, id } = this.props;
+    const allLabelProps = {
+      ...combineClassNames(labelProps, styles.label),
+      ...(id ? { htmlFor: id } : {})
+    };
+    return (
+      <label {...allLabelProps}>
+        <Text>{label} {this.renderSecondary()}</Text>
```
markdalgleish: 🙌

markdalgleish:
Seriously awesome work consolidating all the labels into a single component. As discussed, even if we iterate on the design, the amount of work to do so will now be trivial 👏 

🎨 📦 🚀

PR #145, opened 2017-03-02
feat(text): Removal of nested span in Text

dborysiewicz:
This removes the nested span (while still retaining a central source of style) when using modifiers such as positive, critical, secondary, strong on Text as it was unexpected and not easy to modify the CSS of that inner span.

**Example**
Input before change:
`<Text positive className="foo">Hello</Text>`
Output before change:
`<div className="foo"><span>Hello</span></div>`

Input after change:
`<Text positive>Hello</Text>`
Output after change:
`<div className="foo">Hello</div>`

### DIFF
--- a/react/Text/Text.js
+++ b/react/Text/Text.js
@@ -2,21 +2,10 @@ import styles from './Text.less';
 import React, { PropTypes } from 'react';
 import classnames from 'classnames';
 
-import Secondary from '../Secondary/Secondary';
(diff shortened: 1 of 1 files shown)


markdalgleish:
LGTM, although I'm keen to hear @michaeltaranto's take on this.

markdalgleish:
🎨 📦 🚀

PR #146, opened 2017-03-02
feat(less): Adjust header and footer breakpoint

mengtzu:


mengtzu:
The Header and Footer components currently break between large and small screen presentations at `768px`, while other components break at the defined `@responsive-breakpoint`of `740px`.  This can look awkward if the window width is in the intervening 26px, particularly on Company Reviews.

However both components actually behave well down to `740px`.  As a bridge to a full responsive treatment, this PR ties the components to `@responsive-breakpoint`.  This would allow Company Reviews in particular to use these components rather than keep our bespoke versions and fix them in turn for life at 740px.

stephaniemoss:
I recommend 740px because 768px is tied to a specific Apple device which may or may not change in the future. 740px is at least in isolation of a device.

markdalgleish:
LGTM 👏🎨📦🚀

PR #148, opened 2017-03-07
feat(header): Use strings for authentication state

michaeltaranto:
BREAKING CHANGE: Moved away from true/false/null for tri-state management of `authenticated` on the `<Header />`. Now uses explicit strings of `authenticated`/`unauthenticated`/`pending` being passed into `authenticationStatus`.

### DIFF
--- a/react/Header/Header.js
+++ b/react/Header/Header.js
@@ -16,9 +16,13 @@ const employerLinkHref = locale => locale === 'NZ' ?
   'https://talent.seek.co.nz/' :
   'https://talent.seek.com.au/';
 
+const AUTHENTICATED = 'authenticated';
+const UNAUTHENTICATED = 'unauthenticated';
+const AUTH_PENDING = 'pending';
+
 export default function Header({
   locale,
-  authenticated,
+  authenticationStatus,
   userName,
   linkRenderer,
   activeTab,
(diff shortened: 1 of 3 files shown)


react/Header/Header.js
```diff
@@ -16,17 +16,22 @@ const employerLinkHref = locale => locale === 'NZ' ?
   'https://talent.seek.co.nz/' :
   'https://talent.seek.com.au/';
 
+const AUTHENTICATED = 'authenticated';
```
pokle: Can we have these constants exported?

markdalgleish: I'm not sure it's worth the effort, given we're using `propTypes`, and the API would be a bit awkward. The best I can come up with as a scalable pattern would look something like this, which I'm not a fan of:

```js
import { Header } from 'seek-style-guide/react';
const { AUTHENTICATED, UNAUTHENTICATED, PENDING } = Header.authenticatonStatus;
```

michaeltaranto: I looked into exporting these constants but it required a fair bit of boilerplate for the benefit returned (in terms of wiring up @markdalgleish's code above to the PropTypes). I would prefer this to go in consistently for now, but we discuss how to manage this pattern going forward

pokle: Agreed, not worth the effort.

PR #150, opened 2017-03-08
fix(monthpicker): Fix stacking order of native month picker chevron

georgespyropoulos:
There is an unnecessary `z-index: 2` on the month pickers chevron that is causing issues with elements that overlap the chevron.

This fixes a bug we are seeing in showcase for the new profile page:

screen shot 2017-03-08 at 2 18 24 pm


markdalgleish:
lgtm 🙌 

🎨 📦 🚀

PR #151, opened 2017-03-09
feat(FieldMessage): Allow message to be of type node

conshow:


react/fields/FieldMessage/FieldMessage.js
```diff
@@ -19,7 +19,8 @@ export default class FieldMessage extends Component {
     valid: PropTypes.bool,
     message: React.PropTypes.oneOfType([
       React.PropTypes.string,
```
markdalgleish: You should be able to remove this line now, since a string is also a valid node.

markdalgleish:
lgtm :clap:

When merging, don't forget to delete the commit description with all the individual commit details. The PR title is enough detail for this one :)

🎨📦🚀

PR #152, opened 2017-03-09
fix(Header): Use classic font stack for header and footer

markdalgleish:
This change also includes demo pages for the Header and Footer components.


markdalgleish:
Note that, after talking to @michaeltaranto, I realised that the standardised header and footer enforce Roboto, even though not all applications have migrated for performance reasons. Until we've completed standardising on Roboto, we're opting to enforce the classic font stack in the header and footer. This is a much bigger change than the original title of this PR, so I've renamed it.

michaeltaranto:
:art:

Issue #31, opened 2016-07-05
Buttons loading indicator doesn't work in Firefox


moroshko:
To display a loading indicator in `PinkButton` and `BlueButton`, we rely on animating the `content` property. Although this works in Chrome, this is not part of the CSS spec, and isn't supported by other browsers. [This article](https://css-tricks.com/animating-the-content-property/) sheds more light on this. We need to come up with a solution that is supported in all major browsers.

PR #153, opened 2017-03-10
fix(Button): Ensure loading indicator works cross-browser

adenj:
Changed how the loading indicator was styled, as the previous method was not compatible with
Safari/Firefox/IE

fixes https://github.com/seek-oss/seek-style-guide/issues/31


markdalgleish:
lgtm 👏 

When you merge it, can you delete the description, and make the title: `fix(Button): Ensure loading indicator works cross-browser (#153)`

🎨 📦 🚀

PR #154, opened 2017-03-10
feat(components): Flatten folder structure for public components

dborysiewicz:
Flattening our file structure as much as possible so that style guide consumers can import individual component files directly. In effect, this now makes the folder structure part of our public API. Internal components such as 'ClearField' under 'TextField' stays the same.


markdalgleish:
Note that if this doesn't really have a scope, you can leave it out, e.g. `feat: Flatten file structure`

markdalgleish:
That said, I'd probably call it `feat(components): Flatten folder structure`

markdalgleish:
Thanks so much for doing this work, this is going to have a huge impact on our bundle sizes when leveraged properly 👏 

When merging, can you use the revised title of this PR?

I might follow this up with some documentation on how to hook this up with https://www.npmjs.com/package/babel-plugin-transform-imports.

🎨 📦 🚀

PR #158, opened 2017-03-14
feat(Button): Make transparent buttons responsive

mattsjones:
This PR makes transparent style buttons have responsive text. However, the button will keep it's 5 grid row hit area on all viewport sizes.

The main issue we have here is if the responsive type rules are updated then this rule will not be updated with it as it uses the `touchableType` mixin. I'm thinking we may need an internal mixin for these scenarios that use touchable/raw text mixins but still need the font size to update responsively.


markdalgleish:
@mattsjones just a heads up, I've added a commit to this PR which adds a transparent button demo. Went to the static preview expecting to see it, but realised we must have missed this one.

markdalgleish:
lgtm 👍 

Can you make the commit message `feat(Button): Make transparent buttons responsive` and delete the description?

🎨 📦 🚀

PR #160, opened 2017-03-15
docs(build): Render docs with static site crawler

markdalgleish:
Using the latest version of static-site-generator-webpack-plugin, we no longer need to manually maintain a list of paths to render. Instead, we can automatically crawl all relative links from the two entry points we have currently.

Once we formalise the playground into a pattern library, we won't even need to provide paths at all since all pages with be crawlable from the root path.

### DIFF
--- a/webpack.static.render.config.js
+++ b/webpack.static.render.config.js
@@ -18,22 +18,6 @@ const appPaths = [
   path.resolve(__dirname, 'wip_modules')
 ];
 
-const routes = [
-  '/',
-  '/playground',
-  '/buttons',
-  '/textfields',
-  '/autosuggest',
-  '/textarea',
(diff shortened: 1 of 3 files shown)


webpack.static.render.config.js
```diff
@@ -18,22 +18,6 @@ const appPaths = [
...
-  '/textarea',
-  '/dropdown',
-  '/checkbox',
-  '/monthpicker',
-  '/typography',
-  '/dropdown',
-  '/header',
-  '/footer'
-];
-
```
michaeltaranto: 👏 🎉 👋 

(@mattsjones ☝️)

michaeltaranto:
:art:

PR #161, opened 2017-03-16
feat(Header): Update responsive header design

michaeltaranto:
This is a first pass at improving the small device navigation experience. Still a few aspects to be refined and maybe some CSS refactoring too but felt it important to socialise a preview of the changes sooner.


react/Header/Navigation/Navigation.js
```diff
@@ -79,7 +79,7 @@ Navigation.propTypes = {
   locale: PropTypes.string.isRequired,
   linkRenderer: PropTypes.func.isRequired,
   divider: PropTypes.bool.isRequired,
-  activeTab: PropTypes.oneOf(items.map(({ name }) => name))
+  activeTab: PropTypes.string
```
michaeltaranto: Given the small device menu doesn't share this component, this line could be reverted right?

react/Header/SignInRegister/SignInRegister.js
```diff
@@ -19,7 +16,7 @@ export default function SignInRegister({ linkRenderer, returnUrl }) {
       {
         linkRenderer({
           'data-analytics': 'sign-in',
-          href: `/Login/Standalone${urlSuffix}`,
+          href: appendReturnUrl('/Login/Standalone', returnUrl),
```
michaeltaranto: @dborysiewicz should we be updating the sign in & register url's to the new pages yet?

react/Header/Navigation/Navigation.js
markdalgleish: Nope, because if you mark a mobile-only tab as active, you get a proptype warning when it flows through this component.

react/Header/SignInRegister/SignInRegister.js
dborysiewicz: We are not 100% live yet so no need at this time :(

react/Header/UserAccount/UserAccount.less
```diff
@@ -42,19 +55,39 @@
   right: 0;
   transform-origin: 50% 0;
   transform: translateY(@grid-row-height * -3);
-  transition: opacity .25s cubic-bezier(0.020, 1.505, 0.745, 1.235), transform .25s cubic-bezier(0.020, 1.505, 0.745, 1.235);
+  transition: opacity .25s ease, transform .25s cubic-bezier(0.020, 1.505, 0.745, 1.235);
```
mengtzu: Do we use these animations anywhere else yet?  Might be nice to extract them into less variables at some point.

react/Header/UserAccount/UserAccount.less
```diff
@@ -84,9 +131,22 @@
   transition: transform .3s ease;
   transform-origin: 50% 50%;
   pointer-events: none;
+  .smallDeviceOnly({
+    text-align: center;
+    width: 20px;
```
mengtzu: I assume this is a magic number because while it's coincidentally the same as `grid-gutter-width` it doesn't have any real relation to the grid dimensions?

react/Header/UserAccountMenu/UserAccountMenu.js
```diff
@@ -1,88 +1,215 @@
...
+                key="icon"
+                className={classnames(styles.icon, styles.profile)}
+                svgClassName={styles.iconSvg}
+              />
             ]
           })
         }
       </li>
-      <li>
+      <li className={classnames(styles.aboveSmallDevice, activeTab === 'Saved Searches' && styles.activeTab)}>
```
mengtzu: Does feel like the strings for tab names should be constants somewhere

react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -1,41 +1,139 @@
...
   width: @grid-gutter-width;
   text-align: center;
+
+  .smallDeviceOnly({
+    margin-left: (@grid-gutter-width / 2);
+  });
+}
+
+.icon.jobSearch { color: @sk-highlight; }
+.icon.profile { color: #9556B7; /* sk-purple? */ }
```
mengtzu: Did we run out of colours?

I don't like a straight hex being in the style guide, if that's the profile club colour let's actually make that `sk-purple`.

Also where is the Company Reviews icon with its majestic generic building? :)

react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -1,41 +1,139 @@
...
+.icon.saveJobs { color: @sk-pink; }
+.icon.saveSearches { color: @sk-highlight; }
+.icon.recommendedJobs { color: @sk-green-light; }
+
+.iconSpacer {
+  margin-left: (@grid-gutter-width / 2);
+  width: 20px;
```
mengtzu: Again, coincidental similarity?  It's kind of weird seing `20px` next to `{20px as variable} / 2`

react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -1,41 +1,139 @@
...
 
 .iconSvg {
   width: 14px;
   height: 14px;
+  // Begin bold icon hacks
+  .smallDeviceOnly({
+    overflow: visible;
+    path {
+      stroke: currentColor;
+      stroke-width: 70px;
```
mengtzu: If this is any kind of standard for icons this is probably just as good a candidate for a less variable as font sizes.

mengtzu: (let me guess, being inside a `//hack` comment block indicates it isn't any kind of standard :))

theme/layout/breakpoints.less
```diff
@@ -1,2 +1,20 @@
...
+  }
+}
+
+.aboveSmallDevice(@rules) {
+  @media only screen and (min-width: @responsive-breakpoint) {
+    @rules();
+  }
+}
+
+.classicOnly(@rules) {
```
mengtzu: This is kind of a weird name, I assumed it had something to do with the classic font stack at first.

Not sure it's meaningful outside of Candidate.  Typically use `largeScreen` or similar in hirer.

react/Header/UserAccount/UserAccount.less
markdalgleish: Agreed, but this is something we'll need to revisit across the board.

react/Header/UserAccount/UserAccount.less
markdalgleish: It is meant to be aligned to the grid, so I'll update this to use the variable. We're looking at addressing icons as a whole soon, but this is still a worthy fix in the meantime.

react/Header/UserAccountMenu/UserAccountMenu.js
markdalgleish: IMO it's overkill if we don't have a static typing system, given there's very little harm in getting it wrong.

markdalgleish: Not against discussing static types, but that's a much larger discussion 😄

react/Header/UserAccountMenu/UserAccountMenu.less
markdalgleish: Good catch, forgot to clean this up. I'll add this to the list of brand colours.

react/Header/UserAccountMenu/UserAccountMenu.less
markdalgleish: Looking to standardise this soon, while cleaning up our icons more generally, but didn't want to blow out the scope of this already massive task 😄

theme/layout/breakpoints.less
markdalgleish: Agreed, this name isn't great. @michaeltaranto thoughts?

markdalgleish: I realised that this mixin is only used in `Header.less`, so I'm making it local to that file, and renaming it to `classicDesktopOnly`.

markdalgleish:
Alright, I think we're pretty much ready to release this. Even though this isn't technically a breaking change, I'm planning to make it a major version bump since it's such a significant change to the UI. Any last minute questions or objections?

PR #162, opened 2017-03-20
feat(Text): Update responsive type values

mattsjones:


theme/type/responsiveType.less
```diff
@@ -9,13 +9,23 @@
...
   @row-span: "@{variant}-type-row-span";
+  @weight: "@{variant}-type-weight";
   @mobile-type-scale: "@{type-scale}-mobile";
   @mobile-row-span: "@{row-span}-mobile";
+  @mobile-weight: "@{variant}-type-weight-mobile";
 
   @media only screen and (max-width: @max-breakpoint) {
     .__baselineType(@@mobile-type-scale, @@mobile-row-span);
+
+    & when (not(@@mobile-weight = 400)) {
```
mengtzu: Can you unpack the logic on this one? 400 is normal, but is there a reason not to double apply it?

(also good work to CSS/Less on that `=` syntax, why work the way a dev would expect you to?)

mattsjones: Hmm not really an issue, just thought to only apply the style if necessary. I don't have an issue always applying the 400.

markdalgleish:
I've arguably misused `header` text for the intro paragraph on the style guide home page, which now goes bold on mobile:

screen shot 2017-03-20 at 12 36 05 pm

What should we do about this? Is it worth introducing a `normal` weight variant for text, along the lines of `strong`? I can see this being needed sometimes in an application context, where headings are of variable length, typically with user-provided content.

theme/type/responsiveType.less
mengtzu: Favour simpler code IMO.  Setting `400` to `400` is harmless so may as well save the developer brain cycles, `LESS` won't care :)

react/MonthPicker/MonthPicker.demo.js
```diff
@@ -61,6 +62,7 @@ export default {
         }
       ]
     },
-    ...fieldMessageOptions
+    ...fieldMessageOptions,
+    ...fieldLabelOptions
```
markdalgleish: `Textarea` needs one of these too, I believe.

react/private/FieldLabel/FieldLabel.demo.js
```diff
@@ -0,0 +1,15 @@
+export default [
+  {
+    label: 'Secondary label',
+    type: 'checklist',
+    states: [
+      {
+        label: 'Secondary label',
+        transformProps: props => ({
+          ...props,
+          secondaryLabel: '(Max 30 characters)'
```
markdalgleish: 👏

react/private/FieldLabel/FieldLabel.js
```diff
@@ -69,9 +69,9 @@ export default class FieldLabel extends Component {
     }
 
     return (
-      <span className={styles.secondaryLabel}>
+      <Secondary>
```
markdalgleish: 👏 🎉

react/private/FieldLabel/FieldLabel.less
```diff
@@ -1,11 +0,0 @@
...
-
-.label {
-  @media only screen and (max-width: (@responsive-breakpoint - 1)) {
-    font-weight: @sk-bold;
-  }
-}
-
-.secondaryLabel {
-  font-weight: @sk-light;
-}
```
markdalgleish: 👏 🎉

react/MonthPicker/MonthPicker.demo.js
mattsjones: Yeah I thought so but it doesn't work correctly currently as it has a custom `maxCharacters` prop. We should probably roll that back in favour of secondary label but don't think this is the PR for it.

PR #165, opened 2017-03-23
feat(AsidedLayout): Add interface for className for aside div

dborysiewicz:
Need interface to be able to apply a class to the aside div, screenshot shows use case for where I need to add `align-self: flex-end;`.
screen shot 2017-03-23 at 1 40 52 pm


markdalgleish:
I'm wondering whether a more consistent API might be something like `asideProps`

PR #166, opened 2017-03-23
feat(Header): Add returnUrl for unauthenticated links

mengtzu:
Accepts optional string prop `returnUrl` on the `Header` component.  If present, appends it (URI encoded) to the href for the Sign in and Register links as a query string param `?returnUrl={returnUrl}`


react/Header/SignInRegister/SignInRegister.js
```diff
@@ -4,7 +4,10 @@ import React, { PropTypes } from 'react';
 
 import ScreenReaderOnly from '../../ScreenReaderOnly/ScreenReaderOnly';
 
-export default function SignInRegister({ linkRenderer }) {
+export default function SignInRegister({ linkRenderer, returnUrl }) {
```
mattsjones: Can you add a prop type for `returnUrl`

mengtzu: Types are the cancer that is killing Javascript.

But sure :)

mengtzu: Done

markdalgleish:
@mengtzu can you add a snapshot test, or modify an existing one?

react/Header/Header.demo.js
```diff
@@ -16,7 +16,8 @@ export default {
     authenticationStatus: 'authenticated',
     userName: 'Jane Smith',
     divider: false,
-    linkRenderer: dummyLinkRenderer
+    linkRenderer: dummyLinkRenderer,
+    returnUrl: 'seek-oss.github.io/seek-style-guide'
```
markdalgleish: Super minor nitpick, but could you make this `/jobs`, since we don't support cross-domain return URLs?

mengtzu:
@markdalgleish Added a snapshot, also added a helper script to update them.  My first Jest, let me know if it's wrong.

package.json
```diff
@@ -7,6 +7,7 @@
     "start": "BASE_HREF=/ node server",
     "test": "npm run test-unit && npm run lint && npm run build",
     "test-unit": "BABEL_ENV=test jest --config ./test/jest.config.json",
+    "update-snapshot": "BABEL_ENV=test jest --config ./test/jest.config.json --updateSnapshot",
```
markdalgleish: To ensure this stays consistent, it's probably worth changing this to `npm run test-unit -- --updateSnapshot`

mengtzu: done

PR #167, opened 2017-03-28
fix(TextField): Fix TextField clear icon

markdalgleish:
The icon styles were adjusted via a class override, which causes rendering issues if the order of extracted CSS changes unexpectedly.


markdalgleish:
In this PR I also took the opportunity to upgrade the TextField demo to make it a controlled input with a clear icon.

PR #168, opened 2017-03-30
docs(site): Isolate documentation code in docs folder

markdalgleish:
While this is a worthy refactor by itself, this work is being done so that we can have multiple `dist` folders, opening up the possibility of compiling React components into a format suitable for non-React consumers.


mengtzu:
:art: 📚  :shipitparrot:

PR #171, opened 2017-04-04
fix(styles): Provide browser list to Autoprefixer

markdalgleish:
We're currently relying on the default browser support policy of Autoprefixer, instead of defining our own, as we would typically do in an application context. The list in this PR is taken from our core search product, but further changes may be required.


mengtzu:
:shipitparrot:

PR #169, opened 2017-04-04
feat(Header): Provide compiled header library for non-React consumers

markdalgleish:
This PR is pretty self explanatory, if you look at the diff for `README.md` 😄

PR #173, opened 2017-04-04
fix(dist): Add .npmignore so dist folder is published

dborysiewicz:
It is a copy of what was in .gitignore but swapped dist for docs so we can use https://github.com/seek-oss/seek-style-guide/pull/169.


markdalgleish:
@dborysiewicz just updated the name of the PR so that this fix is published to npm

PR #176, opened 2017-04-10
fix(components): Automatically add displayName to React components

markdalgleish:
Also refactored babel config into the 'config' folder, shared with webpack decorators.

I stumbled across this issue when I noticed that, seemingly since we fixed the minification step, our documentation now shows minified component names in many cases, e.g.

screen shot 2017-04-10 at 1 28 13 pm

### DIFF
--- a/config/babel.config.js
+++ b/config/babel.config.js
@@ -1,9 +1,11 @@
 module.exports = ({ reactHotLoader = false }) => ({
+  babelrc: false,
   presets: [
     'es2015',
     'react'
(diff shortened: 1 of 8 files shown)


markdalgleish:
You can see the fix when you compare [production](https://seek-oss.github.io/seek-style-guide/typography) to the [Surge preview](http://seek-style-guide--5d0c2d1e87cc57249b4680c2e95b6c82eb477624.surge.sh/typography).

PR #178, opened 2017-04-12
feat(Loader): Add Loader component

adenj:
Added the Loader component currently being used in Houston with a fade in animation.


docs/src/demoSpecs.js
```diff
@@ -21,5 +22,6 @@ export default [
   Dropdown,
   Rating,
   Header,
-  Footer
+  Footer,
+  Loader
```
markdalgleish: I'd probably move this higher up the list, maybe before `Header`?

PR #179, opened 2017-04-13
feat(Header): Display first part of user email if no username is present

adenj:
If no user name is on authenticated account the Header will display the first part of the users
email address

### DIFF
--- a/react/Header/Header.js
+++ b/react/Header/Header.js
@@ -33,7 +33,7 @@ export default function Header({
   const userClasses = classnames({
     [styles.user]: true,
     [styles.user_isReady]: authenticationStatus === UNAUTHENTICATED ||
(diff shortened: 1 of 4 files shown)


react/Header/Header.js
```diff
@@ -32,9 +33,11 @@ export default function Header({
   const userClasses = classnames({
     [styles.user]: true,
     [styles.user_isReady]: authenticationStatus === UNAUTHENTICATED ||
-      (authenticationStatus === AUTHENTICATED && userName)
+      (authenticationStatus === AUTHENTICATED && userName || userEmail)
```
michaeltaranto: Can you add parenthesis around `userName || userEmail`, otherwise it will match when even if the `authenticationStatus` is not `AUTHENTICATED`

PR #180, opened 2017-04-20
fix(Button): Remove focus ring from transparent button

michaeltaranto:
A transparent button should have the same focus behaviour as our links, however they are currently displaying like this:
screen shot 2017-04-20 at 10 09 26 am

### DIFF
--- a/react/Button/Button.less
+++ b/react/Button/Button.less
@@ -86,6 +86,10 @@
     transform: none;
   });
 
+  .focusState({
+    box-shadow: none;
+  });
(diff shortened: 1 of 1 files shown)


markdalgleish:
We shouldn't be removing the focus ring entirely—we should be altering its appearance to surround the text rather than the entire container.

michaeltaranto:
@markdalgleish agreed, we went a different way and didnt end up needing this now. We'll revisit when time permits.

PR #182, opened 2017-04-27
fix(Autoprefixer): Update the Autoprefixer configuration

georgespyropoulos:
Now using the latest Seek supported browsers for adding CSS vendor prefixes.

The update of CSS loader has been made to fix a bug where using the `minimize` option with CSS loader (which invokes [CSS nano](http://cssnano.co/)) will strip vendor prefixes from your CSS as an optimisation by default.  The latest version of CSS loader fixes this by turning this behaviour off within CSS nano.  Here's the relevant PR:  https://github.com/webpack-contrib/css-loader/pull/361.

This will fix flexbox issues that are happening in Safari 6-8

screen shot 2017-04-27 at 11 55 46 am


markdalgleish:
Interesting—the webpack build is failing, but the build is still marked as a success:

>Module build failed: BrowserslistError: Unknown version 57 of and_chr

@michaeltaranto I seem to remember that you've run into this webpack issue before too?

michaeltaranto:
@markdalgleish yeah this is a problem in webpack 1 where the build doesnt exit with status code 1 when there is an error (rectified in webpack 2). There is a plugin for this [webpack-fail-plugin](https://www.npmjs.com/package/webpack-fail-plugin), but we opted to roll our own because we wanted to fail fast and stop the build immediately. The plugin finishes the build but returns the correct status code of 1. Here is our implementation:

```
const chalk = require('chalk');

module.exports = function failWebpackBuildOnError() {
  let isWatch = true;

  this.plugin("run", function(compiler, callback) {
    isWatch = false;
    callback.call(compiler);
  });

  this.plugin("done", function(stats) {
    if (stats.compilation.errors && stats.compilation.errors.length) {
      console.error(chalk.red(stats.compilation.errors));
      if (!isWatch) {
        process.exit(1);
      }
    }
  });
};
```
Make sense?

markdalgleish:
@michaeltaranto cool—build speed isn't a big priority here, so we might just opt for webpack-fail-plugin 👍 

@georgespyropoulos could you add this as part of this PR? It'd be a good way to test that it's working, making sure that it fails the build.

georgespyropoulos:
@markdalgleish will do 👍

georgespyropoulos:
I've introduced the `failWebpackBuildOnError` module into the styleguide and it can sucessfully fail a build on the first error https://travis-ci.org/seek-oss/seek-style-guide/builds/226637548.

I updated Autoprefixer to the latest version to fix the `Module build failed: BrowserslistError: Unknown version 57 of and_chr` error.  The latest version of autoprefixer has a newer version of Browsers List which no longer contains this bug

PR #184, opened 2017-05-05
fix(FieldMessage): Add role and aria-live attributes

georgespyropoulos:
Adding `aria-live` and `role` attributes to the Field Message component to help screen readers announce this text.  Users will hear the text announced as it appears in the DOM.

`role="alert"` - instructs a screenreader to announce text as a priority and may interrupt other text from being read out.

`aria-live="polite"` -  instructs a screenreader to announce text at some point after it has finished what it is currently announcing.


react/TextField/TextField.js
```diff
@@ -88,11 +88,12 @@ export default class TextField extends Component {
   }
 
   renderInput() {
-    const { inputProps = {}, id } = this.props;
+    const { inputProps = {}, id, valid } = this.props;
     const { ref } = inputProps;
     const allInputProps = {
       ...combineClassNames(inputProps, styles.input),
       ...(id ? { id } : {}),
+      ...{ 'aria-invalid': valid === false },
```
dborysiewicz: Should we check to see if somebody has already manually set this?

georgespyropoulos: My thinking on this was that you probably wouldn't want to set `aria-invalid` to be out of sync with `valid`.  Can you think of a use case where you would need to set `aria-invalid` to something custom?

dborysiewicz: Good point and not really think of any case but was just thinking around what happened if somebody did have it already set and updated..... But I guess it would still all work just double set so not a huge deal?

georgespyropoulos: I just tested it and the spread ensures that if you pass in an `aria-invalid`, the prop inside the component will take precedence.

dborysiewicz:
I've approved buy maybe get another person from another team to review as well?

react/private/FieldMessage/FieldMessage.js
```diff
@@ -45,13 +45,19 @@ export default class FieldMessage extends Component {
...
       return (
         <Text
           {...restMessageProps}
-          critical={(valid === false && !secondary) || critical}
-          positive={(valid === true && !secondary) || positive}
-          secondary={typeof valid === 'undefined' || secondary}>
+          critical={isCritical}
+          positive={isPositive}
+          secondary={isSecondary}
+          {...(isCritical && !restMessageProps.role) && { role: 'alert' }}
```
mengtzu: I haven't seen this use of spread before.  I assume if the boolean on the left is true, it will spread the property in the literal on the right?

react/private/FieldMessage/FieldMessage.test.js
```diff
@@ -110,5 +110,37 @@ describe('FieldMessage', () => {
...
       });
     });
+
+    describe('aria attributes passed as messageProps', () => {
+      it('will set role="alert" for invalid input if no custom aria attributes are passed in', () => {
+        render(<FieldMessage valid={false} message="Something went wrong" messageProps={{ className: 'message' }} />);
+        expect(message.props.role).to.equal('alert');
+      });
+
+      it('will set role="alert" for invalid input if no custom aria attributes are passed in', () => {
```
mengtzu: This test isn't obviously asserting what its description states?

react/private/FieldMessage/FieldMessage.js
georgespyropoulos: Yes that's correct.  Just had a chat to Matt about it and he also finds this a little confusing to read so I am going to change it to a ternary and return an empty object in the case the condition on the left is false.

react/private/FieldMessage/FieldMessage.test.js
georgespyropoulos: Good catch, that was a sloppy copy / paste job.  I will fix it

tamoore:
`aria-live` is not the correct technique to use here.

Assuming this is a form field that has an associated message that describes it's state (i.e. error, hint, whatever), then the correct aria attribute is `aria-describedby`. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute

`aria-live` is actually probably going to be a little annoying to SR user because it won't have the context that the describedby attribute would.

georgespyropoulos:
I decided on `aria-live` because the text enters the DOM when there's a blur event on the text field.  The problem I was trying to solve is that this text is not currently being announced by screen readers because it only appears after you have navigated away from the field.  

If we change this to `aria-describedby`, will the text be read out once a blur event is set to the field?  It seems that that describedby is more suitable for text that is always present?  Perhaps we could use it for describing help text for form fields.

Here's an demo of the text fields and `aria-live` has been applied to a positive message. http://seek-oss.github.io/seek-style-guide/textfields

tamoore:
It should work on any event, you may have to change the `id` though (possibly a generated id) if the message changes. (Not sure about that). 

I've completely done the same thing before thinking that it was helping and found, to my horror, that it actually just creates confusion. A live region is primarily used for dynamic content updates. The problem with input fields is that the context when interacting with them is really important. A live region in JAWS will go something like:

'Blah blah blah. "Something went wrong"' (depends on `assertive`, `polite`)

Now without any visual context (that we have) that means completely nothing to an SR user. Something went wrong, with what?

Whereas with `describedby` it'll be:

'Edit Email Address (or whatever the label is), Something went wrong'. 
 
That's much more helpful because an SR user now knows what that message is referring to.

Happy to pair on this if you want to get it right, a11y is definitely a minefield of these problems.

markdalgleish:
@georgespyropoulos @tamoore what's the status on this?

georgespyropoulos:
@markdalgleish we're going to pair on it later today to try out `aria-describedby`

markdalgleish:
@georgespyropoulos any development on this?

dborysiewicz:
@georgespyropoulos & @tamoore not sure how you guys went but my 2cents on this would be to not touch/add `aria-describedby` & `aria-live` as it's possibly opening a big can of worms (from past experience) and I'm not sure if it's 100% required.... I would be in favour for just adding `role`.

georgespyropoulos:
Not yet, will have to make some time with Todd to discuss this again.  We tried last week but had some issues getting jaws running.  @dborysiewicz the goal of this was to help screen reader users find text that we're displaying in FieldMessage, particularly when there is an error.  We'll run though it and see what we can do to make things easier for screen reader users

markdalgleish:
Closing this for now. Discussed with @georgespyropoulos, we might re-open this one in the future, but it still needs some further discussion.

PR #185, opened 2017-05-09
Print styles for header and footer

kosanna:
- Restyled header for printing
- Remove footer for printing
- Add print visibility wrapper component

Kayla confirmed that footer and header should be consistently displayed/hidden across all the platforms thus doing it inside the components.

Related issue: https://github.com/SEEK-Jobs/chalice/issues/1702

screen shot 2017-05-08 at 4 45 00 pm


react/PrintVisibilityWrapper/PrintVisibilityWrapper.js
```diff
@@ -0,0 +1,41 @@
...
+
+export default class PrintVisibilityWrapper extends PureComponent {
+
+  static propTypes = {
+    className: PropTypes.string,
+    children: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.node),
+      PropTypes.node
+    ]).isRequired,
+    isScreenOnly: PropTypes.bool,
```
kosanna: Can do one required prop aka oneOf ['screenOnly', 'printOnly'], depends on personal and team preferences. Thoughts?

markdalgleish:
Thanks so much for opening your first PR 👏 🎉 

I have a couple of high level questions:
1. Does `PrintVisibilityWrapper` need to be exported? If possible, I'd prefer this to be a private component for now, especially since we've been floating the idea of introducing a more comprehensive component for hiding content based on media queries, e.g. `<Hidden desktop>`, `<Hidden print>` etc.
2. Is it okay if we delay the introduction of new React APIs until a later PR, possibly after the new header and footer are merged? I'd probably prefer a standalone PR for introducing `prop-types` and `PureComponent`, to make the upgrade path more obvious.

kosanna:
@markdalgleish Yep it needs to be exported because it could be great to use it in chalice as well, that basically was the incentive to create the component so we do not have the same styles everywhere for the same purpose.
2 - got ya, definitely, however would prefer using PureComponent in the new component straight away as it's better performance-wise, but can hold on changing footer and header though

kosanna:
@markdalgleish can do hidden component straight away in a separate PR, how does that sound?

PR #186, opened 2017-05-10
feat(Hidden): Add Hidden component

kosanna:


markdalgleish:
This is looking great! So excited to finally have a component like this 👏 

By the way, it looks like you've forgotten to export it from the main React bundle.

kosanna:
@markdalgleish good point, will do

react/Hidden/Hidden.less
```diff
@@ -0,0 +1,15 @@
...
+.print {
+  @media print {
+    display: none;
+  }
+}
+
+.screen {
+  display: none;
+
+  @media print {
```
markdalgleish: This should be `@media screen`, right?

kosanna: naw, it's exactly print. Because it's a `<Hidden screen>` use-case

markdalgleish: Oh right, yeah, sorry.

Could this be simplified to `@media screen { display: none; }` like the print styles?

PR #188, opened 2017-05-10
feat(FooterHeader): Print styles

kosanna:


react/Footer/Footer.js
```diff
@@ -6,13 +6,14 @@ import connect, { social } from './data/connect';
 import employers from './data/employers';
 import copyright from './data/copyright';
 
-import React, { Component, PropTypes } from 'react';
 import classnames from 'classnames';
+import PropTypes from 'prop-types';
```
markdalgleish: Sorry, could we roll this back until we coordinate a React upgrade in a separate PR?

kosanna: Yep, my bad, missed it

PR #190, opened 2017-05-10
fix(Header): Added position: relative; back to logo

dborysiewicz:
Added position: relative; back to logo to restrict the position: absolute; link.

### DIFF
--- a/react/Header/Header.less
+++ b/react/Header/Header.less
@@ -80,28 +80,32 @@
   }
 }
 
+.logo {
+  position: relative;
+
+  .logoSvg {
+    height: 39px;
+    width: 105px;
+    @media @large-ia {
+      height: 51px;
+      width: 129px;
+    }
+  }
+
+  .logoLink {
+    position: absolute;
+    top: 0;
+    text-indent: 100%;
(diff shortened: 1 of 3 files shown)


react/Header/Header.less
```diff
@@ -80,28 +80,32 @@
...
-  position: absolute;
-  top: 0;
-  text-indent: 100%;
-  overflow: hidden;
-  width: 100%;
-  height: 100%;
+.logoNote {
+  .standardText();
+  margin-left: 0.8rem;
+  margin-bottom: 0.35rem;
```
michaeltaranto: Can i ask what these numbers are? We dont really have/use magic numbers for spacing (i understand this is an aside to what you are fixing, but think it should be done as part of this tidy up)

kosanna: This is a proper code-review @michaeltaranto 👍 These are just adjusting the note to be aligned with the `seek` word in the logo, probably better to have it all just as an image asset all together 🤔

kosanna: Let's do the fix in a separate PR, closing this one

PR #192, opened 2017-05-12
docs(contributing): Update review policy

markdalgleish:


kosanna:
LGTM 🚀 May be can add something around merge commit naming as well? Or I'm missing it somewhere...

markdalgleish:
@kosanna there's already some content in there about commit messages, but we could definitely flesh it out further. Might do this in a later PR though.

PR #194, opened 2017-05-14
refactor(PropTypes): Switch all propTypes imports to prop-types package

kosanna:


markdalgleish:
@kosanna just noticed one component that hadn't been migrated, so figured I'd add another commit to this PR 😄

package.json
```diff
@@ -57,6 +57,7 @@
     "less": "2.7.1",
     "less-loader": "2.2.3",
     "postcss-loader": "0.9.1",
+    "prop-types": "^15.5.10",
```
georgespyropoulos: Should the `prop-types` package be a peer dependency?  There's been an ongoing debate about peer dependencies but at least for now I think it makes more sense for it to be a peer dependency since react is also one.

markdalgleish: I wondered the same thing, but Facebook recommend that you install it as a regular dependency: https://github.com/facebook/prop-types/blob/master/README.md#how-to-depend-on-this-package

georgespyropoulos: Right I see.  I guess it's best we go with the recommendation then

PR #195, opened 2017-05-15
feat(SlideToggle): Add SlideToggle component

georgespyropoulos:
Adding the toggle switch we are currently using in [applied jobs](https://www.seek.com.au/my-activity/saved-jobs) to the style guide because the showcase team is building a new page that will use the same switch.


markdalgleish:
This PR isn't being deployed to Surge for some reason 🤔

markdalgleish:
(Not just because the build is currently broken—the earlier successful commits didn't deploy)

georgespyropoulos:
I got the build passing again and found the surge url in the build log.  This one seems to work for me http://seek-style-guide--19ad2ab1d6e5f968ef0ae273a6e517939b293e79.surge.sh/slide-toggle.  Does it work for you @markdalgleish ?

markdalgleish:
@georgespyropoulos yep, that works now 👍

react/SlideToggle/SlideToggle.js
```diff
@@ -0,0 +1,72 @@
...
+
+  static defaultProps = {
+    position: RIGHT,
+    hideLabel: false
+  };
+
+  handleChange = event => {
+    const { onToggle } = this.props;
+    if (onToggle) {
+      onToggle(event.target.checked);
```
mattsjones: Not sure we need a custom function here. Should we just let the consuming app bind their own `onChange`? Or do we prefer the abstraction of `onToggle`?

react/SlideToggle/SlideToggle.less
```diff
@@ -0,0 +1,82 @@
...
+    background-color: @sk-highlight;
+    left: 21px;
+  }
+
+  &:checked + .slider .slideButton svg {
+    opacity: 1;
+  }
+}
+
+.label {
```
mattsjones: This class is empty. Assuming not required?

react/SlideToggle/SlideToggle.less
```diff
@@ -0,0 +1,82 @@
...
+  left: 10px;
+  top: 5px;
+  margin: 0;
+
+  &:focus + .slider {
+    box-shadow: 0px 0px 10px 0px @sk-focus;
+  }
+
+  &:active + .slider .slideButton {
+    transform: scale(1.15, 0.85);
```
mattsjones: Why are performing a scale here? This causes the switch to shrink when you click on it, that includes the icon which I feel looks a bit weird. What effect are we going for exactly?
From
image

To
image

react/SlideToggle/SlideToggle.js
```diff
@@ -0,0 +1,72 @@
...
+import TickIcon from '../TickIcon/TickIcon';
+import classnames from 'classnames';
+
+const LEFT = 'left';
+const RIGHT = 'right';
+
+export default class SlideToggle extends Component {
+  static propTypes = {
+    id: PropTypes.string.isRequired,
+    text: PropTypes.string.isRequired,
```
mattsjones: Could we call this `label` instead?

react/SlideToggle/SlideToggle.js
```diff
@@ -0,0 +1,72 @@
...
+        )}>{text}</span>
+    );
+  }
+
+  render() {
+    const { children, id, checked, position, text, hideLabel } = this.props;
+    return (
+      <label htmlFor={id} className={styles.root}>
+        {(hideLabel !== true && position === LEFT && !children) && this.renderLabel()}
+        {position === LEFT && children}
```
mattsjones: Is there a scenario when we would need to render children here? I would leave this out personally unless we're actually using it.

react/SlideToggle/SlideToggle.less
```diff
@@ -0,0 +1,82 @@
+@import (reference) "~seek-style-guide/theme";
+
+.root {
+  display: inline-flex;
```
mattsjones: Not sure how well IE 11 plays with `inline-flex`. I'd give it a spin before merging.

react/SlideToggle/SlideToggle.less
```diff
@@ -0,0 +1,82 @@
...
+  line-height: @grid-row-height * 3;
+  height: @grid-row-height * 3;
```
mattsjones: This element is only 3 grid rows high. Should it have a touchable area of 5 to help on mobile? Doesn't need to actually be that big IMO though.

react/SlideToggle/SlideToggle.js
```diff
@@ -0,0 +1,72 @@
...
+    const { onToggle } = this.props;
+    if (onToggle) {
+      onToggle(event.target.checked);
+    }
+  }
+
+  renderLabel = () => {
+    const { position, text } = this.props;
+    return (
+      <span
```
mattsjones: We should probably use the `FieldLabel` component here so it corresponds to the text and label standards

react/SlideToggle/SlideToggle.less
georgespyropoulos: Yes not required, I'll delete it

react/SlideToggle/SlideToggle.less
georgespyropoulos: I took the idea from this toggle https://gimme-houston.cloud.seek.com.au/.  I'm happy to go with the majority on whether the animation adds anything.  @stephaniemoss do you have an opinion on this?  Demo page is here http://seek-style-guide--19ad2ab1d6e5f968ef0ae273a6e517939b293e79.surge.sh/slide-toggle/

react/SlideToggle/SlideToggle.less
georgespyropoulos: Yes that's a good point.  I will increase the touchable area to 5 grid rows

markdalgleish:
Haven't reviewed this, but don't forget to use the new prop-types library instead of React.PropTypes.

react/SlideToggle/SlideToggle.js
georgespyropoulos: I thought it might be a good way to allow people to pass in label with a custom style.  Perhaps a prop for a class on the label might be a better option?

react/SlideToggle/SlideToggle.js
georgespyropoulos: Good idea, I will update to use that component

react/SlideToggle/SlideToggle.js
georgespyropoulos: Yes, will update

react/SlideToggle/SlideToggle.js
mattsjones: Personally, I think less flexible is better and we can add functionality if need be. It keeps everything being less complex. Plus `FieldLabel` caters to most of what you're talking about I think.

markdalgleish:
What about animating the circle but leaving the icon alone? I agree with @mattsjones that the squashed tick looks a little weird.

react/SlideToggle/SlideToggle.less
markdalgleish: What about animating the circle but leaving the icon alone? I agree with @mattsjones that the squashed tick looks a little weird.

georgespyropoulos: Yeah I'll give that a try.  I never intended to animate the icon - you just get the scaling on child elements

react/SlideToggle/SlideToggle.js
georgespyropoulos: I've refactored this to make it simpler - am now binding the passed in `onToggle` function directly to the inputs `onChange` without a custom function in the middle.

react/SlideToggle/SlideToggle.less
georgespyropoulos: I've updated this so the animation does not apply to the icon

react/SlideToggle/SlideToggle.test.js
```diff
@@ -39,12 +39,12 @@ describe('Slide toggle:', () => {
...
         id="testToggle"
-        text="Test toggle"
+        label="Test toggle"
         onToggle={onToggle}
       />
     );
     const input = findRenderedDOMComponentWithTag(slideToggle, 'input');
     Simulate.change(input);
-    expect(onToggle).toHaveBeenCalledWith(false);
+    expect(onToggle).toHaveBeenCalled();
```
markdalgleish: Why are we no longer asserting that the value is being provided to the callback?

georgespyropoulos: Ah, I meant to come back and tidy this up.  I changed the functionality to pass through the entire event object and was struggling to find the doco on how to assert that any object was being passed into the function in Jest.  Have just updated the test.

react/SlideToggle/SlideToggle.test.js
```diff
@@ -0,0 +1,50 @@
...
+    const slideToggle = renderIntoDocument(
+      <SlideToggle
+        id="testToggle"
+        label="Test toggle"
+        onToggle={onToggle}
+      />
+    );
+    const input = findRenderedDOMComponentWithTag(slideToggle, 'input');
+    Simulate.change(input);
+    expect(onToggle).toHaveBeenCalledWith(expect.any(Object));
```
markdalgleish: My concern here is that we're not testing that the right value is coming through. Can we assert that we're getting the correct `true/false` value here?

georgespyropoulos: I was thinking that there probably wasn't much value in a test like that since the passed in `onToggle` function is now bound directly to the `onChange` attribute of the `input` and no value is explicitly passed anymore.  On the other hand, the test might be valuable if lots of changes are made to the component because it'll be more obvious what value the `onToggle` function passes

react/SlideToggle/SlideToggle.test.js
```diff
@@ -45,6 +45,10 @@ describe('Slide toggle:', () => {
     );
     const input = findRenderedDOMComponentWithTag(slideToggle, 'input');
     Simulate.change(input);
-    expect(onToggle).toHaveBeenCalledWith(expect.any(Object));
+    expect(onToggle).toHaveBeenCalledWith(expect.objectContaining({
+      target: expect.objectContaining({
+        checked: false
```
markdalgleish: 👏

react/SlideToggle/SlideToggle.test.js
markdalgleish: Since `onToggle` is just mapping through to `onChange`, can we just call it `onChange` instead?

PR #197, opened 2017-05-16
fix(Dropdown): Fix dropdown chevron styling bug

georgespyropoulos:
screen shot 2017-05-16 at 11 13 29 am

The order of CSS classes in the stylesheet is the only thing that ensures that `display: flex` is the style applied to the dropdown's chevron icon.  For apps that import the chevron before the dropdown, this issue doesn't exist but if the imports happen in the reverse order, this issue is present.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@
 
 Don't post anything or commit any code that isn't ready for the entire world to see. If you're introducing new patterns, try to keep it abstract, rather than making specific reference to the features under development. While a lot of this information is probably harmless, it's better to be safe than sorry.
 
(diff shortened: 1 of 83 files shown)


markdalgleish:
Can we fix this without the use of important? I've dealt with issues like this in the past by introducing a wrapper element, for example.

georgespyropoulos:
Yeah we could do it with a wrapper element.  What's your thinking behind avoiding the use of important for something like this?

markdalgleish:
It's a bit of a blunt instrument to avoid writing styles in a way that avoids clashes in the first place, I'd prefer to always treat `!important` as a last resort.

georgespyropoulos:
Fair enough, I'll go with a wrapping element instead.  I went with the important fix to try and avoid introducing markup that's only there for styling purposes but I agree that using important isn't a great solution either

markdalgleish:
👍

PR #198, opened 2017-05-16
chore(Sketch): Add experimental Sketch file support

markdalgleish:
Sketch files can now be handled in git, but since they are actually zip files, it requires unzipping before checking in, and zipping after checking out. To automate this process, two npm scripts are provided:

After check out:
`npm run git-to-sketch`

Before check in:
`npm run sketch-to-git`

This PR contains an encoded version of `Button.sketch` as a proof of concept. Still very much an experiment, but I'm keen to get some early feedback.


markdalgleish:
Any thoughts on how we should document this? Or should we leave it undocumented until we smooth out the workflow a bit with some early adopters?

PR #203, opened 2017-05-25
feat(z-index): Introduce z-index variables

kosanna:
Would love them to be 1xx as it gives us more freedom, but in order to not break anything let's leave them as x's for now.


theme/layout/z-indexes.less
```diff
@@ -0,0 +1,3 @@
+@z-index-header: 3;
+@z-index-apply-button: 2;
```
markdalgleish: This one doesn't appear to be used internally? This seems too app-specific to be part of the public API.

theme/layout/z-indexes.less
```diff
@@ -0,0 +1,5 @@
+@z-index-header-overlay: 4;
```
markdalgleish: Probably worth incrementing in units of 10 to leave room between them.

kosanna: Yep it definitely is, I would even go for 1xx, but I'm conscious of moving Chalice to these variables, in this case it will be less painful if we miss one or two. But yeah, probably should be just careful in migrating...thoughts?

markdalgleish: Yeah, we can leave the values for now if you like, we can always revisit at a better time.

kosanna: 👍

README.md
```diff
@@ -236,6 +236,18 @@ As much as possible, colors should be directly imported from the style guide. Th
 @sk-yellow
 ```
 
+### Z-indexes
+
+To ensure correct relative elements stacking order z-index variables are provided:
+
+```less
+@z-index-header-overlay: 4;
```
markdalgleish: Is it worth leaving the actual values out of the readme, and only documenting their stacking order?

PR #205, opened 2017-05-26
fix(Footer): Make Footer full width

markdalgleish:
After discussing this with a few of the people integrating the new header and footer, decided that we don't want the footer to use a `Section` component anymore, instead taking up the full width of the containing `PageBlock`.


markdalgleish:
Just updated the PR to add the missing gutters below 1025px.

PR #207, opened 2017-05-29
feat(Hidden): Add Hidden mobile/desktop based on responsive breakpoint mixins

kosanna:


michaeltaranto:
This looks good to me, only question would be around semantics, we are refraining from using `mobile`/`desktop` elsewhere to break away from devices, instead opting for `small` etc. @markdalgleish you have any thoughts on this?

markdalgleish:
@michaeltaranto thinking about this more, I'm not against being a bit more pragmatic here and using the terms that people would use normally. We can always adjust the API if we need to, but "mobile" and "desktop" makes sense for our current approach IMO.

PR #208, opened 2017-05-30
fix(SlideToggle): Styling tweak for slide toggle

georgespyropoulos:


react/SlideToggle/SlideToggle.less
```diff
@@ -50,7 +51,8 @@
 
 .label {
   padding: 0;
-  display: inline;
+  line-height: @grid-row-height * 3;
+  transform: none;
```
markdalgleish: Seems like we're missing a text style here — @michaeltaranto ?

michaeltaranto: This doesn't look like it should be using `Text`, we are using the `raw` prop to remove the padding, setting the line-height and overriding the transform. That's about everything that `Text` gives you so need to define/look for an alternative component in my opinion

georgespyropoulos: I used `Text` because this label should have responsive text.  Is there a nice way to opt into responsive text without using `Text`?

michaeltaranto: Not responsive text without the baseline correction. This is the bandaid that `.touchableText` was striving for but that also missed the mark

georgespyropoulos: The requirement is that this label has 18px text on smaller screens and 14px text on larger screens.  The only other option I can see here is to write a media query here?

georgespyropoulos: @michaeltaranto any more thoughts on this?  I guess the problem is that we currently don't have a concept of responsive text without baseline correction.  Some options I can think of are:

1.  Use the `standardTextResponsive` mixin without the text component.  This would still require overriding basekick.
2.  Change the responsive type mixins so they accept a param to turn off basekick.
3.  Use `touchableText`.  What about this?

```
.label {
  .touchableText();
  margin-top: -@grid-row-height;

  @media only screen and (min-width: @responsive-breakpoint) {
    .touchableText(@standard-type-scale);
  }
}
```

georgespyropoulos: I was chatting with @mattsjones yesterday about how there are some instances where there's a need for responsive text without baseline correction and that perhaps we should make that a thing in styleguide.  I don't think this is something we should use or need often but there are some cases where it is useful, like the slide toggle.  @markdalgleish, do you have any thoughts on this given you've just done a lot of work on responsive text in the styleguide?

markdalgleish: Ha, that's funny, I literally just finished chatting with @dborysiewicz about exactly this. He's planning to propose a change to the text mixins and components, allowing you to write `<Text baseline={false}>`, which I'm guessing will do the trick here?

georgespyropoulos: Yes that's exactly what we need here

PR #211, opened 2017-05-30
fix(Header): Fix employer site link

markdalgleish:


react/Header/employerLinkForLocale.js
```diff
@@ -1,3 +1,3 @@
-export default locale => locale.toLowerCase() === 'NZ' ?
+export default locale => locale.toLowerCase() === 'nz' ?
```
michaeltaranto: I would prefer to regex this `/^nz$/i.test(locale)`

PR #213, opened 2017-06-02
feat(Button): 	Add full width prop

kosanna:


mengtzu:
👍

kosanna:
🤔  I'd rather add them to button themselves as we're using it multiple times across Chalice

mengtzu:
The issue you'll probably have is buttons you want full width for small screens but not for large.  That's fine if it's easy to mess with their containers, but you won't want to change your prop based on screen size.

georgespyropoulos:
I have that exact use case for something I am working on at the moment - I have a button that I only want to be full width on mobile and the prop doesn't help at all

mengtzu:
Arguably though because it is such a common use case on mobile, mobile-first thinking is to use the prop and desktop can just manipulate a wrapper :)

michaeltaranto:
I think the prop names should be more declarative rather than boolean-esc, eg. `full-width`. Not so sure about the text wrap configuration, our use case is 'interesting' and not sure that the style guide should introduce a supported api for this... yet.

kosanna:
Can I merge this one? :)

kosanna:
@mengtzu? Second approval? 😆

PR #215, opened 2017-06-10
feat(Enzyme): add enzyme, tests cleanup

kosanna:
- Remove toJSON everywhere
- Update snapshots
- Tests cleanup

🥃 

Long term goal is decommissioning of `react-test-renderer`, `sinon` and `chai`


test/jest.config.json
```diff
@@ -1,6 +1,6 @@
 {
   "testPathIgnorePatterns": [
-    "<rootDir>[/\\\\](dist|node_modules)[/\\\\]"
+    "<rootDir>/(dist|node_modules)/"
```
markdalgleish: Why was this change made? IIRC this was for Windows support.

test/jest.config.json
```diff
@@ -9,7 +9,7 @@
     "^.+\\.js$": "<rootDir>/test/babel.transform.js",
     "^.+\\.svg$": "<rootDir>/test/svg.transform.js"
   },
-  "transformIgnorePatterns": [
-    "[/\\\\]node_modules[/\\\\].+\\.(js|jsx)$"
```
markdalgleish: Why was this removed? (Not saying it shouldn't have been, just curious)

mengtzu:
This looks good, but I'm not sure I should approve commits made on a long weekend on general principle :p

markdalgleish:
One concern I have here is that we've migrated some tests to use a shallow renderer. This definitely cleans up the snapshots, but it also means we're reducing test coverage in large components like the header and footer. Any thoughts on how we should deal with this? Should we add shallow snapshot tests for all the sub-components?

mattsjones:
My feelings here are that the header and footer should be deep rendered snapshots. The components don't make sense bu themselves so they should be rendered as one.

kosanna:
We're not loosing any actual tests with replacing to shallow anywhere in this case. If something need to be tests deeply, we need to test nested components themselves.

And btw, sinon could go later on too.

test/jest.config.json
kosanna: Because I didn't understand why it should be there and it works without it :P

test/jest.config.json
kosanna: Ok, got it, can roll it back

test/jest.config.json
kosanna: Ah, sorry, I thought you mean [/\\\\]

kosanna: Had issues with toJson serialiser, looks like it's fine now

markdalgleish:
We're not losing tests only in terms of the number of assertions, but we've lost test coverage in the sense that there are now changes that can be made to the header that won't get picked up in our test suite. I feel like this PR should either revert back to deep snapshots where appropriate, or introduce the missing snapshot tests.

kosanna:
Comments are addressed @here. Header and Footer snapshots are fully rendered now. Please revisit the PR :)

kosanna:
Anyone's keen for this PR to go in?

react/Footer/Footer.test.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { render } from 'enzyme';
```
markdalgleish: 👍

PR #217, opened 2017-06-13
feat(Text): Flatten mobile type hierarchy

markdalgleish:
BREAKING CHANGE: Design update required.

This PR makes the following changes to our typography:
- Make all mobile headings 2.1 over 3 rows and bold
- Add `<Regular>` / `<Text regular>` / `@sk-regular` to optionally disable bold text
- Add `<Text substandard>` for standard text that stays at 1.4 on mobile instead of growing to 1.8
- Shrink `<Section header>` vertical padding from 4 rows to 3 rows on mobile, since heading text is now smaller


markdalgleish:
Note that this PR is currently a proposal rather than a finalised change. We'll be actively working through this change soon, but if you're keen to get in early, feel free to try this out in a branch of your product.

theme/type/type.less
```diff
@@ -10,52 +10,60 @@
 @font-descender-height-scale: 0.1365;
 
 // Hierarchy
-@hero-type-row-span: 6;
 @hero-type-scale: 4.2;
-@hero-type-row-span-mobile: 5;
-@hero-type-scale-mobile: 3.6;
+@hero-type-row-span: 6;
```
markdalgleish: There's a much larger diff here because I wanted to re-order all the variables to follow a consistent pattern—otherwise this change was actually really hard to write, let alone read.

markdalgleish:
Looks like we're okay to progress with this from a design perspective, just need to co-ordinate the release as a group 👍

PR #218, opened 2017-06-15
feat(Text): Add option to turn off baseline for text

dborysiewicz:
Added option for type mixins and `Text` component to support turning off the baseline modifier (basekick) this will allow use to vertical centre text.

Not I'm sure if I'm happy with the idea of `.nonBaselineText()` being referenced in `baselineType.less`? We could just not use a mixin and move the code directly or maybe just use `.rawText()` - thoughts?


react/Text/Text.demo.js
```diff
@@ -19,6 +19,13 @@ export default {
             ...props,
             strong: true
           })
+        },
+        {
+          label: 'Baseline False',
```
markdalgleish: Nitpick: Maybe "Disable baseline" instead?

theme/type/baselineType.less
```diff
@@ -2,11 +2,16 @@
...
+.__baselineType(@type-scale; @row-span; @baseline) when (isnumber(@type-scale)) and (isnumber(@row-span)) and (@baseline = true) {
   .basekick(@type-scale,
             @row-span,
             @font-descender-height-scale,
             @base-font-size,
             @grid-row-height);
 }
+
+.__baselineType(@type-scale; @row-span; @baseline) when (isnumber(@type-scale)) and (isnumber(@row-span)) and (@baseline = false) {
+  .nonBaselineText(@type-scale, @row-span);
```
markdalgleish: Should we just inline this mixin, since it's not used anywhere else?

theme/type/baselineType.less
```diff
@@ -2,11 +2,16 @@
 
 @import (reference) "../layout/grid";
 @import (reference) "type";
+@import (reference) "../mixins/type";
 
-.__baselineType(@type-scale; @row-span) when (isnumber(@type-scale)) and (isnumber(@row-span)) {
+.__baselineType(@type-scale; @row-span; @baseline) when (isnumber(@type-scale)) and (isnumber(@row-span)) and (@baseline = true) {
```
markdalgleish: I wonder if this file and its mixins needs to be renamed now that it's not always about being on baseline? 🤔

dborysiewicz: Yep this and your other comment is what I was getting at not 100% happy... I'll chat to you on Slack.

markdalgleish: lol, don't mind me, just stealing your comments and palming them off as my own 💪

react/Text/Text.less
```diff
@@ -20,11 +19,19 @@
   }
 }
 
-.small { .smallTextResponsive(); }
-.subheading { .subheadingTextResponsive(); }
-.headline { .headlineTextResponsive(); }
-.heading { .headingTextResponsive(); }
-.hero { .heroTextResponsive(); }
+.root:not(.baseline) { .standardTextResponsive(@baseline: false); }
```
georgespyropoulos: What about grouping these by type rathen than by baseline / non-base line eg.

```
.root {
  &:not(.baseline) {.standardTextResponsive(@baseline: false); }
  &.baseline { .standardTextResponsive(@baseline: true); }
}
```

mengtzu:
👍

kosanna:
Ah yes, was actually going to ask how we do vertical centering having baseline. Cool!

PR #219, opened 2017-06-19
feat(type): Switch to system font stack

markdalgleish:
We're switching away from web fonts in favour of a more performant system font stack strategy

BREAKING CHANGE: The webpack web font bundle and 'classicFontStack()' mixin have been removed. Your
app will fail to compile until these have been removed from your application.


mattsjones:
Not sure this is a feature seeing as we're not actually adding anything? Also can you update [seek-style-guide-webpack](https://github.com/seek-oss/seek-style-guide-webpack) as well.

markdalgleish:
I tend to use feature for any significant UI change that isn't just a minor bug fix, but happy to change if others disagree.

markdalgleish:
@mattsjones besides the commit message, happy with this change?

PR #220, opened 2017-06-19
feat(Header): Add ability to switch saved/applied links on auth status

dborysiewicz:
Moved authStatusTypes out of /header to /private to make it reusable for both header & footer, added authStatus to Footer & FooterLink and lastly check authStatus to change saved & applied links depending on status in both Header & Footer.

Only issues I see in this one is that the moving of authStatusTypes might break other apps if people where reading them in?
The other is Footer now also requires authStatus to be passed in however it will fail gracefully and revert with the default props.


markdalgleish:
@dborysiewicz just a heads up that I renamed this, since I'm pretty sure the forward slash in the scope trips up semantic release.

react/Footer/FooterLink/FooterLink.js
```diff
@@ -14,6 +15,7 @@ export default function FooterLink({ secondary, partner, analytics, className, l
         linkRenderer({
           'data-analytics': analytics,
           className: styles.link,
+          href: authenticationStatus === AUTHENTICATED && authedHref || href,
```
markdalgleish: Could this be a ternary instead?

mengtzu:
Agree with @markdalgleish  on the ternary, otherwise 👍

markdalgleish:
@dborysiewicz what about recommended jobs?

dborysiewicz:
Closing for now as the nudges is on hold/redesign :(

PR #223, opened 2017-06-27
fix(Textarea): Update font size to 18px to match other input components

dborysiewicz:
Pretty simple change, just up the font size to match the other text input components.

### DIFF
--- a/react/Textarea/Textarea.less
+++ b/react/Textarea/Textarea.less
@@ -11,9 +11,7 @@
   height: @grid-row-height * 10;
   resize: vertical;
   max-height: @grid-row-height * 20;
-  // line-height: 25px was picked as 18px is too tight and 27px looks too much like page copy.
-  // It is a exception to the on grid rule as Textarea itself will be on grid.
-  line-height: 25px;
+  line-height: 25px; // Deliberate deviation to avoid looking like body copy
   color: @sk-charcoal;
   border: @field-border-width solid @sk-mid-gray-light;
   border-radius: @field-border-radius;
--- a/react/Textarea/Textarea.less
+++ b/react/Textarea/Textarea.less
@@ -12,7 +12,7 @@
   resize: vertical;
   max-height: @grid-row-height * 20;
   // line-height: 25px was picked as 18px is too tight and 27px looks too much like page copy.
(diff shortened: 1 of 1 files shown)


markdalgleish:
The line height seems a little tight on this — @stephaniemoss?

screen shot 2017-06-28 at 9 09 54 am

markdalgleish:
Although, increasing the line height by one row makes it look less like a textarea 🤔 

screen shot 2017-06-28 at 9 17 01 am

stephaniemoss:
It should be 27px. What do you mean by it looking less like a text area?

markdalgleish:
I dunno, it's really a gut feel thing, but it looks a bit more like body copy, designed for reading rather than writing.

markdalgleish:
Granted the textarea in my screenshot is abnormally small.

markdalgleish:
@dborysiewicz as @stephaniemoss said, can you bump the line height up another row?

stephaniemoss:
This text doesn't necessarily have to fit on the grid but making the leading tighter would involve introducing some new numbers since 21 would be too tight.

This is Twitter:
image

react/Textarea/Textarea.less
```diff
@@ -6,12 +6,12 @@
...
 
 .textarea {
-  .rawText();
+  .touchableText();
   width: 100%;
   height: @grid-row-height * 10;
   resize: vertical;
   max-height: @grid-row-height * 20;
-  line-height: @grid-row-height * 2;
+  line-height: 25px; // Deliberate deviation to avoid looking like body copy
```
mengtzu: :+1: for comment

react/Textarea/Textarea.less
```diff
@@ -6,12 +6,12 @@
 }
 
 .textarea {
-  .rawText();
+  .touchableText();
```
markdalgleish: Is this mixin worth it, given we're overriding the line height? Thinking it might be better to inline the rules, since it's not semantically "touchable text" anyway.

dborysiewicz: Just spoke to @markdalgleish on Slack, we are happy to leave as is as it's consistent with TextField, Dropdown, Button, etc however in future we should break away all from `touchableText` to something like `superstandard` for inputs.

PR #224, opened 2017-06-28
fix(Header): Fix max-width of Header, Footer & PageBlock

dborysiewicz:
The css rules where not 100% correct and at screen sizes 1000px to 1025px content was actually larger then our agreed limit of 940px, along with the fix I also believe this change simplifies the css.

### DIFF
--- a/react/Footer/Footer.less
+++ b/react/Footer/Footer.less
@@ -1,23 +1,18 @@
 @import (reference) "~seek-style-guide/theme";
 @import (reference) "./Footer.common.less";
 
-@classic-ia: ~"only screen and (min-width: 1025px)";
-
 .root {
   background-color: @sk-footer;
   width: 100%;
 }
 
 .content {
-  padding: (@grid-row-height * 4) @grid-gutter-width;
+  padding: (@grid-row-height * 4) (@grid-gutter-width / 2);
+  max-width: @grid-container-width + @grid-gutter-width;
+  margin: 0 auto;
 
   @media only screen @footer-horizontal {
     padding-bottom: (@grid-row-height * 5);
-  }
-
-  @media @classic-ia {
-    padding-left: 0;
-    padding-right: 0;
   }
 }
 
--- a/react/Footer/Footer.less
+++ b/react/Footer/Footer.less
@@ -7,8 +7,8 @@
 }
 
 .content {
-  padding: (@grid-row-height * 4) (@grid-gutter-width / 2);
-  max-width: @grid-container-width + @grid-gutter-width;
+  padding: (@grid-row-height * 4) @grid-gutter-width;
+  max-width: @grid-container-width + (@grid-gutter-width * 2);
   margin: 0 auto;
 
   @media only screen @footer-horizontal {
--- a/react/Header/Header.less
+++ b/react/Header/Header.less
@@ -1,11 +1,5 @@
 @import (reference) "~seek-style-guide/theme";
 
-.classicDesktopOnly(@rules) {
-  @media only screen and (min-width: 1025px) {
-    @rules();
-  }
-}
-
 .root {
   width: 100%;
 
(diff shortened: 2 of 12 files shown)


markdalgleish:
You can view the changes without whitespace diffs here: https://github.com/seek-oss/seek-style-guide/pull/224/files?w=1 (this PR needs it)

react/Footer/Footer.less
```diff
@@ -1,24 +1,19 @@
...
-@classic-ia: ~"only screen and (min-width: 1025px)";
-
 .root {
   background-color: @sk-footer;
   width: 100%;
 }
 
 .content {
-  padding: (@grid-row-height * 4) @grid-gutter-width;
+  padding: (@grid-row-height * 4) (@grid-gutter-width / 2);
```
kosanna: Why we decrease it twice? I don't get it.

kosanna: That will make it 10 and be quite out of sync 🤔

dborysiewicz: 10px padding is what is currently used on the header to keep the logo/text of the edge of the screen. This is case particular along with matching that 20px is too much as it looks like just another normal PageBlock.

kosanna: Why we decided not to use PageBlock here?

kosanna: I'm just a bit worried that we will have inconsistent gutters everywhere

dborysiewicz: Using page block means we cant easily do the above 10px rule. I'm happy to change this however this was done to match what was in the header already.... Lets move over to Slack.

react/Header/Header.less
```diff
@@ -33,6 +27,7 @@
   align-items: center;
   flex: 1 0 auto;
   padding: 0 (@grid-gutter-width / 2);
+  max-width: @grid-container-width + @grid-gutter-width;
```
markdalgleish: Obviously the end result works, but why do we only add a single gutter width as opposed to two?

react/PageBlock/PageBlock.less
```diff
@@ -1,12 +1,6 @@
...
-
 .content {
-  @media @classic-ia {
-    max-width: @grid-container-width + (@grid-gutter-width * 2);
-    margin: 0 auto;
-    padding-left: @grid-gutter-width;
-    padding-right: @grid-gutter-width;
-  }
+  max-width: @grid-container-width;
+  margin: 0 auto;
```
markdalgleish: Why are removing padding? Won't this push some content to the edges of the screen?

markdalgleish:
@dborysiewicz have you tried out the `PageBlock` changes in your app?

react/Header/Header.less
dborysiewicz: This is sort similar to the footer discussion but because we are only adding only half a gutter padding when there is no outer gutters on page blocks to keep the icons/text off the edge - why we are doing that in the first place I'll have to explore the commit history....

react/PageBlock/PageBlock.less
dborysiewicz: Only content is not in a section, this padding was the 'outer gutters' however @stephaniemoss prefers it to gradually disappear over 20px jump.

dborysiewicz:
@markdalgleish tested using the style guide playground, I can create & deploy a version of sign in & rego with it if we think it's necessary?

react/Header/Header.less
dborysiewicz: The 10px (half gutter) existed from all the way back when the Header lived in Chalice added here https://github.com/SEEK-Jobs/chalice/commit/a9817bc72a49ca74efd787bd4e3005138660b30b I guess only @michaeltaranto would know thats if he remembered...

markdalgleish:
@dborysiewicz not necessarily deployed, but I think sanity checking it in your app would be a good idea.

react/Header/Header.less
markdalgleish: Fair enough 👍

dborysiewicz:
Tested into job apply and looks good on my end, once this is merged will be doing a fast follow with a deploy to production for job apply.

PR #226, opened 2017-06-29
fix(Sketch): Update form Sketch symbols

markdalgleish:
Updated `Button` and `Checkbox`, added `Radio` (currently not implemented!) and `Dropdown`.

Part of this update was also cleaning up the symbol names to improve navigation.


react/Button/Button__sketch/meta.json
```diff
@@ -12,31 +12,31 @@
       "name": "Symbols",
       "artboards": {
         "C31A3239-8FB1-4021-BEA5-09E444FA4E89": {
-          "name": "Secondary button - hover"
+          "name": "Button/Grey/Hover"
```
markdalgleish: Here's a good example of the improved symbol names.

react/Button/Button__sketch/meta.json
```diff
@@ -1,7 +1,7 @@
 {
-  "commit": "2ac3f43dc64a2b7982dc85b584be457e7bbba129",
-  "appVersion": "43.2",
-  "build": 39069,
+  "commit": "62a429066b6505d0a6a211f8b74e8c5a38b40e00",
+  "appVersion": "45",
+  "build": 43475,
```
markdalgleish: This is kind of annoying :-/

markdalgleish:
The `Button.sketch` preview png is marked as different, but it's totally the same when you eyeball it. Not sure why, it's possible Sketch 45 is compressing the PNG or something...

markdalgleish:
By the way, I learnt a valuable lesson with this PR—**the Sketch preview PNG is generated based on the active page**, so you need to make sure you're not on the symbols page when you hit save or your preview will be totally different...

markdalgleish:
Actually, re: my previous comment, it looks like the button PNG did change, some text moved very slightly for no reason. I might need to redo that one.

markdalgleish:
Fixed, the image diff for `Button.sketch` is now gone, it's 100% the same now, just with renamed symbols 👍

PR #231, opened 2017-07-04
feat(Text): Add Paragraph and superstandard type

markdalgleish:
This PR introduces `superstandard`, another variant of `standard` similar to `substandard` but operating in the other direction. Where `substandard` means "keep standard type small on mobile" (14px), `superstandard` means "keep standard type big on desktop" (18px).

The `Paragraph` component is much more self explanatory, rendering a `<p>` tag with a bottom margin of a single grid row.

Both of these changes are being introduced together to ease the creation of long form text, but they're obviously useful in the general case too.


markdalgleish:
@stephaniemoss would love your input on this one 😄

PR #236, opened 2017-07-13
docs(readme): Update readme

markdalgleish:
The readme hasn't been properly reviewed in a while, and was in need of some major changes, particularly to make it easier for newcomers to the project. So, I've made the following changes:

- Mentioned sku before talking about installation, since it has its own setup step that needs to go first.
- Removed documentation for the webpack v1 decorators, since they're only provided for maintenance of older projects.
- Removed the section about alternative bundler support—it's all theoretical, and honestly just noise.
- Made the `<StyleGuideProvider>` code sample valid JavaScript to make it easier for those who are less familiar with this ecosystem.
- Prioritised documentation of components over Less mixins and variables—we've come a long way since the first cut of this readme!
- Also highlighted that we're trying to minimise the use of Less going forward.
- Removed documentation of typographical mixins, since `<Text>` components are now the standard approach.
- Moved "Optimising Imports" to a new "Advanced Usage" section at the bottom of the readme, to get it off the critical path for newcomers (especially since sku takes care of this).
- Made a few other minor edits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,11 @@
 
 # seek-style-guide
 
-Living style guide for [SEEK](https://github.com/seek-oss), powered by [React](https://facebook.github.io/react), [Less](http://lesscss.org/) and [CSS Modules](https://github.com/css-modules/css-modules).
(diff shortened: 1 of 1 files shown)


neerajvisal:
Thanks @markdalgleish for taking time in updating this.

PR #239, opened 2017-07-17
feat(Header): Add 'returnUrl' to sign out link

conshow:


mengtzu:
Does it make sense to send people to `LogoutWithReturnUrl` if they don't have a returnUrl?  Also seems like there should be a test somewhere for the case where there is a return url (maybe not the responsibility of this PR though).

Otherwise looking good.

conshow:
Thanks @mengtzu. Good point. I have made the change to test for whether the return url exists

react/Header/UserAccountMenu/UserAccountMenu.js
```diff
@@ -199,7 +199,9 @@ export default function UserAccountMenu({ locale, authenticationStatus, linkRend
                 'data-analytics': 'sign-out',
                 className: styles.item,
                 onClick: clearLocalStorage,
-                href: '/Login/Logout',
+                // Both urls are existing Jobseeker urls
+                // Login/Logout is a nicer looking url so using that when we can
```
markdalgleish: Can we remove these comments? Not sure they're adding much value here.

PR #241, opened 2017-07-18
feat(Autosuggest): Add mobile backdrop

kosanna:


markdalgleish:
Would you be able to make this fade in? It's quite a jarring transition at the moment.

kosanna:
@markdalgleish added transition

kosanna:
Good to go, will just confirm re backdrop colour with Tim

theme/mixins/backdrop.less
```diff
@@ -0,0 +1,11 @@
+.backdrop() {
+  position: fixed;
+  z-index: 0;
```
michaeltaranto: Is this `z-index` required? If so should we add an entry into [our z-index hierarchy](https://github.com/seek-oss/seek-style-guide/blob/master/theme/layout/z-indexes.less)?

react/Autosuggest/Autosuggest.less
```diff
@@ -65,3 +65,18 @@
...
+
+.containerOpen + .autosuggestBackdrop {
+  .smallDeviceOnly({
+    .backdrop()
+  });
+}
+
+.containerOpen .isLabelCoveredWithBackdrop {
+  display: block;
+  pointer-events: none;
```
markdalgleish: 👏

theme.less
```diff
@@ -1,14 +1,18 @@
...
 @import (reference) "theme/mixins/link";
 @import (reference) "theme/mixins/type";
+
+@import (reference) "theme/palette/accessible-variants";
+@import (reference) "theme/palette/brand";
+@import (reference) "theme/palette/elements";
+@import (reference) "theme/palette/grays";
+@import (reference) "theme/palette/partners";
+
+@import (reference) "theme/type/type";
```
markdalgleish: Was this change deliberate?

kosanna: Sort and split into groups, sorry for being annoying with that

kosanna:
Tim confirmed colors/opacity are ok 👍

markdalgleish:
@kosanna @michaeltaranto What are your thoughts on portrait iPad? Should it have the backdrop like mobile, or go without like desktop?

markdalgleish:
Also, do we want to show a transparent backdrop on desktop to intercept clicks?

kosanna:
@markdalgleish it's a difult functionality of autosuggest so no need to introduce a transparent backdrop

kosanna:
Re landscape tablet...my opinion we should treat just based on width (?) regardless of the device.

markdalgleish:
@kosanna the default behaviour for react-autosuggest is to close suggestions when clicking outside, but the click still registers with whatever you click on. This is different to how our classifications list behaves, and in this PR it's now different to the behaviour on mobile.

markdalgleish:
@kosanna agreed about width, but I feel like showing the backdrop on <=768px widths could be a reasonable exception.

kosanna:
@markdalgleish I've changed classification to behave in a similar way (it's in a PR on chalice), will pull it into a separate one today perhaps if this one is merged

kosanna:
Change backdrop to appear on tablets. As we'll need to use it for classifications as well, we need to have a mixin exposed too

kosanna:
Need second approval on this one :)

PR #244, opened 2017-07-19
feat(Autosuggest): Scroll to autosuggest on focus

kosanna:


react/Autosuggest/Autosuggest.js
```diff
@@ -5,10 +5,13 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import ReactAutosuggest from 'react-autosuggest';
 import IsolatedScroll from 'react-isolated-scroll';
-import omit from 'lodash.omit';
+import { get, invoke, omit } from 'lodash';
```
markdalgleish: This is going to result in lodash being included in its entirety for some users, we should import the things we need individually.

react/Autosuggest/Autosuggest.demo.js
```diff
@@ -24,6 +24,10 @@ class AutosuggestContainer extends Component {
     });
   };
 
+  handleFocus = () => {
+    console.log('focused');
```
markdalgleish: Is this meant to be here?

react/Autosuggest/Autosuggest.js
kosanna: ok, will import individually

react/Autosuggest/Autosuggest.demo.js
kosanna: It's for test, it's demo

markdalgleish: Oops, yep :sweat_smile:

markdalgleish:
So this demos correctly on mobile devices, it might be a good idea to un-fix the header on the style guide demo site in this PR?

react/Autosuggest/Autosuggest.js
```diff
@@ -68,9 +81,28 @@ export default class Autosuggest extends Component {
     );
   }
 
+  scrollOnFocus = () => {
+    // Temporary solution till we do CSS in JS
```
markdalgleish: Is it okay if we remove this comment?

react/Autosuggest/Autosuggest.js
```diff
@@ -68,9 +81,28 @@ export default class Autosuggest extends Component {
...
+
+    if (isMobileWidth) {
+      this.textField.scrollIntoView();
+    }
+  }
+
   renderInputComponent(inputProps) {
+    const onFocus = () => {
+      this.scrollOnFocus();
+      inputProps.onFocus(event);
```
markdalgleish: We should probably be checking that the function exists first, or this could throw.

kosanna: yep, good point

react/Autosuggest/Autosuggest.js
kosanna: just don't want to forget about it

markdalgleish: My main concern is that comments like these tend to hang around forever and age really badly, particularly if plans change.

kosanna:
Help on double tap issue on clear is appreciated. Could be fixed though, I can't reproduce it

docs/src/components/App/Header/Header.less
```diff
@@ -1,15 +1,7 @@
...
-  height: @grid-row-height * 8;
-}
-
 .headerBlock {
   background: @sk-blue;
-  position: fixed;
-  top: 0;
-  left: 0;
-  right: 0;
   z-index: 10;
```
michaeltaranto: With the removal of `fixed` the `z-index` could probably be removed too. Thoughts?

kosanna: Yup, good pickup 👍

kosanna: Looks weird when menu is opened though :D But we could fix it when we have disable scroll component

kosanna:
Updated with profile scroll. Now just need to fix onClear bug. Thoughts?

react/Autosuggest/Autosuggest.js
```diff
@@ -5,9 +5,16 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import ReactAutosuggest from 'react-autosuggest';
 import IsolatedScroll from 'react-isolated-scroll';
-import omit from 'lodash.omit';
+
+import get from 'lodash/get';
+import invoke from 'lodash/invoke';
+import omit from 'lodash/omit';
```
markdalgleish: I don't think lodash is in our dependency list.

react/Autosuggest/Autosuggest.js
```diff
@@ -5,9 +5,16 @@ import PropTypes from 'prop-types';
...
 
 import TextField from '../TextField/TextField';
+import smoothScroll from '../SmoothScroll/SmoothScroll';
```
markdalgleish: Couple of things:
- This is named like a component, even though it's a util. Should start with a lowercase letter.
- This is in the components folder, when it should probably just be in the `Autosuggest` folder for now.

kosanna: But I need to use it for classifications too. So I'll probably just move it to /utils/smoothScroll

react/Autosuggest/Autosuggest.js
mattsjones: We should be installing the specific lodash utils we needs. Eg `lodash.omit`

kosanna: I'm not installing anything here...I'm just using what's already available...

markdalgleish: Right, but because it's not explicitly defined in the dependencies in `package.json`, this only works by accident right now.

kosanna: actually, you're right, there's no get and invoke. Should we add it?

markdalgleish: Yeah, sounds good.

kosanna: ok, sorted, literally and figuratively

package.json
```diff
@@ -110,10 +110,12 @@
     "json-loader": "^0.5.4",
     "jsx-to-string": "^0.2.13",
     "less-vars-to-js": "^1.1.1",
-    "lodash.debounce": "^4.0.6",
-    "lodash.flatten": "^4.4.0",
-    "lodash.omit": "^4.5.0",
-    "lodash.range": "^3.2.0",
+    "lodash.debounce": "4.0.6",
+    "lodash.flatten": "4.4.0",
+    "lodash.get": "4.4.2",
```
markdalgleish: These need to be dependencies, not dev dependencies, since style guide consumers need them.

kosanna: Moved it, but I'm not sure that anyone relies on style-guide lodash deps...

markdalgleish: The issue is that dev dependencies won't get installed by consumers, so they won't have these packages available.

react/Autosuggest/Autosuggest.js
```diff
@@ -5,9 +5,16 @@ import PropTypes from 'prop-types';
...
+
+import get from 'lodash/get';
+import invoke from 'lodash/invoke';
+import omit from 'lodash/omit';
 
 import TextField from '../TextField/TextField';
+import smoothScroll from '../private/smoothScroll';
+
+const responsiveBreakpoint = '740';
+const smallDeviceOnlyMedia = `(max-width: ${responsiveBreakpoint - 1}px)`;
```
markdalgleish: This evaluates to `'740' - 1` — should probably make sure these are both numbers.

react/TextField/TextField.js
```diff
@@ -72,20 +74,22 @@ export default class TextField extends Component {
...
-    const { onClear } = this.props;
-
-    if (typeof onClear === 'function') {
-      onClear(event);
-      this.input.focus();
-      event.preventDefault(); // https://developer.mozilla.org/en/docs/Web/API/HTMLElement/focus#Notes
-    }
+    event.preventDefault(); // https://developer.mozilla.org/en/docs/Web/API/HTMLElement/focus#Notes
+    invoke(this.props, 'onClear', event);
+    invoke(this.input, 'focus');
```
markdalgleish: Is this redundant, since we're calling a browser API that definitely exists?

react/Autosuggest/Autosuggest.js
kosanna: was resolving correctly though

markdalgleish: Yep, but it'll break the second somebody tries to do `responsiveBreakpoint + 1` 😅

markdalgleish:
Just noticed that un-fixing the header on the style guide site means that the menu can get into a broken state when it's open and you scroll down:

screen shot 2017-07-21 at 11 00 41 am

kosanna:
That's what I was talking about yesterday. Looks like it's easier for me to introduce offset there now :)

react/TextField/TextField.js
```diff
@@ -72,20 +74,22 @@ export default class TextField extends Component {
...
   storeInputReference(input) {
     if (input !== null) {
       this.input = input;
     }
   }
 
   handleMouseDownOnClear(event) {
-    const { onClear } = this.props;
-
-    if (typeof onClear === 'function') {
```
markdalgleish: Why was this `if` statement removed?

kosanna: because I used invoke instead, so there's technically no need in it...only if for some reason we don't want to prevent focus too

markdalgleish: That's what I thought at first, but the logic has actually changed now. Previously, this whole block would be skipped if `onClear` wasn't defined, which isn't the case anymore. Is that desired behaviour?

kosanna: I don't see the reason of blocking native onFocus...

kosanna: It's like if clear function is not passed, when you're trying to clear the field don't focus it...can't imagine the case it's needed

kosanna: You probably won't have the clear button in this case at all

kosanna:
@markdalgleish my pleasure, thank YOU for thoroughly going through all that too, #loveit

PR #246, opened 2017-07-19
feat(SmoothscrollPolyfill): Add polyfill to use { behaviour: smooth } scrolling option

kosanna:
Help appreciated with understanding why window is not defined despite of the fact I'm not requiring if it's not there

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "postcss-loader": "0.9.1",
     "prop-types": "^15.5.10",
     "raw-loader": "0.5.1",
+    "smoothscroll-polyfill": "^0.3.5",
     "style-loader": "0.13.1",
     "svgo": "0.7.1",
     "svgo-loader": "1.1.0",
--- a/react/StyleGuideProvider/StyleGuideProvider.js
+++ b/react/StyleGuideProvider/StyleGuideProvider.js
@@ -7,7 +7,8 @@ import classnames from 'classnames';
 import Helmet from 'react-helmet';
 import ScreenReaderOnly from '../ScreenReaderOnly/ScreenReaderOnly';
 
-if (typeof window !== 'undefined') {
+if (typeof window !== undefined) {
+  console.log('lolo: ');
   require('smoothscroll-polyfill').polyfill();
 }
 
--- a/react/StyleGuideProvider/StyleGuideProvider.js
+++ b/react/StyleGuideProvider/StyleGuideProvider.js
@@ -7,7 +7,7 @@ import classnames from 'classnames';
 import Helmet from 'react-helmet';
 import ScreenReaderOnly from '../ScreenReaderOnly/ScreenReaderOnly';
 
-if (window) {
+if (typeof window !== 'undefined') {
   require('smoothscroll-polyfill').polyfill();
 }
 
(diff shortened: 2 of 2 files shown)


react/StyleGuideProvider/StyleGuideProvider.js
```diff
@@ -7,6 +7,10 @@ import classnames from 'classnames';
 import Helmet from 'react-helmet';
 import ScreenReaderOnly from '../ScreenReaderOnly/ScreenReaderOnly';
 
+if (window) {
```
markdalgleish: Should be `typeof window === 'undefined'`

kosanna: It doesn't help I tried... :/

kosanna: weirdo

package.json
```diff
@@ -67,6 +67,7 @@
     "postcss-loader": "0.9.1",
     "prop-types": "^15.5.10",
     "raw-loader": "0.5.1",
+    "smoothscroll-polyfill": "^0.3.5",
```
michaeltaranto: @markdalgleish we are locking other packages to explicit versions, should the `^` be removed here?

markdalgleish: Yeah, probably—although in this case it won't make a difference since its v0.x.

kosanna: Should we lock all other packages too then?

react/StyleGuideProvider/StyleGuideProvider.js
```diff
@@ -7,6 +7,10 @@ import classnames from 'classnames';
 import Helmet from 'react-helmet';
 import ScreenReaderOnly from '../ScreenReaderOnly/ScreenReaderOnly';
 
+if (typeof window !== 'undefined') {
+  require('smoothscroll-polyfill').polyfill();
```
mattsjones: Do we really need to apply this to every app? It's only required for a specific component right?

markdalgleish: Good point, actually. At the very least, we could only import this in components that actually use it?

michaeltaranto: From an experience point of view it would be nice to be consistent on this. While this particular use case is to support the `AutoSuggest` component, by polyfilling at this level any consuming single page app that used `document.body.scrollIntoView()` to scroll to top could have the uniform smoothness. Thoughts @mattsjones @markdalgleish?

mattsjones: I might be a bit biased about this because profile has it's own scroll library as it's a huge part of our experience, which won't match this. We actually wrote it ourselves as we had specific requirements that this simple polyfill doesn't achieve. However, I can see your point.

markdalgleish: I guess this comes back to that broader discussion—is it the style guide's responsibility to ship a bunch of utils and polyfills? I'm not sure that sounds right.

kosanna: Purpose of the style-guide is to make experience consistent across the website....looks like it's the case...@mattsjones what's the library? Probably we should use it too in this case? :/

mattsjones: [ca-profile-ui/scroll](https://github.com/SEEK-Jobs/ca-myprofile-ui/blob/master/app/modules/scroll/index.js) is our scroll module. It's not profile specific though. Could be pulled out. Only thing is it's configurable, so we'd have to agree on settings.

PR #247, opened 2017-07-20
feat(Less): Replace mixins with variables for media queries

seangrant:
Updated some styles to fix stylelint problems that appear to have been ignored while using the mixins


react/Header/UserAccount/UserAccount.less
```diff
@@ -127,22 +129,22 @@
...
 }
 
 .chevronSvg {
   width: 14px;
   height: 14px;
   // Begin bold icon hacks
-  .smallDeviceOnly({
-    path {
+  @media @smallDeviceOnly {
+    path { // stylelint-disable-line selector-max-type
```
markdalgleish: You can avoid the comment with `:global(path)`, that's the intention of this rule in our codebase.

react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -84,39 +84,39 @@
...
 
 .iconSvg {
   width: 15px;
   height: 15px;
   // Begin bold icon hacks
-  .smallDeviceOnly({
+  @media @smallDeviceOnly {
     overflow: visible;
-    path {
+    path { // stylelint-disable-line selector-max-type
```
markdalgleish: `:global(path)` again.

react/Header/UserAccount/UserAccount.less
seangrant: 👍

react/Header/UserAccountMenu/UserAccountMenu.less
seangrant: 👍

markdalgleish:
@seangrant just a heads up that I wordsmithed the PR title 😄

theme/layout/breakpoints.less
```diff
@@ -1,14 +1,6 @@
 // Breakpoint
 @responsive-breakpoint: 740px;
 
-.smallDeviceOnly(@rules) {
-  @media only screen and (max-width: (@responsive-breakpoint - 1)) {
-    @rules();
-  }
-}
+@smallDeviceOnly: %(~"only screen and (max-width: %s)", @responsive-breakpoint - 1);
```
markdalgleish: While we're at it, can we rename this to `@mobile`?

theme/layout/breakpoints.less
```diff
@@ -1,14 +1,6 @@
...
 
-.aboveSmallDevice(@rules) {
-  @media only screen and (min-width: @responsive-breakpoint) {
-    @rules();
-  }
-}
+@aboveSmallDevice: %(~"only screen and (min-width: %s)", @responsive-breakpoint);
```
markdalgleish: Similarly, can we rename this to `@desktop`?

markdalgleish:
@seangrant I wonder if we should leave the old mixins for now and maintain backwards compatibility?

theme/layout/breakpoints.less
```diff
@@ -1,6 +1,10 @@
 // Breakpoint
 @responsive-breakpoint: 740px;
 
+@mobile: %(~"only screen and (max-width: %s)", (@responsive-breakpoint - 1));
+
+@aboveSmallDevice: %(~"only screen and (min-width: %s)", @responsive-breakpoint);
```
markdalgleish: Can we rename this one to `@desktop`?

docs/src/components/Demo/Demo.less
```diff
@@ -22,27 +22,25 @@
 
 .options {
   padding: @grid-row-height @grid-gutter-width (@grid-row-height * 2) @grid-gutter-width;
-
-  .aboveSmallDevice({
+  @media @desktop {
```
markdalgleish: Nice one—now this reads the way we *talk* about this stuff, both devs and designers, which is definitely what we're going for 💯 😎

theme/layout/breakpoints.less
```diff
@@ -1,6 +1,10 @@
 // Breakpoint
 @responsive-breakpoint: 740px;
 
+@mobile: %(~"only screen and (max-width: %s)", (@responsive-breakpoint - 1));
+
+@desktop: %(~"only screen and (min-width: %s)", @responsive-breakpoint);
+
 .smallDeviceOnly(@rules) {
```
markdalgleish: Sorry to nitpick this, but could you add a comment that the mixins are for backwards compatibility?

docs/src/components/Demo/Demo.less
seangrant: 👍  Added separate comment for each variable

theme/layout/breakpoints.less
```diff
@@ -1,12 +1,18 @@
 // Breakpoint
 @responsive-breakpoint: 740px;
 
+@mobile: %(~"only screen and (max-width: %s)", (@responsive-breakpoint - 1));
+
+@desktop: %(~"only screen and (min-width: %s)", @responsive-breakpoint);
+
+// DEPRECATED: Use @mobile variable instead.
 .smallDeviceOnly(@rules) {
   @media only screen and (max-width: (@responsive-breakpoint - 1)) {
```
markdalgleish: Actually, just realised we should probably refactor these mixins to use the new variables too.

seangrant: When you say it like that, it sounds obvious, will do that now

PR #248, opened 2017-07-20
fix(Header): Improve gutter alignment with content on small devices

michaeltaranto:
~Fix the additional guttering introduced to the Header banner in https://github.com/seek-oss/seek-style-guide/pull/224.~ Also bring the Header into line with standard content gutters (instead of arbitrary half gutters) on small devices to better align with content and the footer too.

Before:
screen shot 2017-07-21 at 9 48 20 am

After:
screen shot 2017-07-21 at 9 47 35 am


markdalgleish:
At 940px the header content goes right to the edge of the window without any gutters:

screen shot 2017-07-21 at 8 20 33 am

michaeltaranto:
Yeah that should probably be fixed in `PageBlock` content but wanted to talk through that as that's a bigger change

michaeltaranto:
Updating this PR to improve the content alignment for small devices. I'll move the fix for large devices out to it's own PR to be expedited through.

michaeltaranto:
Updated now to only reflect the small device gutter changes

markdalgleish:
One concern I have is how we treat alignment of the logo. Since it's a circle, it actually needs to be moved further to the left to be visually aligned with traditional rectangular content.

markdalgleish:
Visually *balanced*, I should say.

michaeltaranto:
Yeah agree, not sure that I agree with using "half gutters" though. Let's talk this over with the likes of @stephaniemoss

markdalgleish:
Maybe the best way to handle this is use standard gutters, then give the logo its own negative offset by moving it left a few pixels? Seems to be the right separation of concerns here.

markdalgleish:
@michaeltaranto any thoughts on this?

michaeltaranto:
Happy to close this and revisit later on when we have greater bandwidth for the required discussions

PR #250, opened 2017-07-20
fix(Header): Fix gutters in Header for large devices

michaeltaranto:
Fixes the unexpected indentation of the logo and user account area in the banner on large devices. Also includes some specific media queries to keep the banner content from colliding with the edges of the window at certain resolutions.

### DIFF
--- a/react/Header/Header.less
+++ b/react/Header/Header.less
@@ -39,6 +39,9 @@
   @media only screen and (min-width: (@grid-container-width)) { //IE11 flexbox horizontal centering
     flex: 0 0 @grid-container-width;
   }
+  @media only screen and (min-width: (@grid-container-width + @grid-gutter-width)) {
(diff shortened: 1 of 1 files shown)


react/Header/Header.less
```diff
@@ -162,6 +168,9 @@
 .locale {
   position: absolute;
   right: 0;
+  @media only screen and (min-width: @responsive-breakpoint) and (max-width: (@grid-container-width + @grid-gutter-width - 1)) {
```
kosanna: Maybe a mixin?

michaeltaranto: I thought about that but it's quite bespoke. It's a strange range between device and grid container width.

kosanna: Ah it's -1....ouch.....

PR #252, opened 2017-07-21
feat(Autosuggest): Allow disabling of page scroll while open

kosanna:
And here comes the part when I need this to wire it up to open/close state of autosuggest https://github.com/moroshko/react-autosuggest/issues/404 (404...lol).

Autosuggest stuff might go in a separate PR, but it was good to demo it there, and it's not huge, so...

### DIFF
--- a/react/DisableBodyScroll/DisableBodyScroll.js
+++ b/react/DisableBodyScroll/DisableBodyScroll.js
@@ -15,7 +15,7 @@ DisableBodyScroll.propTypes = {
   isDisabled: PropTypes.bool
 };
 
-function reducePropsToState(propsList) {
+const reducePropsToState = propsList => {
   let isDisabled;
 
   propsList.forEach(props => {
@@ -23,7 +23,7 @@ function reducePropsToState(propsList) {
   });
 
   return isDisabled;
-}
+};
 
 const handleStateChangeOnClient = isDisabled => {
   Object.assign(document.body.style, { overflow: isDisabled ? 'hidden' : 'auto' });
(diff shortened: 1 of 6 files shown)


react/DisableBodyScroll/DisableBodyScroll.js
```diff
@@ -0,0 +1,35 @@
...
+
+  propsList.forEach(props => {
+    isDisabled = props.isDisabled;
+  });
+
+  return isDisabled;
+};
+
+const handleStateChangeOnClient = isDisabled => {
+  Object.assign(document.body.style, { overflow: isDisabled ? 'hidden' : 'auto' });
```
michaeltaranto: Could be better to hoist this ternary higher and only apply `overflow: hidden` if `isDisabled` is true and do nothing if false rather than explicitly setting it to `auto`. Just in case it was being set differently elsewhere, makes it less opinionated.

react/StyleGuideProvider/StyleGuideProvider.js
```diff
@@ -21,19 +23,21 @@ export default function StyleGuideProvider({ fullScreen, children, meta, link, t
...
-        link={link}
-      />
-
-      <ScreenReaderOnly>
-        <h1>{pageTitle}</h1>
-      </ScreenReaderOnly>
-
-      {children}
-    </div>
+    <DisableBodyScroll>
```
michaeltaranto: What is the reason for having `DisableBodyScroll` at this level?

kosanna: Because it aggregates styles so you need something to be on a higher level.

react/DisableBodyScroll/DisableBodyScroll.js
kosanna: You need it to be like that because you won't be able to toggle it otherwise. It will just get stuck with 'hidden' then

PR #256, opened 2017-07-24
feat(Checkbox): Allow checkbox to be wrapped in a component passed as a wrapper

kosanna:
Tests update.
Will be passing NavLink for Classifications filter

### DIFF
--- a/react/Checkbox/Checkbox.demo.js
+++ b/react/Checkbox/Checkbox.demo.js
@@ -6,6 +6,8 @@ import demoStyles from './Checkbox.demo.less';
 import styles from './Checkbox.less';
 import classnames from 'classnames';
 
+const wrapperClickHandler = () => console.log('Wrapper click handler is called')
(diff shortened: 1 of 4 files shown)


markdalgleish:
Sorry, not sure I understand the requirements for this change, might need to take a discussion about this offline first.

kosanna:
@markdalgleish on classifications checked/unchecked we need to track stuff and we need it to use checkbox component from style-guide, so we need to pass a wrapping navlink...as an option we could just try and wrap in locally though...

markdalgleish:
Doesn't this mean that we'll be rendering a *real* checkbox inside of a link?

kosanna:
Not necessary, it could be inside of a div, because navlink accepts components too... actually, it might work if we pass checkbox to navlink as well

kosanna:
Will close it for now and try out a workaround

PR #257, opened 2017-07-24
feat(Header): Add ability to switch saved/applied links on auth status

dborysiewicz:
This is finally ready to go now, old PR was at https://github.com/seek-oss/seek-style-guide/pull/220

Moved authStatusTypes out of /header to /private to make it reusable for both header & footer, added authStatus to Footer & FooterLink and lastly check authStatus to change saved & applied links depending on status in both Header & Footer.

Only issues I see in this one is that the moving of authStatusTypes might break other apps if people where reading them in?
The other is Footer now also requires authStatus to be passed in however it will fail gracefully and revert with the default props.


react/Footer/FooterLink/FooterLink.js
```diff
@@ -36,9 +38,16 @@ FooterLink.propTypes = {
     PropTypes.arrayOf(PropTypes.node),
     PropTypes.node
   ]),
-  linkRenderer: PropTypes.func.isRequired
+  linkRenderer: PropTypes.func.isRequired,
+  authedHref: PropTypes.string,
```
michaeltaranto: This api seems a bit unusual to me. I feel there may be a more generic solution here. Maybe `authRequired` as a boolean and based on this is will either honour your `href` or set it as a returnUrl for the sign-in path. Thoughts?

dborysiewicz: I'm open to suggestions however in this case it's not just setting the returnURL from /sign-in it's also adding the param `nudge=saved` so `https://www.seek.com.au/sign-in?nudge=saved`.

markdalgleish: @dborysiewicz since it seems we're removing the nudge parameter, is this still true?

dborysiewicz: This had changed to `authedHref: PropTypes.bool` this is a outdated code.

michaeltaranto: I agree that the name doesnt fit the function at the moment. It sounds like it takes a different href rather than a boolean. What are your thoughts on something like `authRequired`?

dborysiewicz: Yep I think `authRequired` fits the bill better, I'll switch it over now. Thanks!

react/Header/UserAccountMenu/UserAccountMenu.js
```diff
@@ -78,7 +78,7 @@ export default function UserAccountMenu({ locale, authenticationStatus, linkRend
           linkRenderer({
             'data-analytics': 'header:saved+jobs',
             className: `${styles.item} ${styles.subItem}`,
-            href: '/my-activity/saved-jobs',
+            href: authenticationStatus !== AUTHENTICATED ? '/sign-in?returnUrl=/my-activity/saved-jobs' : '/my-activity/saved-jobs',
```
markdalgleish: Could this ternary be factored out into a function to get rid of the URL duplication? e.g. `urlForAuthStatus(authenticationStatus, '/my-activity/saved-jobs')` (happy for you to come up with something better, that's just to illustrate my point)

PR #259, opened 2017-07-24
fix(Checkbox): Update checkbox styles

kosanna:
- Add activeRoot demo
- Make active and focus states be the same on checkbox
- Organise active/focused state reset to a mixin
- Make checkbox background colour always white
- Make tick be the same colour as focus/active outline with potential opt-out from them being the same


react/Checkbox/Checkbox.less
```diff
@@ -40,16 +40,16 @@
...
+    background-color: @sk-white;
   }
 }
 
 .checkMark {
   display: block;
   width: 100%;
   height: 100%;
-  fill: @sk-black;
+  fill: @sk-focus;
```
markdalgleish: This isn't really semantically correct, because if we tweak our focus styles, it would be unexpected to *also* be changing the colour of a tick. We might need to introduce another variable here, like `@sk-form-accent`?

kosanna: The point of this change was to make them aligned. So it should be sk-focus in case we do change focus color they are still aligned.

markdalgleish: Can probably leave this as-is for now, but this will probably need to trigger a design discussion about whether our focus colours should be coupled to the form element design in this way, cc @stephaniemoss

michaeltaranto: That's fine if they are deliberately in sync at the moment. But going with `@sk-form-accent` future proofs the api. You can set `@sk-form-accent: @sk-focus` to allow the flexibility to change this independently later.

kosanna: Good point @michaeltaranto thought about that. We already had the design discussion re that and it was how I got it, so they always should be in sync...

kosanna:
Moved out to a variable

michaeltaranto:
I really like the new styles, looks a lot fresher. Playing around with the demo site I feel like we should have and active state that adds the accent border too. Without it there is a flicker effect when toggling the checkbox. What do you think?
http://seek-style-guide--14b0015b0121a313adbf09102621f220e54ac9b5.surge.sh/checkbox

react/Checkbox/Checkbox.less
```diff
@@ -40,16 +40,16 @@
     .focus();
   }
 
-  .input:checked + .label & {
-    background-color: @sk-mid-gray-light;
+  .label & {
```
markdalgleish: Does this need to be nested anymore, since it doesn't depend on the `:checked` state?

kosanna:
@michaeltaranto might lead to confusion between active and focused...thoughts?

react/Checkbox/Checkbox.less
```diff
@@ -1,5 +1,14 @@
 @import (reference) "~seek-style-guide/theme";
 
+.activeFocusReset() {
+  .input:not([readonly]):not([disabled]):focus + .label &,
+  .input:not([readonly]):not([disabled]):active + .label & {
```
kosanna: New version of styling correctly handles this indentation. Older one wanted 4 spaces here

react/Header/UserAccount/UserAccount.less
```diff
@@ -62,7 +63,7 @@
   .smallDeviceOnly({
     transform: translateY(@grid-row-height * -1);
     transition: opacity 0.25s ease, transform 0.25s ease;
-    top: (@grid-row-height * 7); // header height
+    top: @headerHeightMobile;
```
kosanna: New version of styling complains on space after declaration, so instead of having a comment a decided to have a variable here, as it's anyway more correct. Scope creep

react/Header/Header.less
```diff
@@ -1,5 +1,8 @@
 @import (reference) "~seek-style-guide/theme";
 
+@headerHeightDesktop: @grid-row-height * 9;
+@headerHeightMobile: @grid-row-height * 7;
```
markdalgleish: 👏 👏 👏

react/Header/UserAccount/UserAccount.less
markdalgleish: Yeah, this is actually a really nice cleanup 👍 

Thanks stylelint! 💯

Issue #258, opened 2017-07-24
Autosuggest styling broken


LeviSeek:
if there are 2 text fields(in this case, job title/company name) with autosuggest, the styling for the 1st autosuggest list will be broken.

![image](https://user-images.githubusercontent.com/16127935/28511785-1255fba6-7092-11e7-880c-0e53e5ba5e1c.png)

georgespyropoulos:
Looks like the issue is that the z-index on the `.suggestionsContainer` class was removed in this PR https://github.com/seek-oss/seek-style-guide/pull/241/files

kosanna:
Looking

kosanna:
Hm...@georgespyropoulos it should not be an issue as it's been transferred to a higher level container

kosanna:
Will do a fix in a sec

markdalgleish:
Looks like this is fixed, so closing the issue.

PR #260, opened 2017-07-25
fix(Autosuggest): Fix overlaying of autosuggest fields

kosanna:
https://github.com/seek-oss/seek-style-guide/issues/258

### DIFF
--- a/react/Autosuggest/Autosuggest.demo.js
+++ b/react/Autosuggest/Autosuggest.demo.js
@@ -49,6 +49,15 @@ class AutosuggestContainer extends Component {
           }}
           onClear={this.handleClear}
         />
+        <DemoComponent
+          inputProps={{
+            type: 'search',
+            onChange: this.handleChange,
+            onFocus: this.handleFocus,
+            value: inputValue
+          }}
+          onClear={this.handleClear}
+        />
       </div>
     );
   }
--- a/react/Autosuggest/Autosuggest.demo.js
+++ b/react/Autosuggest/Autosuggest.demo.js
@@ -49,15 +49,6 @@ class AutosuggestContainer extends Component {
           }}
           onClear={this.handleClear}
         />
-        <DemoComponent
-          inputProps={{
-            type: 'search',
-            onChange: this.handleChange,
-            onFocus: this.handleFocus,
-            value: inputValue
-          }}
-          onClear={this.handleClear}
-        />
       </div>
     );
   }
--- a/theme/mixins/backdrop.less
+++ b/theme/mixins/backdrop.less
@@ -7,5 +7,5 @@
   background-color: @sk-black;
   opacity: 0.7;
   transition: opacity 0.3s;
-  z-index: @z-index-negative;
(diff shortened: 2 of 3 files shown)


react/Autosuggest/Autosuggest.demo.js
```diff
@@ -49,6 +49,15 @@ class AutosuggestContainer extends Component {
           }}
           onClear={this.handleClear}
         />
+        <DemoComponent
```
georgespyropoulos: The 2nd demo component has the same value as the first one - would it be possible for them to be independent?

theme/mixins/backdrop.less
```diff
@@ -7,4 +7,5 @@
   background-color: @sk-black;
   opacity: 0.7;
   transition: opacity 0.3s;
+  z-index: @z-index-negative;
```
georgespyropoulos: I'm not feeling all that comfortable introducing this... any element on a page with an autosuggest with a z-index will sit higher than the backdrop right?  Perhaps another solution could be to put a z-index of 2 on the suggestions container and leave the z-index of 1 on the container?

kosanna: Tried it doesn't work. Backdrop is still kinda work in progress, and you can opt-in when using autosuggest, so it's better to have it for now, till we have access to isOpen state of autosuggest then we could twick z-indexes based on it.

react/Autosuggest/Autosuggest.demo.js
kosanna: I don't want to have it in demo at all being hones, just for demonstration purpose, will merge without it

georgespyropoulos: 👍

theme/mixins/backdrop.less
georgespyropoulos: Yeah I see the problem... so I guess another way to fix this is to have the consuming app provide a z-index value to indicate which box should sit over the other one.

I think using a negative z-index value for the backdrop will almost certainly cause a problem for someone using the backdrop at some point in the future.  If that happens I'm worried that the fix (which would involve removing the negative z-index) could accidentally re-introduce this bug and we're back where we started.  This bug breaks 2 apps that showcase owns and I'd really like for it to not break again

kosanna: Houston will be the first consumer of that I believe. So we'll test this out, for now, this option won't cause issues for houston as we'll be locking scroll when backdrop is shown. Later we will fix it when we merge some stuff to react-autosuggest and it will be good for everyone

kosanna: After this PR it's not breaking current behaviour but everyone else is discouraged to opt-in to backdrop until work on it is fully-finished

kosanna:
OK, after a discussion, we're leaving this PR open till we have a proper fix with isOpen

kosanna:
Forget about that, I fixed it :D

kosanna:
👍  will remove after second approval so other person could play around too

PR #263, opened 2017-07-25
fix(Autosuggest): Fix suggestions not showing in Firefox

kosanna:
Event is a global variable in Webkit/Blink and Trident engines, but it is not in Gecko so in Firefox :/

https://github.com/eslint/eslint/issues/3964

### DIFF
--- a/react/Autosuggest/Autosuggest.js
+++ b/react/Autosuggest/Autosuggest.js
@@ -96,8 +96,8 @@ export default class Autosuggest extends Component {
   renderInputComponent(inputProps) {
     const { labelProps = {} } = inputProps;
 
-    const onFocus = () => {
-      this.scrollOnFocus();
+    const onFocus = event => {
+      this.scrollOnFocus(event);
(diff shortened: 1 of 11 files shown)


react/Autosuggest/Autosuggest.js
```diff
@@ -97,7 +97,8 @@ export default class Autosuggest extends Component {
   renderInputComponent(inputProps) {
     const { labelProps = {} } = inputProps;
 
-    const onFocus = () => {
+    const onFocus = event => {
+      event.preventDefault();
```
markdalgleish: What is this `preventDefault` achieving?

michaeltaranto:
FYI this broke as as a result of Firefox not having `event` as a global variable. So our linter picks this up in the future we have configured the `no-restricted-globals` rule to disallow use of the global `event`.  This is probably expected behaviour for most, so just enforcing it now.

markdalgleish:
If that's the issue, why is `event.preventDefault` being added in this PR?

PR #264, opened 2017-07-26
test(Checkbox): Migrate tests to enzyme and snapshots

kosanna:


kosanna:
To be reopened somewhen

react/Checkbox/Checkbox.test.js
```diff
@@ -1,126 +1,61 @@
...
-} from 'react-dom/test-utils';
-import { createRenderer } from 'react-test-renderer/shallow';
-import SyntheticEvent from 'react-dom/lib/SyntheticEvent';
-import { findAllWithClass } from 'react-shallow-testutils';
-import Checkbox from './Checkbox';
-
-chai.use(sinonChai);
-
-const eventMatcher = sinon.match.instanceOf(SyntheticEvent);
+import { shallow } from 'enzyme';
```
georgespyropoulos: Curious why you are using `shallow` for this test?  We are using `render` (from enzyme) for other tests in the styleguide

react/Checkbox/Checkbox.test.js
```diff
@@ -1,126 +1,61 @@
...
-
-  function render(jsx) {
-    element = jsx;
-    checkbox = renderer.render(element);
-    input = findAllWithClass(checkbox, 'input')[0] || null;
-    label = findAllWithClass(checkbox, 'label')[0] || null;
-    standard = findAllWithClass(checkbox, 'standard')[0] || null;
-    button = findAllWithClass(checkbox, 'button')[0] || null;
-  }
+  const simpleProps = {
```
georgespyropoulos: Perhaps it would make more sense to call this `requiredProps` given these props are marked as required in the prop types

react/Checkbox/Checkbox.test.js
```diff
@@ -1,126 +1,61 @@
...
-        id: 'still-in-role',
-        label: 'still in role',
+        ...simpleProps,
         inputProps: {
           onChange: handleChange,
           checked: false
         }
       };
-      renderToDom(<Checkbox {...props} />);
+      const check = { target: { checked: true } };
```
georgespyropoulos: I would find this test easier to read if this const was called something like `checkedEvent`.  The name seems a bit misleading to me

kosanna: Ok

react/Checkbox/Checkbox.test.js
kosanna: We were using render because we wanted to keep parity with previously written tests that had snapshots. Here's not need to use render as we don't want the full dom render

react/Checkbox/Checkbox.test.js
kosanna: That's why they are simple...just used to that in many projects, looks like some standard thing

georgespyropoulos: Right, I didn't understand that so perhaps it would be clearer if we just called it required props so you don't need to know that simple = required

kosanna:
I wish pullaprove is smart enough to recognise variables renaming :D

kosanna:
Ah it actually not there, cool

react/Checkbox/Checkbox.test.js
georgespyropoulos: So you're using render so `CheckMarkIcon` doesn't get rendered in the snapshots?  I think it'd be good for everyone to agree on an approach to this.  In profile we are explicitly mocking any components that we don't want to render.  Personally I prefer this approach but I'm keen to hear other opinions on this

kosanna: It's rendered as <CheckMarkIcon>, just not as a `<div>`with all the things inside. It's responsibility of CheckMarkIcon snapshots to capture it's guts

georgespyropoulos: Yep I get that and agree that that's the best way for snapshots to be formatted.  In profile we explicitly mock out any subcomponents like this `jest.mock('./ResumeItem/ResumeItem', () => 'ResumeItem');` rather than doing a shallow render.  I prefer to do the mocking manually so that it's clear which sub-components are being mocked and it gives you the flexibility mock some sub components and not others

kosanna: Well, that's a bigger discussion I guess. So I'm merging this PR so it's not continued to be discussed here. However, my opinion is that doing component mocks + render is exactly what shallow is made for (so you don't need mocks)

PR #266, opened 2017-07-26
feat(Autosuggest): Transparent backdrop when Autosuggest is open

kosanna:


theme/mixins/backdrop.less
```diff
@@ -1,11 +1,12 @@
-.backdrop() {
+@backdrop-opacity: 0.7;
+
+.transparrentBackdrop() {
```
michaeltaranto: Curious as to why you are removing the standard backdrop mixin? Sounds like we are looking for either one configurable mixin or two separate mixins — one transparent, one 70% @sk-black.

(Also you have a typo in `transparent` here — double 'r')

react/Autosuggest/Autosuggest.js
```diff
@@ -152,7 +152,12 @@ export default class Autosuggest extends Component {
           ref={this.storeInputReference}
           {...allAutosuggestProps}
         />
-        {showMobileBackdrop ? <div className={styles.autosuggestBackdrop} /> : null }
+        <div
+          className={classnames({
+            [styles.autosuggestBackdrop]: true,
+            [styles.mobileBackdrop]: showMobileBackdrop
```
michaeltaranto: Not sure if this is personal preference or not, but I try to see our CSS module naming as a natural extension of BEM, but for components — component, element, modifier. So there is no need to reiterate `autosuggest` as we are inside the bounds of the component, the element is the `backdrop` and you have `mobile` modifier. Here's my suggestion for `Autosuggest.less`:
```css
.backdrop { ... }
.backdrop_isMobile { ... }
```
Thoughts?

kosanna: we've done it because of naming clashing when it's just backdrop, now as we have transparent we can just do backdrop 👍

theme/mixins/backdrop.less
```diff
@@ -1,11 +1,15 @@
-.backdrop() {
+.transparentBackdrop() {
   position: fixed;
   top: -999px;
   left: -999px;
   right: -999px;
   bottom: -9999px;
   background-color: @sk-black;
```
michaeltaranto: The `background-color` should be in the `shadedBackdrop` mixin too, arguably along with the `transition` as well.

kosanna:
This one is also ok

theme/mixins/backdrop.less
```diff
@@ -1,11 +1,15 @@
...
   left: -999px;
   right: -999px;
   bottom: -9999px;
-  background-color: @sk-black;
-  opacity: 0.7;
   transition: opacity 0.3s;
   z-index: @z-index-inline-overlay;
 }
+
+.shadedBackdrop() {
```
markdalgleish: Should we rename this to `backdrop`?  Having a background colour seems like a sensible default, and this was the old API too.

kosanna: we can't, because there will be a name clash. Alternatively we can rename the file and then it should be fine, let's call it backdrops.less then?

kosanna: ah. ok. Got ya

kosanna: Done 👍

theme/mixins/backdrop.less
```diff
@@ -1,11 +1,16 @@
-.backdrop() {
+.transparentBackdrop() {
```
tamoore: should this also have `pointer-events` disabled?

tamoore: Just confirmed we don't actually need this so all good ship it!

PR #268, opened 2017-07-27
feat(Autosuggest): Add optional prop that allows to lock scrolling on mobile when suggestions are shown

kosanna:


kosanna:
Will reopen into master

PR #269, opened 2017-07-27
docs(readme): Suggest installation as runtime dependency

markdalgleish:
We've somehow missed this until recently, but installing the style guide as a dev dependency can cause issues in server rendered applications when running `npm install --only=production`. This can lead to missing dependencies at runtime, which varies depending on the dependencies of your own application. As a result, we're now recommending that style guide consumers *always* install it as a regular dependency instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ If you'd instead like to work on a custom webpack project, you can use [seek-sty
 ## Installation
 
 ```bash
-$ npm install --save-dev seek-style-guide
+$ npm install --save seek-style-guide
 ```
 
 ## Setup


markdalgleish:
@michaeltaranto does this look good to you?

michaeltaranto:
Woah! Huge. Yeah taking this for a test spin tomorrow, I got the branch ready this afternoon to test it out.

markdalgleish:
What a troll 😛

michaeltaranto:
Slightly more controversial but once this becomes the norm we should pull the `peerDependency` mess out of `package.json` as that was a stop gap solution for getting around the `devDependency` issue.

PR #270, opened 2017-07-28
feat(Autosuggest): Add scroll lock to mobile backdrop

kosanna:
If you choose to opt in to have a mobile shaded backdrop being shown when suggestions are shown, component will also prevent user from scrolling the page.

It does not affect desktop behaviour or page scroll behaviour if you choose not to opt in to backdrop for mobile widths.

Current behaviour: When suggestions are shown and mobile backdrop is shown page is still scrollable

Desired behaviour: When suggestions are shown and mobile backdrop is shown page is NOT scrollable


react/Autosuggest/Autosuggest.js
```diff
@@ -70,16 +71,26 @@ export default class Autosuggest extends Component {
...
     };
 
     return (
-      <IsolatedScroll {...rest} ref={callRef} children={children} />
+      <IsolatedScroll {...rest} ref={callRef}>
+        {children}
+        {
+          areSuggestionsShown &&
+          smallDeviceOnly() &&
+          showMobileBackdrop ?
```
michaeltaranto: Super minor, but maybe change the order of this to short circuit on `showMobileBackdrop` first?

kosanna: 👍

kosanna: Doooone :)

PR #271, opened 2017-07-28
fix(Autosuggest): Scroll through options on mobile if they overflow

kosanna:
Currently user can't observe full suggestions list if it overflows the screen.
screen shot 2017-07-28 at 11 46 04 am

So as a user I should be able to still browse all options even if they overflow. 
This PR fixes it.


markdalgleish:
[Testing Slack integration...](https://trello.com/c/AikFgBNO/1-testing-slack-integration)

PR #272, opened 2017-07-31
fix(Autosuggest): Bind scrolling to suggestions shown event

kosanna:
Currently if user focuses on autosuggest field browser scrolls to the field.

However, when user selects a suggestion then scrolls a bit and then clears suggestions, browser won't scroll back to the field as field is still focused and this focus even is not called, which leads to incorrect behaviour in conjunction with scroll locking on mobile.

This is also the case when you select a suggestion, scroll and then trying to change suggestions again without blurring out of the field. Overlay will be shown, scroll lock will happen, however, browser won't scroll to the field.

This PR binds scrolling to the event when suggestions are shown instead of the focus event.

### DIFF
--- a/react/Autosuggest/Autosuggest.js
+++ b/react/Autosuggest/Autosuggest.js
@@ -7,7 +7,6 @@ import ReactAutosuggest from 'react-autosuggest';
 import IsolatedScroll from 'react-isolated-scroll';
 import ScrollLock from 'react-scrolllock';
 
-import invoke from 'lodash/invoke';
 import omit from 'lodash/omit';
 
 import TextField from '../TextField/TextField';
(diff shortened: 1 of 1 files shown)


react/Autosuggest/Autosuggest.js
```diff
@@ -51,21 +50,23 @@ export default class Autosuggest extends Component {
...
-  constructor() {
-    super();
+  state = {
+    areSuggestionsShown: false
+  }
 
-    this.storeInputReference = this.storeInputReference.bind(this);
-    this.storeTextFieldReference = this.storeTextFieldReference.bind(this);
-    this.renderInputComponent = this.renderInputComponent.bind(this);
+  componentWillUpdate(nextProps, nextState) {
```
mattsjones: If we used `componentDidUpdate` here instead could we avoid the use of `setState`?

kosanna: @mattsjones how would you get the open/close state of the autosuggest in this case?

mattsjones: 🤔  Seems you're right. Was thinking this component owned the the suggestions container for some reason.

PR #273, opened 2017-07-31
BREAKING CHANGE: fix(Autosuggest): label props merge correction

kosanna:
Currently labelProps are not defined as an Autosuggest component's props. 
Component assumes that it comes via `inputProps={labelProps: ...}` which is incorrect because it's a direct prop, thus classNames on label getting merged incorrectly at the moment.

This PR fixes it.


react/Autosuggest/Autosuggest.js
```diff
@@ -122,10 +124,10 @@ export default class Autosuggest extends Component {
     };
 
     const allInputProps = {
+      ...omit(this.props, [ 'inputProps', 'labelProps', 'autosuggestProps' ]),
       ref: this.storeTextFieldReference,
       inputProps: enrichedInputProps,
-      labelProps: enrichedlabelProps,
-      ...omit(this.props, [ 'inputProps', 'autosuggestProps' ])
```
michaeltaranto: What is the reasoning behind changing the order of this spread?

react/Autosuggest/Autosuggest.demo.less
```diff
@@ -0,0 +1,3 @@
+.listItemLabel {
+  color: white;
```
michaeltaranto: Also maybe it's better to not even demonstrate this on the demo site. The demo site is more about visually exploring a component and it's api rather than exploiting every code path like a test suite would.

react/Autosuggest/Autosuggest.js
mengtzu: Does it make any difference?  Are you expecting to overwrite anything with the results of the omit (or vice versa?)

michaeltaranto: Spreading first and last could yield different output yes. Im not expecting there to be any overriding by design, but wanted to know if the reason for moving the spread was to fix a problem caused by overriding.

react/Autosuggest/Autosuggest.demo.less
kosanna: Ok, I'll remove the demo for this thing

react/Autosuggest/Autosuggest.js
kosanna: No reason, just dragged it...that was causing some overwriting before I added 'lableProps' that's why it ended up in another place

markdalgleish:
One thing to be careful of here is that, even though this is fixing an issue, it's technically a breaking API change.

kosanna:
@markdalgleish yep, thought about that. Should probably check other usages if they are different from Chalice

markdalgleish:
Just noticed the label is now white in the style guide demo page: 

screen shot 2017-08-02 at 12 52 54 pm

react/Autosuggest/Autosuggest.js
```diff
@@ -122,10 +124,10 @@ export default class Autosuggest extends Component {
     };
 
     const allInputProps = {
```
markdalgleish: I actually thought there was a bug here for a while because this had been incorrectly named in the past. This shouldn't be `allInputProps` as it's not for the input—it should be `textfieldProps`.

markdalgleish: Even though this wasn't introduced in this PR, could you clean this up? Otherwise it looks like an error that we're passing `labelProps` through via `allInputProps` (which is the issue this PR is supposed to fix)

PR #274, opened 2017-08-01
fix(Scrollock): Allow scrolling on mobile devices when Header or Autosuggest overflow the screen size

kosanna:
Given I'm a user on iPhone I can't scroll within the list.

Issue: if header menu or Autosuggest list overflows the screen and screen is locked with Scrolllock component, scroll events work on inner containers, but touch scroll don't.

This PR fixes it.


react/Autosuggest/Autosuggest.js
```diff
@@ -81,14 +81,16 @@ export default class Autosuggest extends Component {
       }
     };
 
+    const suggestionsContainer = document.getElementById(containerProps.id);
```
mengtzu: You have a guard on the menu version of this, is it not necessary here?

kosanna: I don't get you... ?

mengtzu: The `typeof document !== 'undefined' &&`

kosanna: It's in renderSuggestionsContainer method so I don't need it here, technically....but that's a good point, need to check

kosanna: you're totally right

react/Autosuggest/Autosuggest.js
```diff
@@ -81,14 +81,16 @@ export default class Autosuggest extends Component {
       }
     };
 
+    const suggestionsContainer = typeof document !== 'undefined' && document.getElementById(containerProps.id);
```
michaeltaranto: Not sure i'm keen on having every consuming component needing to implement this. Feel like we should make a higher order component that implements scroll lock and makes `ref` a required property. Then we can resolve the dom node from the ref internally to the HOC. Thoughts?

kosanna: Might be tricky as you probably don't know which area should be scrollable.

kosanna: Got ya, doing

react/Autosuggest/Autosuggest.demo.js
```diff
@@ -63,7 +63,7 @@ export default {
     id: 'jobTitles',
     label: 'Job Titles',
     autosuggestProps: {
-      suggestions: ['Developer', 'Product manager', 'Iteration manager', 'Designer'],
+      suggestions: ['Developer', 'Product manager', 'Iteration manager', 'Designer', 'UX', 'Cleaner', 'HR manager'],
```
markdalgleish: "UX" isn't a job title 😜

kosanna: I'm happy to put there whatever :) `apple banana melon watermelon` :D

react/index.js
```diff
@@ -7,6 +7,7 @@ export { default as Rating } from './Rating/Rating';
 export { default as Header } from './Header/Header';
 export { default as Footer } from './Footer/Footer';
 export { default as Loader } from './Loader/Loader';
+export { default as DisableScrollOutside } from './DisableScrollOutside/DisableScrollOutside';
```
markdalgleish: Not sure this should be exported, seems more like a utility than a company specific style.

kosanna: classifications :trollface:

kosanna:
Closing for now as has issues

PR #279, opened 2017-08-03
feat(Checkbox): Add support for node labels and readonly checkboxes

kosanna:
Why: due to SEO/other reasons it might be desirable to have a link under the hood of checkbox, having a checkbox as a clickable and controlled input field.

Following things has been done to achieve it:
- make label to be of a non-zero height (because the label should be clickable)
- change label PropType to be node instead of string (because link might be passed)
- remove onChange from required input props (because it might be a controlled input)


react/Checkbox/Checkbox.js
```diff
@@ -22,10 +22,10 @@ export default class Checkbox extends Component {
 
   static propTypes = {
     id: PropTypes.string.isRequired,
-    label: PropTypes.string.isRequired,
```
georgespyropoulos: Can we still allow strings as well as nodes for the label prop?

react/Checkbox/Checkbox.js
```diff
@@ -22,10 +22,10 @@ export default class Checkbox extends Component {
...
+    label: PropTypes.node.isRequired,
     className: PropTypes.string,
     inputProps: PropTypes.shape({
-      onChange: PropTypes.func.isRequired,
```
georgespyropoulos: It'd be great to make this prop type a function that checks if the label is a string and makes the prop required if it is

react/Checkbox/Checkbox.js
kosanna: string should be trated as a node, isn't it? 🤔

react/Checkbox/Checkbox.js
kosanna: This is not necessary, you might still want label as a string but a controlled input

react/Checkbox/Checkbox.js
georgespyropoulos: Ah yes that's true.  Perhaps the prop type should be one of string or element then to make it a bit stricter?

kosanna: https://github.com/facebook/flow/blob/v0.50.0/lib/react.js#L12

kosanna: That won't make it stricter...

georgespyropoulos: Well there you go, a string is also valid for an element prop type.  One last thing - I realise it wouldn't make any difference to how it's evaluated but would it still be worth making this a oneOf string or element prop type for the sake of documenting the fact that a string is valid input?

react/Checkbox/Checkbox.js
georgespyropoulos: That's not something we currently support though... just trying to make sure we don't lose any prop type checks here.  I'm happy to remove it because I'm not sure there's much value of enforcing it that prop

react/Checkbox/Checkbox.js
kosanna: I don't feel it's a right thing to do as one day someone will come and delete it, thinking `what a stupid thing they've added here`

PR #280, opened 2017-08-03
feat(Logo): Compact Logo Variation

tamoore:
This PR breaks the SEEK logo into separate parts to allow for the conditional rendering of the name. In conjunction, I've created a new prop called compact that is just the roundel with the arrow (i.e. no 'seek').

A couple of things to note about this PR:

1. We probably need to get a different svg group/path for the logo component that's optimised for a square (instead of rectangle) viewport.
2. It opens up the facility to have other "names": "asia", "business" etc.

We can drop the viewport manipulation if you want and then only make that transformation when we have an appropriate logo asset. 

This should all be backwards compatible with the current header.


react/Logo/SeekLogo.js
```diff
@@ -0,0 +1,52 @@
+import React, { PropTypes } from 'react';
+
+const SeekLogo = ({ circleClass = 'sk-logo-circle', dotClass = 'sk-logo-dots' }) => (
+  <g>
+    <g className={circleClass}>
+      <circle cx="41.9" cy="41.9" strokeWidth="0" r="41.9" />
+    </g>
+    <g className={dotClass}>
+      <path d="m42.883 15.978c0-1.447 1.168-2.62 2.612-2.62 1.451 0 2.622 1.173 2.622 2.62 0 1.442-1.171 2.615-2.622 2.615-1.444 0-2.612-1.173-2.612-2.615z" />
```
tamoore: Creates a lot of jest snapshot noise, a possible alternative here would be dangerously setting on the `g`.

PR #283, opened 2017-08-04
feat(Z-indexes): Add intervals between z-indexes, add separate z-index for backdrop

kosanna:
Due to multiple stacking elements there is a need in extra z-index inbetween 0 and inline-overlay, so  backdrop has been granted it's own higher z-index thus -inline-overlay could be used


kosanna:
Also not sure if it should be marked as a breaking change again. Might be, potentially if hardcoded z-indexes are used on a project apart from variables

PR #284, opened 2017-08-04
feat(Checkbox): multiline checkbox variant

kosanna:
image

image

Some stylings might look weird in the code, but they are actually required because we might be nesting nodes inside label.


markdalgleish:
Honest question—is there a reason this needs to be a prop? Shouldn't all checkboxes wrap to multiple lines, if needed?

react/Checkbox/Checkbox.js
```diff
@@ -28,15 +28,17 @@ export default class Checkbox extends Component {
       onChange: PropTypes.func,
       checked: PropTypes.bool.isRequired
     }),
-    type: PropTypes.oneOf([STANDARD, BUTTON])
+    type: PropTypes.oneOf([STANDARD, BUTTON]),
+    isMultiline: PropTypes.bool
```
markdalgleish: Can we drop the `is` prefix from this? We don't tend to prefix props this way (even though I think it makes a lot of sense)

In fact, maybe `wrap` is a more appropriate name, since it's not *necessarily* going to render to lines?

react/Checkbox/Checkbox.less
```diff
@@ -20,8 +22,9 @@
 
 .label {
   display: block;
+  height: 100%;
 
-  :checked + & {
+  :checked + & .labelContainer {
```
markdalgleish: Why was this change needed?

react/Checkbox/Checkbox.less
```diff
@@ -34,14 +37,20 @@
   @media @desktop {
     .touchableText(@standard-type-scale);
   }
+
+  &.isMultiline {
+    .standardTextResponsive(@baseline: false);
```
markdalgleish: We shouldn't be adding more of usages these mixins, they've been deprecated in favour of the `<Text>` component.

markdalgleish:
Just noticed the text isn't on grid, would you be able to fix this? :)

markdalgleish:
We've shipped an interim solution for this, so closing for now.

PR #285, opened 2017-08-06
chore(GitHub): Add initial pull request template

markdalgleish:
This is totally up for discussion, but figured I'd get the ball rolling with a first pass.

We've had some issues with our automatically generated release notes being too sparse, particularly when it comes to breaking changes, and confusion for contributors around what level of detail is required to address this.

So, to help guide contributors, I've added a pull request template which will show up in the description box when opening a PR.

Let me know what you think. Keen to discuss this more, but hoping to at least get a good initial version in place.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,8 +1,9 @@
-Based on the nature of your change, please use the relevant template below and modify to suit your needs.
+IMPORTANT
+=========
 
-Please ensure you copy the completed "Commit Message For Review" section below into your final commit body when merging to master. This allows us to validate that the generated release notes are clear enough for consumers to follow.
+Please ensure that, when eventually merging this PR to master, you copy the completed "Commit Message For Review" section below into your final commit body. This allows us to review our release notes, validating that they're clear enough for consumers to follow.
 
-Don't forget that what's written here forms the upgrade guide for consumers, and upgrading can easily become a long and arduous process. To counter this, please be as descriptive as possible, explaining the reasoning behind the change. Since there are many different consumers, changes should be documented from the perspective of the style guide itself, so avoid making specific reference to the application you're working on.
+Don't forget that what's written here forms the upgrade guide for consumers, and upgrading can easily become a long and arduous process. To counter this, please be as descriptive as possible, explaining the reasoning behind the change, and showing a clear usage example or migration guide where appropriate. Since there are many different consumers, changes should be documented from the perspective of the style guide itself, so avoid making specific reference to the application you're working on.
 
 
 Major release template
@@ -68,6 +69,7 @@ REASON FOR CHANGE:
 
 Non-release template
 ====================
+
 
 ## Commit Message For Review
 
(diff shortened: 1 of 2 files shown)


markdalgleish:
Also note that I've updated our contribution guide, which we'll also need to discuss.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,74 @@
...
+
+Major release template
+======================
+
+
+## Commit Message For Review
+
+BREAKING CHANGE:
+
+{{ Description of breaking API change }}
```
tamoore: Assuming the handlebars are for github?

markdalgleish: Nope, the idea is that the person opening the PR will replace these. Wasn't sure of the best syntax for this, but this seemed good enough.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,74 @@
+Based on the nature of your change, please use the relevant template below and modify to suit your needs.
```
tamoore: I kind of feel like this intro is a little too much, couldn't this all go in the contribution guidelines and then just be a link with a pointer to the heading of this info?

markdalgleish: Yeah, this is the one thing I wasn't sure about. We could streamline this to a simple one-line instruction?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,74 @@
...
+
+Please ensure you copy the completed "Commit Message For Review" section below into your final commit body when merging to master. This allows us to validate that the generated release notes are clear enough for consumers to follow.
+
+Don't forget that what's written here forms the upgrade guide for consumers, and upgrading can easily become a long and arduous process. To counter this, please be as descriptive as possible, explaining the reasoning behind the change. Since there are many different consumers, changes should be documented from the perspective of the style guide itself, so avoid making specific reference to the application you're working on.
+
+
+Major release template
```
tamoore: I think the break down of each template is a bit overkill, couldn't we progressive enhance the template ;)

markdalgleish: I hear what you're saying, but I'm trying to minimise the amount of messing around you have to do when opening a PR. I want to make it as simple as deleting everything above and below the relevant section, rather than feeling like documentation surgery 😄

.github/PULL_REQUEST_TEMPLATE.md
tamoore: I'm not the best sub editor in the world, but it could be something as simple as:

"Styleguide is semantically version. If semver is unfamiliar to you, have a read (of)[link]; otherwise choose the right template for your version bump below"

Don't use that! But it's succinct. Thoughts?

.github/PULL_REQUEST_TEMPLATE.md
tamoore: Yeah that's fine, I agree that it could be a little fiddly so just deleting a block is probably easier.

.github/PULL_REQUEST_TEMPLATE.md
markdalgleish: Oh, so your issue isn't with the paragraph below? We can just delete this line, then, since I've updated the contribution guide.

tamoore: Not really, that info is actually very good. My main question is whether it really belongs in a PR template. Maybe it is good to have a reminder about upgrades, but possibly only in the major commit template. (seeming as it really only applies there)

markdalgleish:
@tamoore just pushed an update with some tweaks, let me know what you think.

PR #287, opened 2017-08-09
chore(rfcs): Add RFC process documentation and template

markdalgleish:
This is based on [Yarn's RFC template](https://github.com/yarnpkg/rfcs/blob/master/0000-template.md), which is itself based on [Rust's RFC template](https://github.com/rust-lang/rfcs/blob/master/0000-template.md)

## Commit Message For Review

Adding some more structure around how we propose, discuss and review changes to the public API, rather than waiting until a PR is opened, which is often a bit too late in the process. This has been adapted from Yarn's RFC template, which itself was based on Rust's RFC template.


mengtzu:
Do we need it doubled up in two files?  If it auto-populates the issue I'm not sure we need the more explicitly named but superfluous file

markdalgleish:
Yeah, I didn't plan to do this originally, but I realised that it might be useful to refer back to it while editing, potentially after deleting parts of the template. But on second thought, it probably makes more sense to just link to the issue template 😅

markdalgleish:
@mengtzu just updated the PR.

markdalgleish:
After speaking with @dborysiewicz, I've added a link to [Yarn's list of accepted RFCs](https://github.com/yarnpkg/rfcs/tree/master/accepted) to help people get a sense of what to write 👍

markdalgleish:
Realised we don't explicitly ask for an API example. Might be worth adding this.

markdalgleish:
Just realised that we should also link to the RFC in the final commit message too, so added it to the pull request template.

PR #288, opened 2017-08-10
refactor(index): Sort the exported groups in index.js alphabetically

tamoore:
Sorted the exported groups in index.js alphabetically

REASON FOR CHANGE:

Just feel it's a lil easier to find the component that you're looking for, or, to slot in a new one.

### DIFF
--- a/react/index.js
+++ b/react/index.js
@@ -1,5 +1,5 @@
-export { default as SeekApp } from './StyleGuideProvider/StyleGuideProvider';  // Alias for backwards compatibility
 export { default as StyleGuideProvider } from './StyleGuideProvider/StyleGuideProvider';
+export { default as SeekApp } from './StyleGuideProvider/StyleGuideProvider';  // Alias for backwards compatibility
 
 // Components
 export { default as Button } from './Button/Button';
--- a/react/index.js
+++ b/react/index.js
@@ -1,3 +1,6 @@
+export { default as SeekApp } from './StyleGuideProvider/StyleGuideProvider';  // Alias for backwards compatibility
+export { default as StyleGuideProvider } from './StyleGuideProvider/StyleGuideProvider';
+
 // Components
 export { default as Button } from './Button/Button';
 export { default as Footer } from './Footer/Footer';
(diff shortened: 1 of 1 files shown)


react/index.js
```diff
@@ -1,21 +1,21 @@
...
 export { default as Button } from './Button/Button';
-export { default as Logo } from './Logo/Logo';
-export { default as Rating } from './Rating/Rating';
-export { default as Header } from './Header/Header';
 export { default as Footer } from './Footer/Footer';
+export { default as Header } from './Header/Header';
 export { default as Loader } from './Loader/Loader';
+export { default as Logo } from './Logo/Logo';
+export { default as Rating } from './Rating/Rating';
+export { default as SeekApp } from './StyleGuideProvider/StyleGuideProvider';  // Alias for backwards compatibility
```
markdalgleish: These two probably need to stay up top, since they define the CSS reset which is designed to be overridden.

tamoore: We should break them out and place them at the very top:

```javascript
export { default as SeekApp } from './StyleGuideProvider/StyleGuideProvider'; 
export { default as StyleGuideProvider } from './StyleGuideProvider/StyleGuideProvider';

// Components
```
What do you think?

markdalgleish: Yep, exactly 👍

tamoore: Done 😄

react/index.js
```diff
@@ -1,21 +1,22 @@
-// Components
-export { default as StyleGuideProvider } from './StyleGuideProvider/StyleGuideProvider';
 export { default as SeekApp } from './StyleGuideProvider/StyleGuideProvider';  // Alias for backwards compatibility
+export { default as StyleGuideProvider } from './StyleGuideProvider/StyleGuideProvider';
```
markdalgleish: Haha, sorry, but you you mind flipping these two imports around? It makes for a bit of a pointless diff here 😄

tamoore: Fair point. I have it a totally on message commit message... ha

react/index.js
```diff
@@ -27,36 +28,36 @@ export { default as Section } from './Section/Section';
...
-export { default as Textarea } from './Textarea/Textarea';
 export { default as Checkbox } from './Checkbox/Checkbox';
-export { default as MonthPicker } from './MonthPicker/MonthPicker';
+export { default as ClearField } from './TextField/ClearField/ClearField';
 export { default as Dropdown } from './Dropdown/Dropdown';
+export { default as EmailField } from './EmailField/EmailField';
+export { default as MonthPicker } from './MonthPicker/MonthPicker';
 export { default as SlideToggle } from './SlideToggle/SlideToggle';
+export { default as TextField } from './TextField/TextField';
+export { default as Textarea } from './Textarea/Textarea';
```
markdalgleish: These two are around the wrong way 🕵️

tamoore: Hmm, that's interesting I used unix's `sort` program to do this and it looks like it makes a distinction between a capital letter.

I.e. a capital is greater then a non capital. 

(Learn something new everyday)

PR #290, opened 2017-08-14
docs(site): Ensure demo components re-render correctly

### LINKED ISSUE #276: Demo page isn't updating props but rerendering a full new component
Issues like this https://github.com/seek-oss/seek-style-guide/pull/275 are missed as it works in the demo however when implemented into a real application it dosen't work. This would also be a source of why transitions are not seen.

markdalgleish:
The issue here was that we were creating a new component on every render (`<BoundComponent>`) so that we could bake in props ahead of time, but this was causing the component instance to be constantly destroyed and recreated. This meant that all state, both inside the component and in the DOM, wasn't being retained correctly.

To solve this, instead of binding the props to the component, we pass the props down to the `<Demo>` component and let it combine the props internally.

## Commit Message For Review

Fixes #276

### DIFF
--- a/docs/src/components/Demo/Demo.js
+++ b/docs/src/components/Demo/Demo.js
@@ -8,9 +8,12 @@ import Baseline from 'react-baseline';
 import Code from './Code/Code';
 import flatten from 'lodash/flatten';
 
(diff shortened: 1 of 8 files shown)


markdalgleish:
Just updated the description with some more details about the issue.

PR #294, opened 2017-08-16
docs(site): Tidy up generated code samples

markdalgleish:
Code samples are automatically generated, but this has led to a few quirks that make them differ from how you'd actually write the code in practice. To fix this, this commit makes the following changes:
- Replace all instances of `prop={true}` with `prop`, while leaving `prop={false}`, e.g. `<Text hero={true}>` becomes `<Text hero>`
- Hide any custom styling via the `style` prop, which is used purely for visual purposes on the style guide demo site, e.g. `<Section style={{ background: 'blue' }}>` becomes `<Section>`
- Detect and display *all* style guide components in import statements, not just the root component, e.g. `import { Card } from 'seek-style-guide/react'; <Card><Section>...` becomes `import { Card, Section } from 'seek-style-guide/react'; <Card><Section>...`
- Use multiline import statements to better support smaller screens and multiple imports

## Commit Message For Review

N/A

### DIFF
--- a/docs/src/components/Demo/Code/Code.js
+++ b/docs/src/components/Demo/Code/Code.js
@@ -53,12 +53,12 @@ export default class Code extends Component {
       })
       .replace(/\={true}/ig, '')
       .replace(/svgClassName=".*?"/ig, 'svgClassName="..."')
-      .replace(/function noRefCheck\(\) \{\}/ig, '() => {...}');
+      .replace(/function noRefCheck\(\) \{\}/ig, '() => {...}')
+      .replace('<MockContent />', 'Lorem ipsum');
 
       const componentNames = uniq(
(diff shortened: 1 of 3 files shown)


markdalgleish:
Another example to show how big this improvement is...

Before:

```js
import { PageBlock } from 'seek-style-guide/react';

<PageBlock
  style={{
    background: 'pink'
  }}
>
  <Section
    header={true}
    style={{
      background: 'cyan'
    }}
  >
    <Text
      hero={true}
      style={{
        background: 'blue',
        color: 'white'
      }}
    >
      Lorem ipsum
    </Text>
  </Section>
</PageBlock>
```

After:

```js
import {
  PageBlock,
  Section,
  Text
} from 'seek-style-guide/react';

<PageBlock>
  <Section header>
    <Text hero>
      Lorem ipsum
    </Text>
  </Section>
</PageBlock>
```

PR #295, opened 2017-08-21
feat(Less): Add accessible variant for mid-gray-dark on gray-medium

kosanna:


markdalgleish:
Just a heads up that I renamed this PR from a "refactor" to a "feat" — since this changes the public API, it's a feature release. Also, refactors don't trigger a release, if I remember correctly.

markdalgleish:
Just noticed the readme hasn't been updated with the new colour. Would you be able to add that?

kosanna:
Yup, will update

kosanna:
Updated

markdalgleish:
I don't want to block this PR over it, but we might need to rethink our naming convention around accessible variants. The current approach isn't very scalable at all.

kosanna:
They are all gray. Aren't you bugged with css `color` property, not `colour`?

markdalgleish:
Sorry, my comment was aimed at your spelling in this case—you've done the right thing and remained consistent.

theme/palette/accessible-variants.less
```diff
@@ -1,5 +1,7 @@
 // Accessible variants
 @sk-mid-gray-on-white: #959595;
+@sk-mid-gray-dark-on-gray-medium: #646464;
```
michaeltaranto: As discussed this should be `@sk-mid-gray-dark-on-gray-light` as that is what the [background is set to](https://github.com/seek-oss/seek-style-guide/blob/master/theme/palette/elements.less#L10).

PR #298, opened 2017-08-23
Autosuggest bug case

vgoulko:
## Commit Message For Review

When there's two Autosuggest components on a page, clicking the label on either focusses on the first input.


kosanna:
Could we closed this PR please? As it's been discussed that we don't want such things appear in the demo.

PR #300, opened 2017-08-24
Removing label props to show the warning message

vgoulko:
The bellow warning appears when label props are not specified for autosuggest label

Warning: Failed prop type: Specifying `labelProps` is redundant when `label` is not specified in FieldLabel.

### DIFF
--- a/react/Autosuggest/Autosuggest.demo.js
+++ b/react/Autosuggest/Autosuggest.demo.js
@@ -63,7 +63,6 @@ export default {
   container: AutosuggestContainer,
   initialProps: {
     id: 'jobTitles',
-    label: 'Job Titles',
     autosuggestProps: {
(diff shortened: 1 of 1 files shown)


kosanna:
If we want to remove warning, then we might look at fixing component, not the demo? 🤔

vgoulko:
@kosanna you are right,
I have created this pull request with the hope to create [link](http://seek-style-guide--a4c5d77d1e8f077fb0fb37009eedce16a5a12cff.surge.sh) so it will be possible to reproduce the issue.
I have submitted an item, can remove this pull request if this is unnecessary.

kosanna:
@vgoulko yup, not sure we need a PR here ;)

vgoulko:
Closing this pull request as the purpose was to create a visible reference for the issue.
Have raised the item instead

PR #302, opened 2017-08-24
feat(Header): Reorder sign in, register and sign out links in the mobile header

dborysiewicz:
## Commit Message For Review

Reorder sign in, register, sign out links and added pending animation in the mobile header. Pull out loader less so it is reusable in other components.

REASON FOR CHANGE:

To make sign in/register for candidates (our main audience) on mobile more clear/important then the link to the employer site.

**New Left** vs Old Right
screen shot 2017-08-24 at 3 34 08 pm

Includes the removal of activeTab === 'Employer Site' as it make no sense in the current implementations of Header


react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -122,17 +122,27 @@
...
 }
+
+.mobileSigninSignout {
+  margin-top: @grid-row-height;
+  height: @grid-row-height * 9;
+
+}
+
+.authLoader {
+  .loader(4px, 3px, -4px (@grid-gutter-width * 1.5) 0 0);
```
michaeltaranto: Rather than duplicating the markup and CSS, what do you think of adding an `inline` prop to the `Loader` component? Then we could also use it for the `loading` state in the `Button` component. Thoughts?

react/Header/UserAccountMenu/UserAccountMenu.js
```diff
@@ -165,36 +216,8 @@ export default function UserAccountMenu({ locale, authenticationStatus, linkRend
...
-              }
-              <div className={styles.iconSpacer} />
-            </span>
-          </li>
-        ) : null
-      }
-      {
         authenticationStatus === AUTHENTICATED ? (
-          <li>
+          <li className={`${styles.aboveSmallDevice}`}>
```
kosanna: Is this class doing anything?

kosanna: Could we just replace it with Hidden component?

react/Header/UserAccountMenu/UserAccountMenu.less
dborysiewicz: Adding this now 👍

react/Header/UserAccountMenu/UserAccountMenu.js
dborysiewicz: Only reason was it's the current way the rest of the items in the menu where being hidden/shown - I guess it was done before `Hidden` component existed. I'll switch everything in this component over to `Hidden`.

react/Header/UserAccountMenu/UserAccountMenu.js
```diff
@@ -57,31 +62,35 @@ export default function UserAccountMenu({ locale, authenticationStatus, linkRend
...
-                key="icon"
-                className={classnames(styles.icon, styles.saveSearches)}
-                svgClassName={styles.iconSvg}
-              />
-            ]
-          })
-        }
-      </li>
+
+      <Hidden mobile>
```
kosanna: Sorry for this pick, that was really random, I'm worried that putting ul li inside span doesn't work in all browsers (and it's invalid)...I'll play around there probably. That was bugging my brain long enough, I guess we need Hidden to accept a component

react/Loader/Loader.js
```diff
@@ -1,12 +1,22 @@
...
+      <div className={styles.ball} />
+      <div className={styles.ball} />
+      <div className={styles.ball} />
+    </div>
+  );
+}
 
-export default Loader;
+Loader.propTypes = {
+  inline: PropTypes.bool.isRequired
```
kosanna: There's no need for it to be required prop as we set default value

kosanna:
Is it by design that :hover state is not highlighting from border to border?

kosanna:
Could potentially shave lots more with Hidden, but that's anyway not the purpose of this PR, maybe someone will do that in the future

react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -108,31 +108,27 @@
...
-    display: none;
-  }
-}
-
 .firstItemInGroup {
   @media @mobile {
-    padding-top: @grid-row-height * 2;
+    padding-top: @grid-row-height * 4;
     position: relative;
     &::before {
```
kosanna: I wonder why we're not just doing something like that 
```
    padding-top: @grid-row-height * 2;
    margin-top: @grid-row-height * 2;
    border-top: 1px solid fade(@sk-white, 20%);
```
instead of having all that ::before 🤔

kosanna:
I'm not quite sure that just showing loader is enough 🤔  as it's a bit confusing, I mean as a user I would not understand that it's authorising...but probebly doesn't matter as it's just a weird edge-case

kosanna:
And I'm curious on how we're going to implement pending status on our side as we have authenticated status as false by default

react/Header/UserAccountMenu/UserAccountMenu.less
dborysiewicz: I believe we switched to `::before` as borders take space and put the grid off alignment?

react/Loader/Loader.js
dborysiewicz: Yep good catch I will remove.

dborysiewicz:
@kosanna the design around ::hover & the loading message has been signed off by design - The messaging/loader icon had a bit of back and forth and it's what the settled on for now.

Around the auth status that is fine as the loader is the default for the switch so will work fine?

kosanna:
@dborysiewicz I just mean that no-one will see the loader ever :D

markdalgleish:
Spoke with @dborysiewicz about this, and decided against changing the public `Loader` API until we clearly define the long term direction for this pattern. Since this smaller variant is only being used inside the style guide itself, we've opted for making a pseudo-private `_small` prop as a short-term fix.

PR #303, opened 2017-08-29
feat(Hidden): Add ability to accept a component that will wrap children

kosanna:
## Commit Message For Review

feat(Hidden): Add ability to pass down a component that will wrap children

RFC URL: None

REASON FOR CHANGE:

Additional unnecessary markup should be avoided when using hidden, because it might lead to semantically incorrect markup such as li items wrapped in spans, etc.

EXAMPLE USAGE:

```js
<Hidden component='li'>foo</Hidden>
```

### DIFF
--- a/react/Hidden/Hidden.js
+++ b/react/Hidden/Hidden.js
@@ -5,11 +5,12 @@ import classNames from 'classnames';
 
 import styles from './Hidden.less';
 
-const Hidden = ({ children, component, className, print, screen, mobile, desktop }) => {
+const Hidden = ({ children, component, className, print, screen, mobile, desktop, ...restprops }) => {
   const Component = component ? component : 'span';
 
   return (
     <Component
+      {...restprops}
       className={classNames({
         [className]: className,
         [styles.desktop]: desktop,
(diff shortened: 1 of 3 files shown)


react/Hidden/Hidden.js
```diff
@@ -5,21 +5,30 @@ import classNames from 'classnames';
 
 import styles from './Hidden.less';
 
-const Hidden = ({ children, className, print, screen, mobile, desktop }) => (
-  <span
-    className={classNames(
-      className,
-      {
+const Hidden = ({ children, component, className, print, screen, mobile, desktop, ...restprops }) => {
+  const Component = component ? component : 'span';
```
markdalgleish: Isn't this line redundant with the `defaultProps` defined below?

kosanna: Totally, will remove

kosanna: Done 👍

markdalgleish:
@kosanna before merging, could you please edit your PR description to follow the template? https://github.com/seek-oss/seek-style-guide/blob/master/.github/PULL_REQUEST_TEMPLATE.md

markdalgleish:
Release notes look good 👏👏👏

PR #304, opened 2017-08-30
fix(Header): Update analytics data attr for sign in, register & sign out links

dborysiewicz:
## Commit Message For Review

Prefix the `data-analytics` tag for sign in, register & sign out links in the header with `header:`

REASON FOR CHANGE:

At the moment there is no way to distinguish between different sign in, register & sign out links on the page as they are all tracked as the same thing. The patten of `header:sign-in` follows all the other links in the header.

**While this shouldn't be a breaking change to anybody (hence not marked as one). I have found a issue in the way the homepage team has implemented sign in/register links which ill break. I'm working with them on this issue.**

### DIFF
--- a/react/Header/SignInRegister/SignInRegister.js
+++ b/react/Header/SignInRegister/SignInRegister.js
@@ -16,7 +16,7 @@ export default function SignInRegister({ linkRenderer, returnUrl }) {
       </ScreenReaderOnly>
       {
         linkRenderer({
-          'data-analytics': 'sign-in',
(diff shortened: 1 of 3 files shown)


markdalgleish:
Should this be marked as a fix instead, since the API doesn't change?

markdalgleish:
@dborysiewicz is this PR ready to go?

dborysiewicz:
@markdalgleish just waiting for confirmation from Company Reviews but on my end looks good - hoping to have it merged this afternoon.

PR #306, opened 2017-09-04
feat(ScreenReaderOnly): Add support for different component types

richardPFisk:
### Reason for change

To be able to output valid HTML when used in different contexts. For example:

```
<div className={styles.root}>
  <ScreenReaderOnly component="div">
    <div>some block content</div>
  </ScreenReaderOnly>
</div>
```

```
<span>
  <ScreenReaderOnly>
    <span>some text or inline content</span>
  </ScreenReaderOnly>
</span>
```

ScreenReaderOnly is used as a child of both div and span elements with different child elements of
Block and Inline elements types. To be able to output valid HTML, choosing the component type can
allow this. 

### Backwards compatibility
By retaining default of rendering the children within a span element makes the change
backwards compatible.


richardPFisk:
Thanks for that @mattsjones, I neglected to fill out one of the options when I ran `npm run commit`

markdalgleish:
@richardPFisk feel free to merge this when you're ready 👍

richardPFisk:
@markdalgleish, thanks for the reviuew. I'm currently don't have access to any of the `seek-oss` repositories and can't merge...

richardPFisk:
Thanks 😃

markdalgleish:
@richardPFisk too quick— I was busy writing this message to tell you I sent you an invite :smile:

Issue #307, opened 2017-09-05
RFC: Add 'tertiaryLabel' option to all form fields


dborysiewicz:
# Summary

There is a need to support a extra label (message) in my case is a `TextLink` that is right aligned of text field's primary label.

# Motivation

To reduce the form height by allowing context specific messaging to be positioned directly above the form field, instead of being forced to place it in a separate row below the field, or at the bottom of the form.

For example, this is useful for password reset links:

<img width="430" alt="screen shot 2017-09-05 at 11 03 45 am" src="https://user-images.githubusercontent.com/7741507/30041370-fa76367e-922c-11e7-8b08-92fe852182d9.png">

# Detailed design

This will be implemented in the internal `FieldLabel` component so that all form components have access to this new label style.

In order for it to support links, it should accept any node, whether a string or a React element:

```js
<TextField {...} tertiaryLabel="..." />
```

```js
<TextField {...} tertiaryLabel={<TextLink {...} />} />
```

# How We Teach This

This is a relatively straightforward addition since it follows established patterns (`secondaryLabel` already exists), and fits within a larger pattern of primary / secondary / tertiary variants.

Form field examples on the demo site will need to be updated to feature this new option.

# Drawbacks

This unfortunately adds extra complication to the form field APIs for something that is not very common.

In addition, this is also hard-coding the visual placement of this message, which might not suit future needs and lead to even further API complication. It's possible we might need to revisit this in the future.

# Alternatives

We could provide lower level render control via something like a `labelRenderer` function, but we then run the risk of allowing too much style divergence across projects.

markdalgleish:
I paired with @dborysiewicz on this, so it lgtm 👍 

Keen to get a second opinion on it, though. Does this sound reasonable?

mattsjones:
Sounds good.

PR #308, opened 2017-09-05
feat(FieldLabel): Add tertiary label option

dborysiewicz:
## Commit Message For Review

Add `tertiaryLabel` option to `FieldLabel` to support a extra label (node) that is right aligned of text field's primary label.

RFC URL:

https://github.com/seek-oss/seek-style-guide/issues/307

REASON FOR CHANGE:

To reduce the form height by allowing context specific messaging to be positioned directly above the form field, instead of being forced to place it in a separate row below the field, or at the bottom of the form.

EXAMPLE USAGE:

```js
<TextField {...} tertiaryLabel={<TextLink {...} />} />
```


markdalgleish:
The demo site can get into a broken state with this option:

screen shot 2017-09-05 at 2 36 20 pm

react/private/FieldLabel/FieldLabel.demo.js
```diff
@@ -9,6 +12,13 @@ export default [
           ...props,
           secondaryLabel: '(Max 30 characters)'
         })
+      },
+      {
+        label: 'Tertiary label',
+        transformProps: props => ({
+          ...props,
+          tertiaryLabel: <TextLink>Tertiary TextLink</TextLink>
```
markdalgleish: Could you make this a more realistic example, like the `secondaryLabel` above? Maybe 'Help'? That would probably fix the text wrapping issue, too.

markdalgleish:
Review done. Looking really good, just some minor feedback on the demo site.

markdalgleish:
@dborysiewicz ready to ship this?

PR #318, opened 2017-09-12
fix(RainbowLogo): Only show rainbow logo in AU locale

markdalgleish:
## Commit Message For Review

REASON FOR CHANGE:

This message is irrelevant in New Zealand 🇳🇿


react/Logo/Logo.js
```diff
@@ -15,6 +15,7 @@ const viewportDimensions = type => ({
 const getViewPortDimension = (type, dimension) => viewportDimensions(type)[dimension];
 
 export default function Logo({
+  locale,
```
markdalgleish: FYI I had to add this so we don't get warnings about passing `locale` to the underlying `div`.

PR #319, opened 2017-09-13
fix(standalone): Default standalone header to rainbow logo

dborysiewicz:
## Commit Message For Review

Default the standalone `Header` to use `LogoRainbow`.

REASON FOR CHANGE:

Currently no other way to do this, have discussed with @markdalgleish and this is temporary.

### DIFF
--- a/standalone/header-footer/src/makeStandalone.js
+++ b/standalone/header-footer/src/makeStandalone.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { render } from 'react-dom';
 import StandaloneProvider from './StandaloneProvider/StandaloneProvider';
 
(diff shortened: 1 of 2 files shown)


standalone/header-footer/src/makeStandalone.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { render } from 'react-dom';
 import StandaloneProvider from './StandaloneProvider/StandaloneProvider';
 
-export default Component => (el, initialProps) => {
+export default (Component, defaultProps) => (el, initialProps) => {
```
markdalgleish: Actually, do we need to default this? `(Component, defaultProps = {})`?

PR #320, opened 2017-09-13
fix(standalone): Spread existing componentProps with passed in componentProps then set as state

dborysiewicz:
## Commit Message For Review

Persist previously set props when using updateProps with standalone components.

REASON FOR CHANGE:

If you update props any default ones (ie: `{ logoComponent: LogoRainbow }`) was lost.

### DIFF
--- a/standalone/header-footer/src/StandaloneProvider/StandaloneProvider.js
+++ b/standalone/header-footer/src/StandaloneProvider/StandaloneProvider.js
@@ -20,7 +20,9 @@ export default class StandaloneProvider extends Component {
   }
 
   updateProps(componentProps) {
-    this.setState({ componentProps });
+    this.state.componentProps = { ...this.state.componentProps, ...componentProps };
+
+    this.setState(this.state);
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


standalone/header-footer/src/StandaloneProvider/StandaloneProvider.js
```diff
@@ -20,7 +20,9 @@ export default class StandaloneProvider extends Component {
   }
 
   updateProps(componentProps) {
-    this.setState({ componentProps });
+    this.state.componentProps = { ...this.state.componentProps, ...componentProps };
```
markdalgleish: We shouldn't be editing state directly like this. Why is the call to set state not working correctly?

standalone/header-footer/src/StandaloneProvider/StandaloneProvider.js
```diff
@@ -20,7 +20,9 @@ export default class StandaloneProvider extends Component {
...
+
+    this.setState(this.state);
```
markdalgleish: Instead, this should probably be solved like this:

```
this.setState({ componentProps: { ...this.state.componentProps, ...componentProps }});
```

dborysiewicz: Happy to switch over if thats the preferred way, change incoming 🚀

Issue #321, opened 2017-09-14
RFC: Add '<ButtonGroup>' component, introduce '<CardGroup>' to replace '<Card group>'


dborysiewicz:
# Summary

There is a need for consistent (& responsive) buttons that are being used with forms.

# Motivation

The initial motivation was to make a form submit button more predominant on mobile by going full-width. However, after discussing this, we've decided it might be a good idea to make this a more general pattern for groups of `Button` components in the context of forms.

Example of implementation:
<img width="1678" alt="screen shot 2017-09-14 at 10 16 25 am" src="https://user-images.githubusercontent.com/7741507/30406590-0f506f3c-9936-11e7-937b-1f9826cced3e.png">

# Detailed design

This will be implemented as a separate `ButtonGroup` component

```js
<ButtonGroup>
  <Button {...} />
  <Button {...} />
</ButtonGroup>
```

`<Button group>` will be a div wrapper that will have `display: flex;` set but also set extra/override css on all nested buttons. This will also include targeting certain button colors with different css overrides.

We currently have a `<Card group>` component, but we're proposing that we also migrate this to a separate `<CardGroup>` component, even if it's just a surface level change for now, rendering a `<Card group>` internally.

# How We Teach This

This should be fairly straightforward since it follows established patterns, i.e. `<Card group>`, soon the be renamed as `<CardGroup>`.

Since it's a new component, we'll add a new `ButtonGroup` page to the demo site.

# Drawbacks

It's not a silver bullet for all implementations of button but should work for most cases with forms.
As usual, it does adds extra complication to the API.

# Alternatives

We could provide a prop (`flexible`?) directly to button without wrapping them in a group, but we'd still need to nest them within a container with `display: flex`. To better support this, we've talked about making a generic `Flex` component, but felt like this would be branching off into a very different pattern to what we current have.

mengtzu:
Sounds reasonable, go ahead.

markdalgleish:
👍

markdalgleish:
Just updated this issue based on the new `<ButtonGroup>` API (instead of `<Button group>`) and noted that we're also aiming to change the `<Card group>` API to `<CardGroup>`.

markdalgleish:
Done 👍

PR #322, opened 2017-09-14
feat(ButtonGroup): Add button group

dborysiewicz:
## Commit Message For Review

Add a new `ButtonGroup` component to support a consistent & responsive `Buttons` when they are being used with forms.

RFC URL:

https://github.com/seek-oss/seek-style-guide/issues/321

REASON FOR CHANGE:

To create consistent & responsive buttons when they are being used with forms.

EXAMPLE USAGE:

```js
<ButtonGroup>
  <Button {...} />
  <Button {...} />
</ButtonGroup>
```


react/Button/Button.js
```diff
@@ -10,9 +10,13 @@ export default class Button extends Component {
   static displayName = 'Button';
 
   static propTypes = {
-    color: PropTypes.oneOf([
-      'pink', 'blue', 'gray', 'transparent'
-    ]).isRequired,
+    /* eslint-disable consistent-return */
+    color: (props, propName, componentName) => {
+      if (!props.group && ['pink', 'blue', 'gray', 'transparent'].indexOf(props.color) === -1) {
```
tamoore: `!props.group &&`

Maybe this is an rfc concern and I have no issue with the above implementation just to be clear, but I could imagine a world where the following is quite useful:

```js
<Button group color="pink">
  <Button />
  <Button />
  <Button />
</Button>
```

react/Button/ButtonGroup.demo.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
```
tamoore: This seemed a bit odd to me, seeming as `group` is just a prop on Button it felt like it should be a component of the Button demo. Can appreciate though that this might be cleaner to do.

react/Button/Button.js
```diff
@@ -10,9 +10,13 @@ export default class Button extends Component {
...
+        return new Error(`When ${componentName} doesn't not have \`group\`, it should have an \`color\`.`);
```
markdalgleish: The fact that we need to write code like this is making me reconsider this pattern 🤔

react/Button/Button.js
```diff
@@ -46,7 +51,11 @@ export default class Button extends Component {
   }
 
   render() {
-    const { color, className, loading, fullWidth, children, component, ...restProps } = this.props;
+    const { color, className, loading, fullWidth, children, component, group, ...restProps } = this.props;
+
+    if (group) {
+      return <div className={styles.group}>{children}</div>;
```
georgespyropoulos: This line is making me wonder whether it might be better to have a seperate `<ButtonGroup>` component instead?

dborysiewicz:
After a discussion with @markdalgleish & @georgespyropoulos it was decided that shifting towards a standalone component over a prop was the best way forward hence the latest commit has introduced `ButtonGroup`. In following along the same lines I have also created `CardGroup` for the `Card` component however this is just a implementation of `<Card group/>` and is a non breaking change. I will/have updated the original description & title to reflect this work.

react/Button/Button.less
```diff
@@ -111,3 +111,47 @@
 .fullWidth {
   width: 100%;
 }
+
```
mengtzu: Does it make sense to have the group style in the same less file?  Seems like it's big enough that a file of its own in the `ButtonGroup` directory would be appropriate

markdalgleish: @mengtzu The issue is that we need to reference classes from both components, since there's a parent/child relationship.

react/Button/Button.less
```diff
@@ -111,3 +111,47 @@
...
+    & + .root_isTransparent {
+      margin-left: 0;
+    }
+  }
+
+  .root_isTransparent {
+    padding-left: @grid-gutter-width;
+    padding-right: @grid-gutter-width;
+  }
+  @media only screen and (max-width: 960px) {
```
kosanna: Why we are not using our standard breakpoints here 🤔

react/ButtonGroup/ButtonGroup.test.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import ButtonGroup from './ButtonGroup';
+import Button from '../Button/Button';
+
+describe('ButtonGroup:', () => {
+  it('should render a button group', () => {
+    const wrapper = shallow(<ButtonGroup />);
```
kosanna: This test should throw a warning as children is a required prop

kosanna: This test should throw a warning since children is a required prop, I'd rather test with 1it) with node  2it) with classname

react/Button/Button.less
```diff
@@ -111,3 +111,47 @@
...
+    .root {
+      flex: 33.33% 0;
+    }
+
+    .root_isTransparent {
+      flex: 20% 0;
+    }
+  }
+  @media only screen and (max-width: 500px) {
```
kosanna: seems very specific to me, because what if I want to implement them in a narrow container for example. will be weird

react/Button/Button.less
```diff
@@ -111,3 +111,47 @@
...
+    }
+  }
+
+  .root_isTransparent {
+    padding-left: @grid-gutter-width;
+    padding-right: @grid-gutter-width;
+  }
+  @media only screen and (max-width: 960px) {
+    .root {
+      flex: 33.33% 0;
```
kosanna: I'd think perhaps it would make sense to pass from outside in each individual case

react/Button/Button.less
dborysiewicz: It's a hard one - I know that the css rules look all bit random but it's sorta just works best for the different sizes. This rule I can switch to our mixin but example the 500px below was picked as breaking this earlier the buttons looked to big, I have demo'ed this all with @stephaniemoss and she is happy with how it looks if it's any consideration?

react/Button/Button.less
michaeltaranto: Interested to know why the 33.3%? Wouldnt this be better expressed as a `max-width`?

react/Button/Button.less
```diff
@@ -111,3 +111,47 @@
 .fullWidth {
   width: 100%;
 }
+
+.group {
+  display: flex;
+  width: 100%;
+
+  .root {
+    overflow: visible;
```
michaeltaranto: This might created more unexpected results, a button truncates overflow, this is explicitly negating that which could be an undesired side-effect of putting wrapping a `<Button />` in a `<ButtonGroup />`.  You can see below that with this property a button's label can awkwardly extend into the padding:
screen shot 2017-09-19 at 10 12 39 am

PR #324, opened 2017-09-18
feat(CardGroup): Add CardGroup

dborysiewicz:
## Commit Message For Review

Added `CardGroup` component to support a more future proofed API for grouping components of the same type. `CardGroup` is just a implementation of `<Card group />` so at the time of writing this is a non breaking change.

RFC URL:

https://github.com/seek-oss/seek-style-guide/issues/321

REASON FOR CHANGE:

While working on `ButtonGroup` it became evident the keeping the same interface, that being `<Card group/>` & `<Button group />` didn't make sense for the component `Button`.

MIGRATION GUIDE:

Before:

```js
<Card group>
  <Card {...} />
  <Card {...} />
</Card>
```

After:

```js
<CardGroup>
  <Card {...} />
  <Card {...} />
</CardGroup>
```

### DIFF
--- a/react/CardGroup/CardGroup.js
+++ b/react/CardGroup/CardGroup.js
@@ -2,16 +2,14 @@ import Card from '../Card/Card';
 import React from 'react';
 import PropTypes from 'prop-types';
 
-export default function CardGroup({ className, children, transparent, ...restProps }) {
+export default function CardGroup({ children, ...restProps }) {
   return (
-    <Card {...restProps} group className={className} transparent={transparent}>
+    <Card {...restProps} group>
       {children}
     </Card>
   );
 }
 
 CardGroup.propTypes = {
-  className: PropTypes.string,
-  children: PropTypes.node.isRequired,
-  transparent: PropTypes.bool
+  children: PropTypes.node.isRequired
 };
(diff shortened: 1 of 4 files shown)


markdalgleish:
Might be worth calling out explicitly that we're maintaining backwards compatibility, and adding an upgrade guide for those who want to fully upgrade now.

react/CardGroup/CardGroup.js
```diff
@@ -0,0 +1,17 @@
+import Card from '../Card/Card';
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function CardGroup({ className, children, transparent, ...restProps }) {
+  return (
+    <Card {...restProps} group className={className} transparent={transparent}>
```
markdalgleish: Why do we need to manually pass through `className` and `transparent`?

react/CardGroup/CardGroup.js
```diff
@@ -0,0 +1,15 @@
+import Card from '../Card/Card';
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function CardGroup({ children, ...restProps }) {
+  return (
+    <Card {...restProps} group>
+      {children}
```
michaeltaranto: You don't really need to access the `children` here, i would refine this to `<Card group {...props} />` which simplifies the code here, but importantly leaves the prop type validation of `children` in the `<Card>` component itself.

PR #333, opened 2017-09-27
fix(React): Add support for React v16

markdalgleish:
## Commit Message For Review

REASON FOR CHANGE:

React v15 deprecated React.PropTypes in favour of the 'prop-types' package, but React v16 removed it entirely. To ensure consumers can use the style guide with React v16, we've removed all references to React.PropTypes by cleaning up a few leftover references and updating our dependencies. This also required upgrading our test infrastructure, which included a major upgrade to Enzyme.


react/TextField/TextField.test.js
```diff
@@ -9,14 +9,11 @@ import {
   scryRenderedDOMComponentsWithClass
 } from 'react-dom/test-utils';
 import { createRenderer } from 'react-test-renderer/shallow';
-import SyntheticEvent from 'react-dom/lib/SyntheticEvent';
```
markdalgleish: `react-dom/lib/SyntheticEvent` is no longer available, but it seemed like the assertion it was being used for was arguably overkill anyway.

test/setupPolyfills.js
```diff
@@ -0,0 +1 @@
+require('raf').polyfill();
```
markdalgleish: React v16 now makes use of `requestAnimationFrame`, so we need to polyfill it in node environments.

react/Autosuggest/Autosuggest.test.js
```diff
@@ -50,8 +50,7 @@ describe('Autosuggest', () => {
   it('should render with mobile backdrop', () => {
     const props = getAutosuggestProps();
     props.labelProps = {
-      className: 'LABEL_TEST_CLASS',
-      someLabelProp: 'bar'
```
markdalgleish: This prop is pretty useless from a testing standpoint, and React v16 was actually rendering this to the underlying label when the previous version didn't. Rather than adding this pointless prop to the snapshot, I figured it'd make more sense to just remove it entirely.

react/Autosuggest/__snapshots__/Autosuggest.test.js.snap
```diff
@@ -36,7 +36,9 @@ exports[`Autosuggest should render with label 1`] = `
         role="combobox"
         type="text"
         value=""
-      />
+      >
+        
+      </input>
```
pokle: Is this just the way `hydrate()` renders elements?

test/setupPolyfills.js
pokle: Is this a concern that should be left for apps like Chalice?

react/Autosuggest/__snapshots__/Autosuggest.test.js.snap
markdalgleish: No, this is a change to the renderer in Enzyme.

docs/src/components/Demo/Baseline/Baseline.js
```diff
@@ -0,0 +1,64 @@
...
+  lineHeight: PropTypes.number,
+  color: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  style: PropTypes.object
+};
+
+Baseline.defaultProps = {
+  isVisible: true,
+  type: 'line',
+  lineHeight: 9,
```
michaeltaranto: you can require in the grid row height here if you like:
```
import lessToJs from 'less-vars-to-js';
import grid from '!!raw!seek-style-guide/theme/layout/grid.less';
const gridValues = lessToJs(grid);
const rowHeight = parseInt(gridValues['@grid-row-height'], 10);
```

That will maybe better facilitate the migration to the new grid too.

package.json
```diff
@@ -111,17 +112,17 @@
     "opn": "^4.0.1",
     "opn-cli": "^3.1.0",
     "postcss-prefix-selector": "^1.6.0",
-    "react": "^15.3.1",
+    "raf": "^3.3.2",
```
michaeltaranto: What was the reason for bringing in `raf` in this PR?

markdalgleish: Covered this here: https://github.com/seek-oss/seek-style-guide/pull/333#discussion_r141238096

michaeltaranto: Just saw it's usage further down.... as you were

docs/src/components/Demo/Baseline/Baseline.js
markdalgleish: Yeah, nice idea 👍

markdalgleish: Done 👍

PR #334, opened 2017-09-28
docs(site): Configure router basename

markdalgleish:
## Commit Message For Review

Upgrading React Router seems to have broken the docs on GitHub Pages, even though they work on Surge. It seems likely to be caused by deprecation of transparent base href tag support, but regardless of whether this is the actual cause of the issue or not, we now need to explicitly provide the basename to our history instance.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "gh-pages": "^0.11.0",
     "glob": "^7.1.1",
     "hex-rgb": "^1.0.0",
+    "history": "^3.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -102,6 +102,7 @@
     "gh-pages": "^0.11.0",
     "glob": "^7.1.1",
     "hex-rgb": "^1.0.0",
+    "history": "^3.3.0",
```
markdalgleish: FYI I made sure to install the same version that React Router v3 depends on.

PR #339, opened 2017-10-05
chore(github): Streamline RFC and PR templates

markdalgleish:
As discussed, trying to make the process of proposing changes and issuing pull requests a bit easier for consumers, without sacrificing the benefits of the existing approach.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,18 +1,6 @@
-ARE YOU PLANNING TO CHANGE THE STYLE GUIDE'S PUBLIC API? If so, please make sure you've first gone through the RFC approval process by opening a new issue and using the provided template.
+** Please provide as much detail as possible, but feel free to remove irrelevant sections **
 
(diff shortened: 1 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,30 +1,14 @@
...
-
-
-## Commit Message For Review
+{{ Reason for change }}
 
 BREAKING CHANGE:
 
 {{ Description of breaking API change }}
 
 RFC URL:
```
michaeltaranto: Worth adding conditions for sections like these? eg. `RFC URL (for api changes only):` and `BREAKING CHANGE (if any):

markdalgleish: Yeah, it's tricky. I'm trying to minimise the amount of careful editing you need to perform on this.

markdalgleish: That's why I added the first line to this file about removing irrelevant sections.

PR #341, opened 2017-10-06
feat(webpack): Upgrade to webpack 3, remove webpack 1 decorators

markdalgleish:
Internally, this project has been stuck on webpack 1 for quite a while, even though almost all consumers have migrated to later versions of webpack, purely so that we can maintain support for the old webpack 1 config decorators exported by this package. Luckily, usage of them has dropped so low that it's about time we finally remove them.

The upside, of course, is that we now get to ensure that the style guide works with the latest versions of everything. This also greatly reduces the number of dependencies, since the webpack logic is no longer exposed to clients. As a result, all webpack-related packages have moved from dependencies to dev dependencies.

BREAKING CHANGE:

Webpack config decorators have been removed.

MIGRATION GUIDE:

If you're still using the webpack 1 decorators, you'll first need to migrate to the latest version of webpack. Then, install the separate `seek-style-guide-webpack` package:

```
$ npm install --save-dev seek-style-guide-webpack
```

Finally, update package references in your webpack config:

Before:

```js
require('seek-style-guide/webpack')
```

After:

```js
require('seek-style-guide-webpack')
```


webpack/index.js
```diff
@@ -1,260 +0,0 @@
...
-        ].join('!'))
-      }
-    ]
-  });
-};
-
-module.exports = {
-  decorateServerConfig,
-  decorateClientConfig
-};
```
markdalgleish: 👋😘

standalone/header-footer/webpack.config.js
```diff
@@ -1,9 +1,12 @@
+// Alias 'seek-style-guide' so 'seek-style-guide-webpack' works correctly
```
mengtzu: I wish this wasn't necessary.  Is there a way around it by modifying `seek-style-guide-webpack`?

markdalgleish: Possibly, but I think this is actually easier to reason about anyway, even if it's a little hacky.

package.json
```diff
@@ -126,20 +120,24 @@
...
     "surge": "^0.18.0",
+    "svgo": "0.7.2",
+    "svgo-loader": "1.2.1",
     "validate-commit-msg": "^2.8.2",
-    "webpack": "^1.13.1",
-    "webpack-dev-server": "^1.14.1",
-    "webpack-fail-plugin": "^1.0.6",
+    "webpack": "^3.6.0",
+    "webpack-dev-server": "^2.9.1",
+    "webpack-fail-plugin": "^2.0.0",
```
michaeltaranto: The fail plugin I believe was only required in webpack 1, where it didn't have the correct exit code if there was an error.

markdalgleish: Nice catch 👍

markdalgleish:
Agreed that it could use a clean up too, but I'd probably want to keep that as a separate commit.

docs/webpack.dev.config.js
```diff
@@ -25,43 +28,70 @@ const config = decorateClientConfig({
...
+            loader: 'css-loader',
+            options: {
+              modules: true,
+              localIdentName: '[name]__[local]___[hash:base64:5]'
+            }
+          },
+          {
+            loader: 'postcss-loader',
+            options: {
+              plugins: [autoprefixer]
```
michaeltaranto: Is there a reason for not passing the [`autoprefixer.config.js`](config/autoprefixer.config.js) as you do [here](https://github.com/seek-oss/seek-style-guide/pull/341/files#diff-242962979875c719b0652867bfbe48f2R66)

markdalgleish: This was a straight port, not sure what the original reasoning was, if there was any. Makes sense to pass it in here too 👍

docs/webpack.static.render.config.js
```diff
@@ -27,34 +29,56 @@ const config = {
...
+            loader: 'css-loader/locals',
+            options: {
+              modules: true,
+              localIdentName: '[name]__[local]___[hash:base64:5]'
+            }
+          },
+          {
+            loader: 'postcss-loader',
+            options: {
+              plugins: [autoprefixer]
```
michaeltaranto: As mentioned [here](https://github.com/seek-oss/seek-style-guide/pull/341/files#r143368472)

standalone/header-footer/webpack.config.js
```diff
@@ -25,39 +28,66 @@ const config = {
...
+              loader: 'css-loader',
+              options: {
+                modules: true,
+                localIdentName: '[name]__[local]___[hash:base64:5]'
+              }
+            },
+            {
+              loader: 'postcss-loader',
+              options: {
+                plugins: [autoprefixer]
```
michaeltaranto: As mentioned [here](https://github.com/seek-oss/seek-style-guide/pull/341/files#r143368472)

docs/webpack.dev.config.js
michaeltaranto: Also further to this we are now [exporting the autoprefixer config from the seek-style-guide-webpack module](https://github.com/seek-oss/seek-style-guide-webpack/blob/master/autoprefixer.config.js), so might be a good clean up to move onto that one given they are identical?

michaeltaranto: As discussed offline, let's come back to this clean up, as the autoprefixer config should really be its own module.

docs/webpack.static.render.config.js
markdalgleish: This one isn't necessary because we're not actually generating a style sheet in the static render pass—it's just to generate the localised class names.

PR #342, opened 2017-10-06
docs(site): Upgrade to React Router v4

markdalgleish:
Long overdue. Turned out to be much easier than I thought 🎉


markdalgleish:
In case you're wondering, this was only easy in this case because we aren't doing any data fetching before rendering server side. If we were using something like [redial](https://github.com/markdalgleish/redial), this would have been quite a bit trickier.

PR #343, opened 2017-10-09
 chore(deps): Upgrade dev dependencies

markdalgleish:
As part of this, I also migrated from babel-preset-es2015 to babel-preset-env.

This upgrade has also forced the following changes:
- Migrate several lint rule definitions
- Fix newly surfaced lint warnings
- Update snapshots to use self-closing tags (this was broken in the last enzyme upgrade)
- Migrate calls to sinon's stub API
- Upgrade to node v8 (via nvmrc) to support semantic-release
- Move Jest's config to the root directory due to (what seems like) a bug in how it resolves the `<rootDir>` string when your config file isn't actually in the root directory 😞


markdalgleish:
@mattsjones agreed—keen to address that in a subsequent clean up 👍

PR #346, opened 2017-10-10
feat(palette): Add sk-teal to brand palette

kosanna:
NAMING TBD, please let me know which one you prefer. 
http://chir.ag/projects/name-that-color/#00A5AD

This PR adds an existing brand colour to theme variables.
This colour will be used as a background colour for the app install banners and also on a button on the home page banner for app stores link button.

screen shot 2017-10-10 at 11 34 22 am


markdalgleish:
Just had a chat with some of our designers—can we rename this to `@sk-teal`?

kosanna:
Renamed

markdalgleish:
Sorry, could you also add this to the readme? This is the only place we document these colours, for now at least.

PR #348, opened 2017-10-11
fix(Columns): Remove implied min-width for Columns

jahredhope:
## Reason for change
At the moment the columns allow flexing based on content.
E.g Column 1 with wide content might be at 60% while Column 2 with thin content might be at 40%.
This may not be ideal default behavior.

From [MDN min-width](https://developer.mozilla.org/en-US/docs/Web/CSS/min-width)
> auto
> The default minimum width for flex items, providing a more reasonable default than 0 for other layouts.

## Change
This change will ensure column widths default to 1fr and do not change based on content.

#### Before change
Note: In both cases 'overflow-wrap: break-word;' is applied to the text
screen shot 2017-10-11 at 2 16 58 pm

#### With change
screen shot 2017-10-11 at 2 16 38 pm

## Migration Guide
The default behaviour of columns will now be remain equal widths even when content would attempt to push one column wider.
To enable the previous behaviour use the flexible prop.
``` javascript
<Columns flexible={true}>
 ...
</Columns>
```

## Commit comment for review
fix(Columns): Remove implied min-width for Columns

Columns component will no-longer flex based on content width

BREAKING CHANGE: Columns no-longer flexible by default. Use `flexible` prop to enable previous behaviour.

### DIFF
--- a/react/Columns/Columns.js
+++ b/react/Columns/Columns.js
@@ -8,12 +8,13 @@ const renderColumn = (el, index) => (
   <div key={index} className={styles.column}>{el}</div>
 );
 
-export default function Columns({ children, tight }) {
+export default function Columns({ children, tight, flexible }) {
   return (
     <div
       className={classnames({
         [styles.columns]: true,
-        [styles.columns_tight]: tight
+        [styles.columns_tight]: tight,
+        [styles.columns_flexible]: flexible
       })}>
       {children.map(renderColumn)}
     </div>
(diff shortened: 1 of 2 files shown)


markdalgleish:
I'd say this is technically a breaking change, even if it's desirable in most (if not all) cases. Just in case, it might be a good idea to add a `flexible` prop that enables the old behaviour, which we could add to the migration guide. If it turns out nobody uses this prop, we could then deprecate it.

jahredhope:
Will re-implement with the ability to enable previous behaviour.

PR #350, opened 2017-10-12
feat(Columns): Make Columns equal width by default

jahredhope:
## Reason for change
At the moment the columns allow flexing based on content.
E.g Column 1 with wide content might be at 60% while Column 2 with thin content might be at 40%.
This may not be ideal default behavior.

From [MDN min-width](https://developer.mozilla.org/en-US/docs/Web/CSS/min-width)
> auto
> The default minimum width for flex items, providing a more reasonable default than 0 for other layouts.

## Change
This change will ensure column widths default to 1fr and do not change based on content.

#### Before change
Note: In both cases 'overflow-wrap: break-word;' is applied to the text
screen shot 2017-10-11 at 2 16 58 pm

#### With change
screen shot 2017-10-11 at 2 16 38 pm

## Commit message for review
feat(Columns): Make Columns equal width by default

BREAKING CHANGE: Columns no-longer flexible by default.

MIGRATION GUIDE:

If you'd like to preserve the previous behaviour, use the new `flexible` prop.

Before:

```js
<Columns>
  ...
</Columns>
```

After:

```js
<Columns flexible>
  ...
</Columns>
```

### DIFF
--- a/react/Columns/Columns.js
+++ b/react/Columns/Columns.js
@@ -8,12 +8,13 @@ const renderColumn = (el, index) => (
   <div key={index} className={styles.column}>{el}</div>
 );
 
-export default function Columns({ children, tight }) {
+export default function Columns({ children, tight, flexible }) {
   return (
     <div
       className={classnames({
         [styles.columns]: true,
-        [styles.columns_tight]: tight
+        [styles.columns_tight]: tight,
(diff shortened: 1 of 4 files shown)


markdalgleish:
Just updated the description to move the migration guide into the release notes.

PR #357, opened 2017-10-20
docs(site): Remove hard-coded component list, order menu alphabetically

markdalgleish:


markdalgleish:
Pretty self explanatory, but this is so that we don't have to manually maintain a centralised component list, but a nice side effect is that it enforces alphabetical order.

In the future, if we want to group components into sections, I'm thinking we can do it with a `category` flag in the demo spec, or something similar.

PR #367, opened 2017-11-01
chore(package): Update seek-style-guide-webpack to v2.1.4

markdalgleish:
This is to fix issues with SVGO and svgo-loader version clashes and trigger a rebuild of the site.

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "react-test-renderer": "^16.0.0",
     "rewrite-files": "^1.0.1",
     "rimraf": "^2.6.2",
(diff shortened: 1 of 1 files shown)


markdalgleish:
You can see the documentation preview here, with fixed SVGs visible on the home page: http://seek-style-guide--50d9cd9c541e8657db43f8f5f43813f8dcee1a99.surge.sh

markdalgleish:
Sorry, realised this isn't actually necessary, just re-ran the last master build since it wasn't a release commit.

PR #368, opened 2017-11-08
feat(MonthPicker): Make year options configurable

mattsjones:
Add min & max year options to MonthPicker. Also, add sort order options (ascending/descending).

Currently, the `MonthPicker` has a static list of year options which is the current year going back 100 years. This PR makes the years configurable including which direction they are sorted. There are no breaking changes as the default props invoke the old behaviour. 

**Warning**: This change only affects the `CustomMonthPicker` which utilizes browser dropdowns. The `NativeMonthPicker` behaviour remains unchanged. Do not rely on these props for validation purposes as they do not ensure the data returned from the component in native mode. 

EXAMPLE USAGE:

```js
<MonthPicker
  id="startMonth"
  label="Start Month"
  maxYear={2048}
  minYear={2000}
  ascendingYears={true} // default is false
/>
```

### DIFF
--- a/react/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
+++ b/react/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
@@ -4,23 +4,9 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
 import classnames from 'classnames';
-import range from 'lodash/range';
 
+import getYearOptions from './getYearOptions';
 import Dropdown from '../../Dropdown/Dropdown';
-
-const getYearOptions = () => {
-  const maxYear = new Date().getFullYear();
-  const minYear = maxYear - 100;
-
-  return range(maxYear, minYear - 1).map(value => {
-    const stringValue = String(value);
-
-    return {
-      value: stringValue,
-      label: stringValue
-    };
-  });
-};
 
 const months = [
   { value: '1', label: 'Jan' },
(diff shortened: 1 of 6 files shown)


react/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -4,24 +4,10 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
 import classnames from 'classnames';
-import range from 'lodash/range';
 
+import getYearOptions from './getYearOptions';
 import Dropdown from '../../Dropdown/Dropdown';
 
-const getYearOptions = () => {
-  const maxYear = new Date().getFullYear();
```
michaeltaranto: I cant see where this behaviour is being retained. Should `maxYear` not be a default prop of this year?

mattsjones: Max year defaults to the current year. Are you saying it shouldn't?

michaeltaranto: I agree it should, but since removing this line I cant see where the defaulting is done

mattsjones: Check out the public `MonthPicker` component. https://github.com/seek-oss/seek-style-guide/pull/368/files#diff-36c8b55b5646996302a1b706cd547849R49

michaeltaranto: Ignore me, my PR view was missing files for some reason 🤔

PR #370, opened 2017-11-14
feat(index): Created module defintiion for StyleGuide

tamoore:
This PR creates a flow-type lib defintion. See the [lib defintion section](https://flow.org/en/docs/libdefs/) of the flow-type docs.

The main goal of this PR is to create a module defintiion of the style guide that will allow a consumer to add the libdef to their project and get back type checks on the various components that the style guide provides.

So for example, once you include this libdef, you, as a consumer, will get errors if you use a value that is not defined for the Button color enum (pink, blue etc).

This will have no impact on the authoring of StyleGuide components.


tamoore:
Closing this in lieu of a POC.

PR #372, opened 2017-11-16
feat(Header): Add menu toggle callback

tamoore:
I need to know what the state of the menu is in the header. So I have added an option callback
function that could be called to allow us to know what it's current state is.

### DIFF
--- a/react/Header/Header.js
+++ b/react/Header/Header.js
@@ -27,7 +27,8 @@ export default function Header({
   linkRenderer,
   activeTab,
   divider,
-  returnUrl
+  returnUrl,
+  onMenuToggle = () => {}
 }) {
   const isAuthenticated = (authenticationStatus === AUTHENTICATED && (userName || userEmail));
   const isUnauthenticated = (authenticationStatus === UNAUTHENTICATED);
@@ -68,6 +69,7 @@ export default function Header({
                   linkRenderer={linkRenderer}
                   returnUrl={returnUrl}
                   activeTab={activeTab}
+                  onMenuToggle={onMenuToggle}
                 />
               </div>
               <div className={styles.signInRegisterWrapper}>
@@ -132,7 +134,8 @@ Header.propTypes = {
     'Advice & Tips'
   ]),
   divider: PropTypes.bool,
-  returnUrl: PropTypes.string
+  returnUrl: PropTypes.string,
+  onMenuToggle: PropTypes.func
 };
 
 Header.defaultProps = {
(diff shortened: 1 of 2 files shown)


react/Header/Header.js
```diff
@@ -27,7 +27,8 @@ export default function Header({
   linkRenderer,
   activeTab,
   divider,
-  returnUrl
+  returnUrl,
+  menuToggleCallBackFunction = () => {}
```
markdalgleish: Not a fan of the name—could we rename it to something like `onMenuToggle`?

react/Header/UserAccount/UserAccount.js
```diff
@@ -46,7 +51,10 @@ export default class UserAccount extends Component {
   }
 
   handleMenuToggleClick = () => {
-    this.setState(state => ({ menuOpen: !state.menuOpen }));
+    this.setState(state => {
+      this.props.menuToggleCallBackFunction(!state.menuOpen);
```
markdalgleish: Just to future proof the API a bit, it might be a good idea to pass an object to the callback so that we could add more stuff later without it being a breaking change. My vote is for `{ open: !state.menuOpen }`, since it seems close to what a native DOM API might look like.

tamoore:
Great suggestions, have amended it to capture both of those.

react/Header/UserAccount/UserAccount.js
```diff
@@ -46,7 +51,13 @@ export default class UserAccount extends Component {
   }
 
   handleMenuToggleClick = () => {
-    this.setState(state => ({ menuOpen: !state.menuOpen }));
+    this.setState(state => {
+      const menuState = {
+        menuOpen: !state.menuOpen
```
markdalgleish: Given the `onMenuToggle` event is explicitly scoped to the menu, I feel like it should be `open` instead of `menuOpen`. We probably need to have different objects for the event and the local component state.

tamoore: It makes some sense to keep these consistent, what do you think. So I should perhaps alter the internal implementation to use `open` too (assuming the context is obvious).

markdalgleish: To be honest, I feel like this is a risky refactor because it's creating a relationship where there shouldn't be one. What we choose to expose as an event object and the object we pass to `setState` are not logically the same thing, even if they happen to be the same right now.

markdalgleish: As an example of why this isn't logically correct, the event handler now has the ability to mutate the object being sent to `setState`.

tamoore: These are really good points and I hadn't considered them, folllowing a chat with Mark I agree that decoupling the state of the UserAccount from argument of the event handler is a must here. Have amended accordingly.

markdalgleish:
Re ☝️: if we ever wanted to allow external control of the menu, we'd make it a prop rather than allowing access to the internal state.

PR #377, opened 2017-11-22
 feat(Button): Add ghost button styles

kosanna:
This PR: 
- adds a non-required `ghost` prop to the Button
- adds styles to ghost blue buttons
- adds style to ghost white button

Example:

blue
screen shot 2017-11-23 at 8 02 19 pmscreen shot 2017-11-23 at 8 02 35 pmscreen shot 2017-11-23 at 8 02 44 pm

transparrent white
screen shot 2017-11-23 at 8 04 49 pmscreen shot 2017-11-23 at 8 04 56 pmscreen shot 2017-11-23 at 8 05 04 pm

Example usage:
```
<Button color="white" ghost {...restProps} />
```

### DIFF
--- a/react/Button/Button.js
+++ b/react/Button/Button.js
@@ -4,7 +4,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
 import classnames from 'classnames';
-import { capitalize } from 'lodash';
+import capitalize from 'lodash/capitalize';
 
 export default class Button extends Component {
   static displayName = 'Button';
--- a/react/Button/Button.js
+++ b/react/Button/Button.js
@@ -11,7 +11,7 @@ export default class Button extends Component {
 
   static propTypes = {
     color: PropTypes.oneOf([
-      'pink', 'blue', 'gray', 'transparent'
+      'pink', 'blue', 'gray', 'transparent', 'white'
     ]).isRequired,
     children: PropTypes.oneOfType([
       PropTypes.arrayOf(PropTypes.node),
--- a/react/Button/Button.js
+++ b/react/Button/Button.js
@@ -4,6 +4,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
 import classnames from 'classnames';
+import { capitalize } from 'lodash';
 
 export default class Button extends Component {
   static displayName = 'Button';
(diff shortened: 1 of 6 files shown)


react/Button/Button.js
```diff
@@ -4,6 +4,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
 import classnames from 'classnames';
+import { capitalize } from 'lodash';
```
markdalgleish: Heads up, this line is pulling in the entirety of lodash. Should be importing from `lodash/capitalize`.

react/Button/Button.js
```diff
@@ -21,40 +22,34 @@ export default class Button extends Component {
...
   };
 
-  constructor() {
-    super();
-
-    this.storeButtonReference = this.storeButtonReference.bind(this);
-  }
-
-  storeButtonReference(button) {
+  storeButtonReference = button => {
```
markdalgleish: Nice clean up 👏

stephaniemoss:
This is looking great! Main comment right now is the the text isn't on grid/centred yet.

stephaniemoss:
Also I need to go through the colours, especially for the hover/click states. We've only previously done link blue and white. I'll get back to you soon.

kosanna:
@stephaniemoss text is exactly on the same position as it is for solid buttons atm, and we are going to use a white variation only for now, I just drafter blue and pink so it makes a little bit of sense in the demo, I'm not sure about usages for those.

stephaniemoss:
ghost button test specs

This is the ghost button specs that we most commonly use across Profile and Onboarding.

I'll work on if we need the other colours and what they would look like if we have them.

Thanks!

kosanna:
Perfect @stephaniemoss. Thanks heaps for doing that, I'll redo the styles

michaeltaranto:
Be interesting to work out of we need hover and active states for each colour variant or if we can fade the colour like we do for the standard button states. That way the values are dynamic based which variant you choose.

kosanna:
@michaeltaranto I fade everything, no worries 👍

react/Button/Button.less
```diff
@@ -75,10 +88,33 @@
...
+  border: 2px solid;
+  font-weight: @sk-bold;
+  line-height: 0;
+}
+
+.root_isBlue.root_isGhost {
+  .ghostColor(@sk-link)
+}
+
+.root_isWhite.root_isGhost {
```
michaeltaranto: Why does this class not use the `.ghostColor` mixin?

kosanna: Because it's not a `colored` ghost button, it's white and it's different

stephaniemoss:
I've advised Anna that we make the focus state hover tint + blue outline to make it more clear that it is focused on and to keep the bounce for now since we don't have a reason to make it different from other buttons at the moment :)

react/Button/Button.less
```diff
@@ -75,10 +97,45 @@
...
```
georgespyropoulos: The prop type validation here (https://github.com/seek-oss/seek-style-guide/pull/377/files#diff-bdfcd58f19e20b24ebd558c8a351c45fL13) needs to allow for white now

react/Button/Button.less
```diff
@@ -75,10 +97,45 @@
...
 .root_isTransparent {
-  color: @sk-link;
   background-color: transparent;
   box-shadow: none;
+}
+
+.root_isGhost {
+  border: 2px solid;
+  font-weight: @sk-bold;
+  line-height: 0;
```
georgespyropoulos: I'm wondering if setting line-height to `normal` might be a better option here? From reading through the code this value didn't feel right to me. I guess another option could be to set it to 45px - 4px of border?

react/Button/Button.less
kosanna: Great pickups @georgespyropoulos 👍 Thanks, fixed

PR #378, opened 2017-11-22
feat(standalone): Provide pre-rendered Header and Footer HTML snippets

markdalgleish:
Standalone header/footer consumers that want to improve initial rendering performance now have access to pre-rendered HTML snippets that can be manually integrated into their project, regardless of tech choice.

The existing client-only rendering option still exists, primarily to maintain backwards compatibility.

Preview versions of each snippet (which include the surrounding markup that the snippet typically requires in the context of an app) are rendered so that they can be manually opened in a browser and verified. These could serve as the basis for some standalone header/footer integration tests, but that felt like a lower priority right now.

This also fixes an existing issue with the standalone header and footer where they weren't including the provider styles that are typically included via `<StyleGuideProvider>`.


standalone/header-footer/src/StandaloneProvider/StandaloneProvider.js
```diff
@@ -31,7 +31,9 @@ export default class StandaloneProvider extends Component {
 
     return (
       <div className={cssSelectorPrefix.replace('.', '')}>
-        <StandaloneComponent {...componentProps} />
+        <div className={styles.root}>
```
michaeltaranto: Whats the reason for needing this second `div` element, it's not immediately obvious?

markdalgleish: Ah, nice catch, I forgot to call this out.

I actually found an issue that we didn't include the provider styles in the standalone components, which meant that everything was in Times New Roman, etc. The provider styles need to be attached to a div nested within the existing div so that we can scope them to the `cssSelectorPrefix`.

As part of this PR, I extracted the provider styles into a mixin so that it can be used across both Provider components.

I'll update the description with this detail.

react/Header/UserAccount/UserAccount.js
```diff
@@ -1,15 +1,12 @@
...
 import PropTypes from 'prop-types';
 import ScrollLock from 'react-scrolllock';
-
 import ChevronIcon from '../../ChevronIcon/ChevronIcon';
 import ScreenReaderOnly from '../../ScreenReaderOnly/ScreenReaderOnly';
 import UserAccountMenu from '../UserAccountMenu/UserAccountMenu';
 import { AUTHENTICATED, UNAUTHENTICATED, AUTH_PENDING } from '../../private/authStatusTypes';
-
 import smallDeviceOnly from '../../private/smallDeviceOnly';
+import styles from './UserAccount.less';
```
markdalgleish: Note: I had to move this to the bottom of the import list because of a minor CSS import order issue which caused the user account menu chevron to be (slightly) the wrong size. This style sheet attempts to override styles provided by `ChevronIcon.less`, so it was logically incorrect to include it first.

PR #382, opened 2017-11-28
feat(Button): Add ghost buttons

kosanna:
This PR:
- adds a non-required `ghost` prop to the Button
- adds ghost blue button styles
- adds ghost white button style
- if both transparent and ghost attributes applied, ghost gets ignored
- bug fix for button line height
- triggers unreleased changes from the previous PR for ghost buttons #377 

Example usage:
```
<Button color="white" ghost {...restProps} />
<Button color="blue" ghost {...restProps} />
```

### DIFF
--- a/react/Button/Button.less
+++ b/react/Button/Button.less
@@ -138,12 +138,15 @@
   color: @sk-link;
   padding-left: 0;
   padding-right: 0;
+
   .hoverState({
     text-decoration: underline;
   });
+
   .activeState({
     transform: none;
   });
+
   @media @desktop {
     .touchableText(@standard-type-scale);
   }
--- a/react/Button/Button.less
+++ b/react/Button/Button.less
@@ -104,11 +104,11 @@
 }
 
 .root_isGhost {
-  @border-thickness: 2px;
+  @border-width: 2px;
 
-  border: @border-thickness solid;
+  border: @border-width solid;
   font-weight: @sk-bold;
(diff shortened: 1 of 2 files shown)


markdalgleish:
Could you please add the release notes to this PR? :)

react/Button/Button.less
```diff
@@ -104,9 +104,11 @@
 }
 
 .root_isGhost {
-  border: 2px solid;
+  @border-thickness: 2px;
+
+  border: @border-thickness solid;
   font-weight: @sk-bold;
-  line-height: normal;
+  line-height: @touchableTextHeight - 2 * @border-thickness;
```
markdalgleish: For readability, could you add parentheses here? e.g. `touchableTextHeight - (@border-thickness * 2)`

Also, for consistency with CSS naming, could you also call this variable `@border-width`?

markdalgleish:
I feel like a transparent ghost button should also be bold, otherwise it will look out of place next to a regular ghost button. What do you think?

kosanna:
I think that transparrent button is kind of a link and we should not touch it's styles if ghost is supplied. There is no use case for that so far, and no design, so lets leave it for now

kosanna:
Otherwise it looks weird that just to get a bold link you supply ghost. Just style it externally with @sk-bold if needed

markdalgleish:
Fair enough 👍

PR #383, opened 2017-12-01
build(sketch): Generate Sketch library with html-sketchapp

markdalgleish:
It's happening

With this PR, on every build, JSON files are generated via [html-sketchapp](https://github.com/brainly/html-sketchapp) that can be trivially converted to Sketch files thanks to html-sketchapp's corresponding Sketch plugin.

This is achieved by extracting elements from a new, hidden route called `/sketch-exports` which renders everything exported from every `*.sketch.js` file in the `react` and `theme` folders.

To keep the project clean, I also removed the legacy Sketch files and related code since it was a bit of a failed experiment, and this clearly supersedes it.


markdalgleish:
🎉🎉🎉

PR #385, opened 2017-12-06
feat(Rating): Add sizes and optional text rating to stars

kosanna:
## Support for `superstandard`, `substandard`, `heading` sizes

Scales icons to respective typographys' sizes. For example if `heading` is supplied, icon size will be 21px.

Example:
```js
<Rating rating={4} heading />
```

## New prop `showTextRating`
If prop is supplied, rating text will be shown on the right side of the stars
screen shot 2017-12-06 at 4 00 13 pm.

In this case text respects typography sizes and icons are scaled down for x0.7

Example:
```js
<Rating rating={3.5} superstandard showTextRating />
```
## Support for typography sizes on Icon component

Respects all standard typography states.

Example:
```js
<Icon markup={HalfStarIcon} hero />
<StarIcon headline />
```

## Other changes
- colour of stars is changed to `@sk-orange`
- spacing is added in between icons
- half star icon svg is updated to match filled/empty stars
- migrate tests to snapshots

TODO (not in this PR):
- think on changing star icons shapes (consistent/not with eg. Save Job button)


markdalgleish:
The first thing that jumps out at me is that the text size API differs from `<Text>`, when I'd rather they use the same conventions. Could this be `<Rating heading>` too? Or, alternatively, do we need to talk about changing `<Text heading>` to `<Text size="heading">`?

markdalgleish:
Also, the icon size convention being introduced here is what I'd like to see rolled out to all icons. Would you be able to update the implementation so that the size prop is passed down to `<StarIcon>` and `<HalfStarIcon>`?

kosanna:
@markdalgleish I started from passing them the same as we do for Text but later went away from this decision because it's not limiting the options to just this 3, which is not ideal, because there are no styles for subheading or standard, so they can't be used, means passing those props won't have the effect that you would imagine it will

kosanna:
Yup, will do icon size tomorrow 👍 Good idea

kosanna:
@markdalgleish I'm more or less happy with it now, please have a look and let me know if you thought about something different (and also if I should start splitting it up, initially anyway needed it here to see how it works with rating changes).

markdalgleish:
I think, whichever way we go on this, the API for specifying text size should be the same across all relevant components to make things easier to remember. To keep the change smaller, do you think you could change this to the boolean style API like `Text`? I think it's okay if they're not all implemented at first.

kosanna:
I prefer non-boolean style like that, however I don't want to make this crazy braking change to Text. And also I don't feel that Rating necessarily needs to support it exactly the same way (we don't use it as inline text element). So I'm happy to change Icon, but probably leave Rating as-is. Plus it's a nice restriction (see my comment above) on what you can pass and what you can't, rather then when you pass something and nothing happens

kosanna:
Plus I feel that's really not elegant when size could be passed via different non-linked bollean-type props. And so you need to mimic same code that's in text typing all the classnames manually, which is...not really...
```
classnames({
        [styles.root]: true,
        [className]: className,
        [styles.small]: small,
        [styles.substandard]: substandard,
        [styles.superstandard]: superstandard,
        [styles.subheading]: subheading,
        [styles.headline]: headline,
        [styles.heading]: heading,
        [styles.hero]: hero,
        [styles.raw]: raw,
        [styles.baseline]: baseline
      })
```

kosanna:
Updated to match Text api ⬆️

kosanna:
Could I please get some 👁 👁  on it?

react/Rating/Rating.less
```diff
@@ -1,5 +1,35 @@
 @import (reference) "~seek-style-guide/theme";
 
+@elements-spacing: 0.3em;
```
michaeltaranto: 👌

theme/mixins/iconSize.less
```diff
@@ -0,0 +1,12 @@
...
+  @icon-size: @typography-scale * @icon-size-baze;
+  @icon-size-mobile: @typography-scale-mobile * @icon-size-baze;
+
+  width: @icon-size;
+  height: @icon-size;
+  @media @mobile {
+    width: @icon-size-mobile;
+    height: @icon-size-mobile;
+  }
+}
```
michaeltaranto: 😍

theme/mixins/iconSize.less
```diff
@@ -1,4 +1,5 @@
-.iconSize(@typography-scale, @icon-scale: 1, @typography-scale-mobile: @typography-scale) {
+.iconSize(@typography-scale, @typography-scale-mobile: @typography-scale) {
+  @icon-scale: 0.7;
```
markdalgleish: I'm not sure we want to scale *all* icons to 0.7, it seems like this would be highly dependent on the icon itself.

kosanna: Depends on what you're trying to achieve, I can leave scale variable in the mixin, so we could use it.
image

markdalgleish: Yeah, I think it makes sense to leave this one as an argument so you can tweak each icon separately.

kosanna:
Closing for now as more work to be done

PR #391, opened 2017-12-17
feat(Icons): Introduce icons typography, adjust Rating component

kosanna:
## Icon sizes
Icons now could be user the same way as Text component, eg. 

```js
<StarIcon heading />
```

So if used inline it will look like that:
screen shot 2017-12-18 at 9 56 40 am

screen shot 2017-12-18 at 9 56 56 am

### Supports `superstandard`, `substandard`, `heading` sizes, eg. 

Example:
```js
<Rating rating={4} heading />
```
### New prop `showTextRating`
If prop is supplied, rating text will be shown on the right side of the stars
Text respects typography sizes and icons are scaled down for x0.7

Example:
```js
<Rating rating={3.5} superstandard showTextRating />
```

So if used inline it will look like that:
screen shot 2017-12-18 at 9 56 47 am

## withTextProps decorator
Text, Icon and Rating are decorated now with withTextProps decorator which will convert `size` prop to respective boolean size prop and visa versa so both might be used to produce the same result, eg

```js
<Text hero>Wololo!</Text>
<Text size="hero">Wololo!</Text>
```

## Other changes
- spacing is added in between rating icons relative to the font-size
- migrate Rating tests to snapshots
- add a few tests to Text component
- add examples to playground page
-fix gray buttons on playground


react/private/withTextProps.js
```diff
@@ -0,0 +1,39 @@
...
+    return;
+  });
+
+  return sizeProps;
+};
+
+const withTextProps = () => WrappedComponent => {
+  return props => {
+    const newProps = {
+      ...props,
```
markdalgleish: We should be omitting boolean size props here before spreading props, otherwise both will get passed down to `WrappedComponent`. We actually see this in the snapshots in this PR, where `size="heading" heading={true}` etc. is coming through.

react/SlideToggle/__snapshots__/SlideToggle.test.js.snap
```diff
@@ -27,7 +27,7 @@ exports[`Slide toggle: should render a checked state 1`] = `
         />
       </div>
     </div>
-    <Text
+    <Component
```
markdalgleish: Looks like the way we're wiring up the HOC is breaking the display name. Probably worth looking into how we can fix this.

Looks like this breaks the docs too, if you check out the preview site: http://seek-style-guide--c749262dc5ddddd97f329a8a754138f75705374d.surge.sh/text/

react/private/withTextProps.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+const sizes = [
+  'small',
+  'substandard',
+  'superstandard',
```
markdalgleish: Looks like you're missing `standard` here.

react/Rating/Rating.js
```diff
@@ -28,17 +39,39 @@ const getStar = (percent, key, starClassName) => {
   return <StarIcon {...props} />;
 };
 
-const Rating = ({ rating, starClassName, ...restProps }) => {
+const Rating = ({
+  rating,
+  starClassName,
+  showTextRating,
```
markdalgleish: I feel like this should either be a separate PR, or called out in the PR title. The current title makes it sound like only icons are affected.

docs/src/components/Playground/Playground.js
```diff
@@ -468,6 +473,78 @@ export default class Playground extends Component {
...
+        <PageBlock>
+          <Section header>
+            <Text hero>Text with inline icons</Text>
+          </Section>
+          <Card>
+            <Section>
+              {[
+                'small',
+                'substandard',
+                'superstandard',
```
markdalgleish: Missing `standard` here too.

docs/src/components/Playground/Playground.js
```diff
@@ -158,8 +163,8 @@ export default class Playground extends Component {
             <Card>
               <Section>
                 <ButtonGroup>
-                  <Button color="grey">Save job</Button>
-                  <Button color="grey">Send job</Button>
+                  <Button color="gray">Save job</Button>
```
markdalgleish: Haha, nice catch 👍

react/Text/Text.js
```diff
@@ -29,37 +31,35 @@ export default function Text({
...
+    className={classnames({
+      [styles.root]: true,
+      [className]: className,
+      [styles.small]: small,
+      [styles.substandard]: substandard,
+      [styles.superstandard]: superstandard,
+      [styles.subheading]: subheading,
+      [styles.headline]: headline,
+      [styles.heading]: heading,
+      [styles.hero]: hero,
```
markdalgleish: Now that we have the `withTextProps` decorator creating a single `size` prop, we should probably clean this up, like you did in `Icon.js`.

react/Text/Text.test.js
```diff
@@ -0,0 +1,20 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+import Text from './Text';
+
+const sizes = ['small', 'substandard', 'superstandard', 'subheading', 'headline', 'heading', 'hero'];
```
markdalgleish: We're missing `standard` here too.

docs/src/components/Playground/Playground.js
```diff
@@ -468,6 +473,78 @@ export default class Playground extends Component {
           </CardGroup>
         </PageBlock>
 
+        <PageBlock>
+          <Section header>
+            <Text hero>Icons</Text>
```
markdalgleish: Love that you added this to the playground—makes reviewing this so much easier 👏👏👏

docs/src/components/Playground/Playground.js
```diff
@@ -468,6 +473,78 @@ export default class Playground extends Component {
...
+                  ))}
+                </Section>
+              </Card>
+            ))}
+          </Columns>
+        </PageBlock>
+
+        <PageBlock>
+          <Section header>
+            <Text hero>Text with inline icons</Text>
```
markdalgleish: Love that you added this too 👏👏👏

react/private/Icon/Icon.js
```diff
@@ -21,10 +34,14 @@ export default function Icon({ markup, className, svgClassName, ...restProps })
 Icon.propTypes = {
   markup: PropTypes.string.isRequired,
   svgClassName: PropTypes.string,
-  className: PropTypes.string
+  className: PropTypes.string,
+  size: PropTypes.string
```
markdalgleish: I wonder if we should provide a shared set of size prop types, so the the developer experience is a bit nicer 🤔 

Don't have to do this right away, since we don't have a good prop types experience today with the boolean props.

react/private/Icon/Icon.less
```diff
@@ -9,3 +9,47 @@
...
+
+.substandardSvg {
+  .iconSize(@substandard-type-scale);
+}
+
+.smallSvg {
+  .iconSize(@small-type-scale);
+}
+
+.interactionSvg {
```
markdalgleish: This makes me wonder whether we need to make `interaction` more of an official size, so `<Text size="interaction">` works, otherwise you have to remember which components support it and which don't.

Again, don't have to do this right now, just raising the question.

react/Rating/Rating.js
kosanna: Added to title

react/SlideToggle/__snapshots__/SlideToggle.test.js.snap
kosanna: Seems that docs are not broken 🤔 or am I looking in a wrong place?

react/Text/Text.js
kosanna: Was thinking about it, I think we could leave it like that, so we have freedom to use it as we like, and have components coded in any way as well, so it's universal both ways...thoughts?

react/private/Icon/Icon.js
kosanna: Great pickup, do we have a centralised place to have prop types descriptions in styleguide? 🤔 Or I can just put it in the text component / decorator and refer from there

kosanna: Exported from to decorator 👍

react/private/withTextProps.js
kosanna: That's by design, that's the idea, so you can use both

react/SlideToggle/__snapshots__/SlideToggle.test.js.snap
markdalgleish: If you look at the link above, the JSX example says `<No Display Name hero>`

markdalgleish: Also says this, which made me laugh:

```js
import {
  No
} from 'seek-style-guide/react';
```

react/Text/Text.js
markdalgleish: Sorry, not sure what you mean—might need to catch up about this one.

react/private/withTextProps.js
markdalgleish: You should still be able to use both, this decorator just allows you to simplify the decorated component by taking care of boolean props for you. Might pair on this one :)

react/SlideToggle/__snapshots__/SlideToggle.test.js.snap
kosanna: Ahahha, yeah, I see now :D Lol

react/private/Icon/Icon.less
kosanna: Yeah, I was thinking about that too, I think we need to add it

react/private/withTextProps.js
```diff
@@ -0,0 +1,44 @@
...
+      sizeProps.size = props[size] ? size : sizeProps.size;
+    }
+
+    return;
+  });
+
+  return sizeProps;
+};
+
+const withTextProps = () => WrappedComponent => {
```
markdalgleish: Is there a reason for the extra arrow function here?

react/Text/Text.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+describe('Text', () => {
+  it('should render with defaults', () => {
+    expect(shallow(<Text>Hello</Text>)).toMatchSnapshot();
+  });
+
+  describe('sizes', () => {
+    sizes.forEach(size => {
+      it(`should render as ${size}`, () => {
+        expect(shallow(<Text {...{ [size]: true }}>Hello</Text>)).toMatchSnapshot();
```
markdalgleish: This test probably shouldn't be shallow either.

react/private/withTextProps.js
```diff
@@ -0,0 +1,46 @@
...
+  'substandard',
+  'superstandard',
+  'standard',
+  'subheading',
+  'headline',
+  'heading',
+  'hero'
+];
+
+export const SizePropTypes = PropTypes.oneOf(sizes);
```
markdalgleish: I think this should probably be an object that contains the boolean props too, which then gets spread into other proptypes objects, maybe renamed to something like `withTextPropTypes`, e.g.

```js
Text.propTypes = {
  ...withTextPropTypes
};
```

theme/mixins/iconSize.less
```diff
@@ -0,0 +1,12 @@
+.iconSize(@typography-scale, @typography-scale-mobile: @typography-scale, @icon-scale: 0.7) {
+  @icon-size-baze: @base-font-size * 1px * @icon-scale;
```
markdalgleish: Should this be `base`?

react/private/withTextProps.js
```diff
@@ -0,0 +1,56 @@
...
+  'subheading',
+  'headline',
+  'heading',
+  'hero'
+];
+
+const getBooleanSizes = () => {
+  const booleanProps = {};
+
+  forEach(sizes, size => booleanProps[size]: PropTypes.bool);
```
markdalgleish: Shouldn't this be an assignment? `booleanProps[size] = PropTypes.bool`?

react/private/Icon/Icon.less
markdalgleish: Are we adding it in this PR? What do you think?

react/Rating/Rating.demo.js
```diff
@@ -19,6 +33,20 @@ export default {
             rating
           })
         }))
+    },
+    {
+      label: 'Size',
+      type: 'radio',
+      states: ['substandard', 'superstandard', 'heading']
```
markdalgleish: These should be title case, and `Standard` should probably be the first option, removing `substandard` from the initial props.

react/Rating/Rating.demo.js
```diff
@@ -5,9 +5,23 @@ export default {
   title: 'Rating',
   component: Rating,
   initialProps: {
-    rating: 0
+    rating: 0,
+    substandard: true
```
markdalgleish: Still need to remove this :)

kosanna:
All comments are addressed, 👀  👓  are welcome

react/private/withTextProps.js
```diff
@@ -0,0 +1,60 @@
...
+  'hero'
+];
+
+const getBooleanSizes = () => {
+  const booleanProps = {};
+
+  forEach(sizes, size => {
+    booleanProps[size] = PropTypes.bool;
+
+    return;
```
markdalgleish: This return is redundant :)

react/private/withTextProps.js
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  return sizeProps;
+};
+
+const withTextProps = OriginalComponent => {
+  const DecoratedComponent = props => {
+    const newProps = {
+      ...parseBooleanSize(props),
+      ...omit(props, sizes)
```
michaeltaranto: The order of these should be flipped I think, if `props` contained `size` as well as one of `sizes`, it would override the result of `parseBooleanSize` which is not desired right?

react/private/withTextProps.js
```diff
@@ -0,0 +1,58 @@
...
+  'heading',
+  'hero'
+];
+
+const getBooleanSizes = () => {
+  const booleanProps = {};
+
+  forEach(sizes, size => {
+    booleanProps[size] = PropTypes.bool;
+  });
```
michaeltaranto: We should be returning `booleanProps` here after the `forEach`

kosanna: Fixed

react/private/withTextProps.js
kosanna: Depends on what we want to achieve here and which ones are more important. Could just add a proptype check that either boolean or size should be supplied and not both

kosanna: Because technically that's not the usecase

kosanna: Added custom proptypes checker :)

PR #397, opened 2017-12-22
feat(Header): New badge for header links

georgespyropoulos:
Various teams would like the ability to draw attention to new features in their app. This change
allows a new badge to be switched on header navigation items with a new prop for the header `newBadgeTab`

EXAMPLE USAGE:

```js
<Header
  activeTab="Profile"
  authenticationStatus="authenticated"
  divider={false}
  newBadgeTab="Profile"
  userName="Olivia"
/>
```


markdalgleish:
Just checked out the preview site, and the "Saved & Applied Jobs" and "Company Reviews" badges don't seem to work: http://seek-style-guide--44f6bf419c730404b3af80e6472d95c4b2ca97df.surge.sh/header

react/Header/Navigation/Navigation.js
```diff
@@ -59,11 +60,20 @@ export default function Navigation({ locale, linkRenderer, activeTab, divider })
...
+                          className={styles.newBadge}
+                        />
+                      )
+                    ],
                     'data-analytics': analytics,
                     className: classnames({
                       [styles.link]: true,
-                      [styles.link_isActive]: name === activeTab
+                      [styles.link_isActive]: name === activeTab,
+                      [styles.link_smallText]: locale === 'AU' && newBadgeTab === 'Company Reviews'
```
michaeltaranto: I understand why we are doing this but it still doesnt feel right, are we happy to drop the font size of the whole primary nav to accomodate a new badge? 🤔

react/Header/Navigation/Navigation.less
```diff
@@ -54,3 +56,16 @@
     box-shadow: inset 0 -2px @sk-white;
   }
 }
+
+.link_smallText {
+  font-size: 16px;
```
michaeltaranto: If we retain this behaviour I would recommend using the `Text` component and using the boolean flags to select another hierarchy element conditionally rather than apply bespoke font sizing. It keeps all `font-size` declarations inside of `Text` which is desirable

react/Header/Navigation/Navigation.less
```diff
@@ -54,3 +56,16 @@
...
+}
+
+.newBadge {
+  margin-left: 5px;
+  @media only screen and (max-width: (@grid-container-width + @grid-gutter-width)) {
```
michaeltaranto: For simplicity i dont see the need to add a single gutter width here. The container width seems to work fine IMHO.

react/Header/NewBadge/NewBadge.less
```diff
@@ -0,0 +1,17 @@
+@import (reference) "~seek-style-guide/theme";
+
+@height: 18px;
+
+.root {
+  content: 'New';
```
michaeltaranto: This looks like it's left over from a pseudo element.

react/Header/NewBadge/NewBadge.less
```diff
@@ -0,0 +1,17 @@
+@import (reference) "~seek-style-guide/theme";
+
+@height: 18px;
```
michaeltaranto: Should this be two grid rows or they just happen to be equal?

react/Header/NewBadge/NewBadge.less
```diff
@@ -0,0 +1,17 @@
...
+
+.root {
+  content: 'New';
+  display: inline-block;
+  text-transform: uppercase;
+  font-size: 12px;
+  font-weight: @sk-medium;
+  color: @sk-white;
+  background-color: @sk-link;
```
michaeltaranto: The contrast of this when on the active tab (or in the mobile menu) is quite low. Are we happy with that? (Given the text is still accessible)

react/Header/NewBadge/NewBadge.less
```diff
@@ -0,0 +1,17 @@
...
+  height: @height;
+  line-height: @height;
+  padding: 0 6px;
+  border-radius: @height / 2;
```
michaeltaranto: Should the border radius relate to the height?

react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -94,6 +95,13 @@
   }
 }
 
+.newBadge {
+  margin-right: (@grid-gutter-width / 4);
```
michaeltaranto: This is probably better expressed as `1ch` or factor of. Given we want it to be closer to a space rather than relate to the gutter width. If you are concerned about below Safari 7 i'd suggest `1ex` as a fallback. Scaling to factors of these units will keep the spacing proportionate to the text which i think is the goal

react/Header/Navigation/Navigation.js
markdalgleish: I tried uncommenting this line, and it seemed fine to me. The new badge is placed on top of the tab when the screen is between 740px and 960px wide, so it doesn't run out of room.

react/Header/NewBadge/NewBadge.less
stephaniemoss: Yes. Navi has seen and approved this. Part of the reason is because we do not have the logic to make the label disappear after being interacted with so we lowered it's visual prominence for this iteration. It used to be pink haha

markdalgleish:
Just pushed a fix for my first comment, adding support for all mobile tabs.

react/Header/NewBadge/NewBadge.less
markdalgleish: We discussed this in person, leaving as is.

react/Header/Header.demo.js
```diff
@@ -80,6 +80,27 @@ export default {
         })
       }))
     },
+    {
+      label: 'New Tab',
+      type: 'radio',
+      states: [
+        undefined, // eslint-disable-line no-undefined
+        'None',
```
markdalgleish: Should this be `null`?

react/Header/Header.js
```diff
@@ -144,6 +152,7 @@ Header.defaultProps = {
   linkRenderer: defaultLinkRenderer,
   authenticationStatus: AUTH_PENDING,
   activeTab: null,
+  newBadgeTab: 'Profile',
```
markdalgleish: 👏👏👏

react/Header/Header.demo.js
```diff
@@ -80,6 +80,33 @@ export default {
...
+        '$150k+ Jobs',
+        'Profile',
+        'Saved & Applied Jobs',
+        'Recommended Jobs',
+        'Company Reviews',
+        'Advice & Tips'
+      ].map(newBadgeTab => {
+        let label = newBadgeTab;
+        if (!newBadgeTab) {
+          label = newBadgeTab === undefined ? 'Default' : 'None';
```
markdalgleish: Sorry, one last nitpick—could this say `Default "new" badge` and `No "new" badge`? The dropdowns aren't labeled, so the default value needs to act as a descriptor too, otherwise it just says "Default" and users won't know what it's referring to.

PR #402, opened 2018-01-05
docs(site): Add polyfills to support IE11

markdalgleish:


package.json
```diff
@@ -79,6 +79,7 @@
     "chalk": "^2.1.0",
     "commitizen": "^2.9.6",
     "commitlint-config-seek": "^1.0.0",
+    "core-js": "^2.5.3",
```
michaeltaranto: You might want to add this to the greenkeeper ignore list too

markdalgleish: Yep, nice catch 👍

markdalgleish: Done.

PR #403, opened 2018-01-08
docs(site): Add component categories, add missing component demos

markdalgleish:


markdalgleish:
You can preview this update here: seek-style-guide--149f9e4f11560c0f7505205384db42402c258222.surge.sh

markdalgleish:
Worth calling out, I'm most interested in feedback about the categories, their names, whether components are grouped correctly, whether we're missing any other categories, etc.

PR #404, opened 2018-01-10
feat(TextLink): Make TextLink understand typography sizes

kosanna:
As we've recently introduced a concept for icons that could be used as text (using same typography props as regular Text component), replacing custom css icon sizing.

TextLink component is now decorated with typography props. So if used with Text component it will resize respectively:

```
<TextLink heading component={Text} />
```

Chevron sizes up/down respectively as per typography.

When upgrading, no action is required. All TextLink component instances will be updated automatically with the new responsive `standard` icon size.

Other minor changes:

- change TextLink tests to `render` instead of `shallow`

### DIFF
--- a/react/TextLink/TextLink.demo.js
+++ b/react/TextLink/TextLink.demo.js
@@ -1,27 +1,19 @@
-import React from 'react';
-import PropTypes from 'prop-types';
 import TextLink from './TextLink';
 import Text from '../Text/Text';
 
-const TextLinkContainer = ({ component: DemoComponent, componentProps }) => (
-  <Text>
-    <DemoComponent {...componentProps} />
-  </Text>
-);
-TextLinkContainer.propTypes = {
-  component: PropTypes.any,
-  componentProps: PropTypes.object.isRequired
-};
+import { sizes } from '../private/withTextProps';
+import capitalize from 'lodash/capitalize';
 
 export default {
   route: '/text-link',
   title: 'Text Link',
   category: 'Typography',
   component: TextLink,
-  container: TextLinkContainer,
   initialProps: {
     href: '#',
-    children: 'Text link'
+    children: 'Text link',
+    component: Text,
+    chevron: 'up'
   },
   options: [
     {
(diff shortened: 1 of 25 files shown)


markdalgleish:
Would you be able to add a bit more detail and rationale to this PR for consumers who are attempting to upgrade?

Also, the build is currently broken due to an issue with Puppeteer on Travis, so I'm going to open a separate PR to unblock the build pipeline.

react/TextLink/TextLink.js
```diff
@@ -13,9 +13,9 @@ const renderChevron = chevron => {
 
   return (
     <ChevronIcon
+      standard
```
michaeltaranto: TextLink may need to be decorated `withTextProps` and pass it through to here, as a `TextLink` relies on being either a parent/child of `Text` to set the font size, the chevron should honour this.

markdalgleish:
Should this be a `fix` instead, since we're not changing the public API?

react/TextLink/TextLink.js
kosanna: Agree, that’s the right way. Was just hardcoded previously, so was not worried. I will change 👍

kosanna:
Now it's a `feature`

react/TextLink/TextLink.demo.js
```diff
@@ -1,27 +1,19 @@
...
   route: '/text-link',
   title: 'Text Link',
   category: 'Typography',
   component: TextLink,
-  container: TextLinkContainer,
   initialProps: {
     href: '#',
-    children: 'Text link'
+    children: 'Text link',
+    component: Text,
```
markdalgleish: Not sure we should be showing this as a standard usage pattern. It seems like `TextLink` is designed to be nested within a `Text` component, so maybe we make the demo show this `Text > TextLink` nesting instead? 🤔

markdalgleish: If we want to do this, we'd probably need to tweak the way demos work, since this works quite differently to existing demos.

kosanna: That feels weird that it's meant to be nested in Text, whilst you could supply Text as a component as well, thus ending up with two nested Text components...plus in case of just wrapping you can't pass the props of the wrapper down 🤔  Obviously if something like anchor tag is wrapped that's okay. If it's meant to be wrapped in Text why don't we just include this Text as an essential part of this component 🤔  hm...

markdalgleish: Might need to chat about this to figure out how to make the API handle this cleanly, it's a bit confusing.

kosanna:
Closing in favour of another one. Perhaps will just pass down iconProps={{ size: 'hero' }} (short terms). 

Long term: split inline and block link components...someday...somwhen :)

PR #407, opened 2018-01-12
fix(Icon): Add missing subheading styles

kosanna:
```
<SomeIcon subheading /> or <SomeIcon size="subheading" />
```

was not styling icon correctly as per subheading size so size was fallbacking to default 16px.

PR fixes it.

Plus playground cleanup

### DIFF
--- a/react/private/withTextProps.js
+++ b/react/private/withTextProps.js
@@ -10,11 +10,11 @@ import forEach from 'lodash/forEach';
 export const sizes = [
   'small',
   'substandard',
-  'superstandard',
   'standard',
+  'superstandard',
   'subheading',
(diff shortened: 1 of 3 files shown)


react/private/withTextProps.js
```diff
@@ -10,11 +10,11 @@ import forEach from 'lodash/forEach';
 export const sizes = [
   'small',
   'substandard',
-  'superstandard',
   'standard',
+  'superstandard',
   'subheading',
-  'headline',
   'heading',
+  'headline',
```
kosanna: That's just a reshuffle so they look better in increasing order on playgound

PR #408, opened 2018-01-12
fix(Footer): Use icon typography

kosanna:
Instead of custom css styles on ChevronIcon added `substandard`.

Looks a bit tidier now too:
image

vs
image


markdalgleish:
Just updated the scope to "Footer" since consumers don't interact with the private "ToggleContainer" component.

react/Footer/ToggleContainer/ToggleContainer.js
```diff
@@ -10,7 +10,7 @@ export default function ToggleContainer({ children, secondary, label, name, 'dat
       <input id={name} className={styles.toggle} type="checkbox" data-automation={dataAutomation} />
       <label className={`${styles.toggleLink} ${styles.toggleLabel}`} htmlFor={name}>
         {label}
-        <ChevronIcon direction="down" className={styles.chevron} svgClassName={styles.chevronSvg} />
+        <ChevronIcon direction="down" className={styles.chevron} superstandard />
```
markdalgleish: Shouldn't this be using the same size as the text rather than `superstandard`? Otherwise it seems like a sign we've got the icon sizing wrong.

kosanna: Happy to go with substandard, however as it's bottom aligned it looks a bit odd due to the icon shape (or maybe it's just me, see screenshot reattached in description). May be it's alright.

react/Footer/ToggleContainer/ToggleContainer.less
```diff
@@ -21,11 +21,6 @@
   transform-origin: 50% 50%;
 }
 
-.chevronSvg {
-  width: 12px;
-  height: 12px;
-}
```
michaeltaranto: Now that we are changing the size of the chevron it no longer feels necessary to [translate it down a pixel](https://github.com/seek-oss/seek-style-guide/pull/408/files#diff-b5d3636b3c1d3de341bb87967b7051feR20), thoughts?

kosanna: Agree, removed

kosanna:
And this one too please 👀  👀  👀

PR #409, opened 2018-01-15
Bump Semantic Versioning Up to 2.0.0

jamesleejiacheng:
JobStreet Header now has a different API contract. And should be considered a breaking change instead of a minor change.

### BREAKING CHANGE:
JobStreet Header component now takes in more props:
```
{
  username: PropTypes.string,
  userToken: PropTypes.string,
  authenticationStatus: PropTypes.oneOf([
    'authenticated',
    'unauthenticated',
    'pending'
  ]).isRequired,
  language: PropTypes.string.isRequired,
  country: PropTypes.string.isRequired,
  activeNavLinkTextKey: PropTypes.string
}
```

With language and country being a must to drive the language, at the moment it only supports Malaysia and English.

### RFC URL: 
N/A

### MIGRATION GUIDE:
Before:
```App/components/Header/Header.js
<JobStreetHeader user={{}} data-automation={currentTenant} />
```

After:
```App/components/Header/Header.js
<JobStreetHeader language={process.env.APP_LOCALE} country={process.env.APP_COUNTRY} activeNavLinkTextKey={header.searchText} data-automation={currentTenant} />
```


jamesleejiacheng:
Sorry wrong target.

PR #410, opened 2018-01-15
chore(sketch): Rename Text symbols to increase usability

markdalgleish:
This solves a few major usability issues with the current `<Text>` symbols:

- Sketch displays symbols in alphabetical order, which means that our type hierarchy is all mixed up, so we now prefix all symbol names with a number.
- Since the text styles are surfaced in the library as symbols, designers have no idea what font sizes they'll be getting, so we now provide the mobile and desktop font sizes in the symbol name.
- The list was exceedingly long because all font variants (`Secondary`, `Strong`, `Positive`, `Critical`) were being displayed at the top level, so these are now nested within the `Superstandard`, `Standard` and `Substandard` menu items.

Here's a couple of screenshots to show what this looks like in practice.

Before:

screen shot 2018-01-16 at 9 54 25 am

After:

screen shot 2018-01-16 at 9 54 36 am

I also added an `html-sketchapp-dev` script so that you can generate Sketch files from the local webpack dev server, rather than having to run through the entire build process while iterating on the Sketch library.


react/Text/Text.sketch.js
```diff
@@ -3,26 +3,26 @@ import mapKeys from 'lodash/mapKeys';
...
-  'Superstandard Critical': <Text superstandard critical>Superstandard critical text</Text>,
-  'Superstandard Secondary': <Text superstandard secondary>Superstandard critical text</Text>,
-  'Superstandard Strong': <Text superstandard strong>Superstandard strong text</Text>,
-  'Substandard': <Text substandard>Substandard text</Text>,
-  'Substandard Positive': <Text substandard positive>Substandard positive text</Text>,
-  'Substandard Critical': <Text substandard critical>Substandard critical text</Text>,
-  'Substandard Secondary': <Text substandard secondary>Substandard critical text</Text>,
-  'Substandard Strong': <Text substandard strong>Substandard strong text</Text>,
-  'Small': <Text small>Small text</Text>
+  '1. Hero — Desktop: 42px, Mobile: 21px': <Text hero>Hero text</Text>,
```
michaeltaranto: These values smell like something that could rot, what about importing the less vars and multiplying them by the base font size so these update automatically?

markdalgleish: Yeah, I had the same concern, but was worried that the solution would be too much of a headache. I'll have a look.

(Wish we had these available in JS rather than Less)

markdalgleish: Done! Nice suggestion, feel so much better about this now 👍

PR #411, opened 2018-01-16
feat(TextLink): Support chevron icon sizing and left direction

kosanna:
- Make TextLink accept an object with `chevronProps` via which chevron icon appearance might be manipulated
- Set default icon size to `standard` to follow the default text sizing rules.

```
<TextLink chevron="right" chevronProps={{ size: 'hero' }}>Text link</TextLink>
```
screen shot 2018-01-11 at 9 07 40 am

Since TextLink is designed to be nested within the Text component, now both Text and Chevron should be supplied with the sizing prop and thus respectively resized.

- Added left direction (e.g. so we can use it for "back" links) which appears on the left of the text

```
<TextLink chevron="left">Text link</TextLink>
```
screen shot 2018-01-16 at 11 55 37 am


kosanna:
❤️   💃   🎉

kosanna:
Could I please get a second pair of eyes on this
     👀
☝️ 👄

PR #412, opened 2018-01-16
refactor(flowtype): Added flowtype

tamoore:
## Background

In discussions in the past there has been some desire for integrating flow type in the style guide so that it can be utilised on an ad-hoc basis. This may have changed and the desire may have waned, but if not this PR sets up the minimum requirements to begin incrementally working towards some types.

## Approach

I just wanted to add a minimal flow config and also a npm script that will allow users to run flow-type at their discretion. It has been integrated into the build pipeline but won't be reporting any errors until we start adding `@flow` declarations to files:

https://flow.org/en/docs/usage/#toc-prepare-your-code-for-flow

## Using

```shell
npm run flow
```

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,7 +9,11 @@
   "plugins": [
     "babel",
     "react",
-    "jest"
+    "jest",
+    "flowtype"
+  ],
+  "extends": [
+    "plugin:flowtype/recommended"
   ],
   "rules": {
     "comma-dangle": 2,
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,18 @@
+[ignore]
+.*/node_modules/react-element-to-jsx-string/.*
+.*/node_modules/stylelint/.*
+.*/node_modules/config-chain/.*
+.*/node_modules/jsx-to-string/.*
+.*/node_modules/immutable/.*
+.*/node_modules/npmconf/.*
+
+[include]
+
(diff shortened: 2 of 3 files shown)


tamoore:
I'm going to update this branch to include what we discussed at FEED. Namely, a consumer focused demo that includes a generated lib definition. I'll also add flow-type proptype transformation in the mix.

tamoore:
So the approach I'm leaning to at the moment is to use babylon to parse an AST of the styleguide and then generate a lib definition based on files that have been marked with an @flow comment.

It might be worth closing this PR and converting the discussion over to an issue? 

The reason for this is two fold:

1. the codemon project michael shared only really deals with the migration of code from propTypes to flowTypes. Also, it's a little dated in it's type definitions.

2. The command on flow-bin that can automatically convert those flow types to a definitions has  a bug at the moment and none of the workarounds appear to work. https://github.com/facebook/flow/issues/3281

I was considering doing a PR to flow to fix it but I think that might be after the fact instead of before as I'm not 100% (or even, in truth, 5%) on OCaml.

I think it would make a nice oss project that solves a problem which I don't think there are any solutions for right now.

markdalgleish:
@tamoore yep, if we're pivoting to something a bit more involved, it probably makes sense to close this PR for now. Thanks for digging into this a bit further 👍

PR #413, opened 2018-01-17
feat(PageBlock): Add support for wide attribute

jahredhope:
# Reason

Some consumers of the seek-style-guide currently use 990px width for some content.
This can create a disjoint between content using PageBlock and other content.
This is not a breaking change but may make future changes harder due to the support of two widths.

# Usage
This change doesn't change the current recommended behaviour, however consumers requiring wider size will need to add the `wide` attribute to any PageBlock to
```js
<PageBlock wide>
  <Section>
    {/* ... */}
  </Section>
</PageBlock>
```


tamoore:
It's been awhile on this PR, still required @jahredhope?

jahredhope:
Closing, to be supported by individual consumers.

PR #414, opened 2018-01-19
feat(Header): Rename ‘Advice & Tips’ to ‘Career Advice’ in AU

markdalgleish:
This PR makes a few notable changes for AU:
- As the title says, rename "Advice & Tips" to "Career Advice"
- Move the "new" badge from "Profile" to "Career Advice"
- Show the Career Advice menu item on mobile, in the lower section

This PR also makes the following general changes:
- Generate a standalone header build of Career Advice
- Remove `react-scrollock` from the mobile menu, since the menu now requires scrolling on 320x480 phones
- Shrink the whitespace between menu groups on mobile

Note that, to ensure this isn't a breaking change, we still support "Advice & Tips" as a prop for `activeTab` and `newBadgeTab`, converting it to "Career Advice" on the way in.

### DIFF
--- a/react/Header/Header.js
+++ b/react/Header/Header.js
@@ -16,6 +16,18 @@ import employerLinkForLocale from './employerLinkForLocale';
 import StructuredDataSchema from './StructuredDataSchema/StructuredDataSchema';
 import { AUTHENTICATED, UNAUTHENTICATED, AUTH_PENDING } from '../private/authStatusTypes';
 
+const defaultNewBadgeTabForLocale = locale => (
+  locale === 'AU' ?
+    'Career Advice' :
+    'Profile'
+);
+
+const handleLegacyTabName = tabName => (
+  tabName === 'Advice & Tips' ?
+    'Career Advice' :
+    tabName
+);
+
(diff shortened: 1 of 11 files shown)


react/Header/Header.js
```diff
@@ -16,6 +16,18 @@ import employerLinkForLocale from './employerLinkForLocale';
 import StructuredDataSchema from './StructuredDataSchema/StructuredDataSchema';
 import { AUTHENTICATED, UNAUTHENTICATED, AUTH_PENDING } from '../private/authStatusTypes';
 
+const defaultNewBadgeTabForLocale = locale => (
+  locale === 'AU' ?
+    'Career Advice' :
+    'Profile'
```
kosanna: I'm just not sure wether we should not have taken off them all (from both AU and NZ). Is it a requirement?

markdalgleish: We're basically maintaining the existing behaviour in NZ for now.

kosanna: Yeah I get it, I just feel that it's a bit not logical, like it stopped being new for Australian people, but not for New Zealand people? Do they think slower than Australians? :/

Issue #330, opened 2017-09-21
RFC: Alert styles


georgespyropoulos:
# Summary

Alert styles are a way to highlight certain pieces of copy to give them the correct visual weight and tone to communicate a message to the user.

Big thanks to @stephaniemoss for putting this together!  I am merely the messenger here

# Motivation

Currently our styling for these types of alerts is inconsistent, so there is no pattern to deal with them. This means they get implemented in different ways across streams and products which can confuse the user.

Having a consistent set of patterns for dealing with alerts helps with overall communication, consistency and gives everyone the tools to make sure their copy is seen and understood in the right contexts.

# Detailed design

The alerts are a set of visual patterns that have different types and a hierarchy that can be applied to a number of different implementations.

### Positive

This alert uses green and ticks. It is used for positive and success alerts to let the user know that something has succeeded.

![image](https://user-images.githubusercontent.com/1896277/30675283-f8f29e44-9ec2-11e7-9660-beeda8aa68ba.png)

### Info

This alert uses purple and an i within a circle (info icon). It is used for important information that a user should definitely read.

![image](https://user-images.githubusercontent.com/1896277/30675302-0b7deb90-9ec3-11e7-8d81-2cf761681005.png)

### Critical

This alert uses pink and an exclamation mark in a triangle (error icon). It is used for errors to let a user know that something has gone wrong and they can’t continue.

![image](https://user-images.githubusercontent.com/1896277/30675324-25e9e01a-9ec3-11e7-8161-b5f59b184600.png)

### Help

This alert uses grey and a question mark in a circle (help icon). It is used for helpful but not critical information, usually accessed voluntarily by the user in case they need further instructions or help.

![image](https://user-images.githubusercontent.com/1896277/30675345-41816726-9ec3-11e7-94fb-d576f86d41d2.png)

### Primary

Used when you want the alert to be loud and draw attention. Dark background, light text, light icons.

![image](https://user-images.githubusercontent.com/1896277/30675473-2750c562-9ec4-11e7-8607-64d9642e6383.png)

### Secondary

Used when you want the alert to be on a normal volume level. Light background, dark text, dark icons, underline.

![image](https://user-images.githubusercontent.com/1896277/30675470-24692150-9ec4-11e7-8144-bb9fb439dd59.png)

### Tertiary

Used when the alert should be integrated into the design and not confused with another type of element like a button. Dark icon and text but no background tint or underline.

[...]

PR #415, opened 2018-01-19
feat(Alert): Add Alert component

georgespyropoulos:
New alert component for the styleguide. The RFC (https://github.com/seek-oss/seek-style-guide/issues/330) has all the details about how it looks and what it does.

Demo:
http://seek-style-guide--1c01440566bdfe7fee00c9d25931229d761cbce6.surge.sh/alert/
http://seek-style-guide--1c01440566bdfe7fee00c9d25931229d761cbce6.surge.sh/section/

When implementing it, @markdalgleish and I came to the decision that a some of the concepts for alerts could be applied to other elements of the styleguide hence the decision to change `<Section />` to take props for `primary|secondary` and `positive|info|error|help`. `<Alert />` uses a `<Section />` internally to achieve it's styling. 

I've also added a slim prop for section eg. `<Section slim />`, this allows you to have a section with no top / bottom padding.

RFC URL:  
https://github.com/seek-oss/seek-style-guide/issues/330

EXAMPLE USAGE:

```js
<Alert
  type="positive"
  level="primary"
  message="I'm some timely text that helps the user understand something"
  showCloseButton
  onClose={() => {...}}
/>
```

Closes https://github.com/seek-oss/seek-style-guide/issues/330


react/Section/Section.js
```diff
@@ -4,14 +4,40 @@ import React from 'react';
 
 import styles from './Section.less';
 
-export default function Section({ children, className, header, ...restProps }) {
+export const TYPE = {
+  POSITIVE: 'positive',
+  INFO: 'info',
+  CRITICAL: 'critical',
+  HELP: 'help'
+};
```
markdalgleish: 👏 I quite like this pattern, I wonder if we should start doing this more 🤔

react/Section/Section.less
```diff
@@ -1,11 +1,74 @@
 @import (reference) "~seek-style-guide/theme";
 
+@gutter-adjustment: 100px;
```
markdalgleish: This `100px` value feels like a bit of a magic number. Would it make sense to define it using a calculation instead?

react/Alert/Alert.demo.js
```diff
@@ -0,0 +1,122 @@
...
+
+Container.propTypes = {
+  component: PropTypes.any,
+  componentProps: PropTypes.object.isRequired
+};
+
+export default {
+  route: '/alert',
+  title: 'Alert',
+  category: 'Typography',
```
markdalgleish: I don't think this component belongs under "Typography", but I'm not sure where we'd put it 🤔 

I feel like "Form" is a more accurate category, at least for now, even if it's useful in other contexts, but definitely open to other ideas here.

react/Alert/Alert.js
```diff
@@ -0,0 +1,91 @@
...
+import HelpIcon from '../HelpIcon/HelpIcon';
+import CrossIcon from '../CrossIcon/CrossIcon';
+
+import { TYPE, LEVEL } from '../Section/Section';
+
+const ICONS = {
+  [TYPE.POSITIVE]: TickCircleIcon,
+  [TYPE.INFO]: InfoIcon,
+  [TYPE.CRITICAL]: CriticalIcon,
+  [TYPE.HELP]: HelpIcon
```
markdalgleish: 👏👏👏

react/Alert/Alert.test.js
```diff
@@ -0,0 +1,78 @@
...
+    const alert = renderAlert({ hideIcon: true });
+    expect(alert).toMatchSnapshot();
+  });
+
+  it('should render a close button', () => {
+    const alert = renderAlert({ showCloseButton: true });
+    expect(alert).toMatchSnapshot();
+  });
+
+  it('should call a passed in onClose function when the checkbox fires a change event', () => {
```
markdalgleish: Should be "button fires a click event".

react/Section/Section.js
```diff
@@ -21,5 +47,16 @@ export default function Section({ children, className, header, ...restProps }) {
 Section.propTypes = {
   children: PropTypes.node.isRequired,
   className: PropTypes.string,
-  header: PropTypes.bool
+  header: PropTypes.bool,
+  pullout: PropTypes.bool,
+  slim: PropTypes.bool,
+  type: PropTypes.oneOf([TYPE.POSITIVE, TYPE.INFO, TYPE.CRITICAL, TYPE.HELP]),
```
markdalgleish: Should we refer to this as "tone" instead, since that's the terminology we've already been using? e.g. http://seek-oss.github.io/seek-style-guide/typography

react/Section/Section.js
```diff
@@ -4,14 +4,40 @@ import React from 'react';
...
+  SECONDARY: 'secondary',
+  TERTIARY: 'tertiary'
+};
+
+export default function Section({
+  children,
+  className,
+  header,
+  pullout,
+  slim,
```
markdalgleish: Might be worth adding this one to the demo page too.

react/Alert/Alert.js
```diff
@@ -0,0 +1,91 @@
...
+
+    const Icon = type ? ICONS[type] : null;
+
+    return (
+      <div className={styles.alert}>
+        {(!hideIcon && Icon) && <Icon className={styles.icon} />}
+        <div className={styles.text}>
+          <Text raw baseline={false}>{message}</Text>
+        </div>
+        {showCloseButton && (
```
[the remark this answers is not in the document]
georgespyropoulos: Hmmmm yes I suppose we can. Can't think of a case where you would want the close button without a click handler

theme/palette/elements.less
```diff
@@ -10,3 +10,11 @@
 @sk-footer: @sk-mid-gray-light;
 @sk-background: @sk-gray-light;
 @sk-form-accent: @sk-focus;
+@positive-light: #f3f9f2;
```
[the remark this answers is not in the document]
georgespyropoulos: Yeah I guess we probably should. We've prefixed literally every colour with `sk-` so it feels kind of meaningless now

react/Section/Section.less
georgespyropoulos: The definition is 'bigger than the gutter' so I'll go with `@gutter-adjustment: @grid-gutter-width + 50px;` unless there's a better suggestion

react/CriticalIcon/CriticalIcon.js
```diff
@@ -0,0 +1,10 @@
+import svgMarkup from './CriticalIcon.svg';
+
+import React from 'react';
+import Icon from '../private/Icon/Icon';
+
+export default function CriticalIcon(props) {
```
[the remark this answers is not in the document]
georgespyropoulos: The `<Icon />` component is already decorated with that prop and that appears to be the way we've applied it to other icons...  am I missing something or is the pattern applied already?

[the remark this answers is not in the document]
markdalgleish: Ah, don't mind me then :facepalm:

georgespyropoulos:
@markdalgleish I've addressed all of your feedback

tamoore:
Epic piece of work. 👏 Can't wait to use this component myself, looks great and the api is sweeet.

PR #425, opened 2018-02-02
chore(sketch): Add symbols for Dropdown, Textarea, Text links

markdalgleish:


react/Dropdown/Dropdown.demo.js
```diff
@@ -1,8 +1,8 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-
 import Dropdown from './Dropdown';
 import styles from './Dropdown.less';
+import * as sketch from './Dropdown.sketch';
```
tamoore: It looks like you only have the one export, I'd probably just use the name `symbol` directly.

react/Header/Header.sketch.js
```diff
@@ -12,7 +12,7 @@ DummyLinkRenderer.propTypes = {
 
 export const blockSymbols = {
   'Header/Australia/Unauthenticated': <Header linkRenderer={DummyLinkRenderer} authenticationStatus="unauthenticated" />,
-  'Header/Australia/Authenticated': <Header linkRenderer={DummyLinkRenderer} authenticationStatus="authenticated" userName="Olivia" />,
+  'Header/Australia/Authenticated': <Header linkRenderer={DummyLinkRenderer} authenticationStatus="authenticated" userName="Name" />,
```
tamoore: Cute userName

react/Dropdown/Dropdown.demo.js
markdalgleish: It's imported this way so that we can pass the `sketch` object as part of the demo spec further down the file, which is what renders the sketch content to the page.

tamoore: So the shape of that object is something like:
```javascript
{
  symbols: {
    [key: string]: React$Node
  }
}
```

Just trying to understand it, all makes sense to me and you know the most about this code.

markdalgleish: Yep, exactly 👍

react/Dropdown/Dropdown.sketch.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Dropdown from './Dropdown';
+import styles from './Dropdown.sketch.less';
+
+const Container = ({ children }) => (
+  <div style={{ maxWidth: '520px' }}>
```
michaeltaranto: I feel like we should capture this boilerplate somewhere to be shared between all sketch files. Thoughts?

markdalgleish: Yep, makes sense. At some point, it's probably worth discussing whether this rule should be applied to the fields themselves, since we don't generally want fields to get wider than this, regardless of the surrounding layout.

markdalgleish: Just pushed a refactor 👍

PR #426, opened 2018-02-02
[style] Added transition for the input

franciscop:
Added transition for the input to make it more pleasant to use.


markdalgleish:
First of all, thanks for taking an interest in our work—it means a lot.

As I'm sure you can appreciate, changes of this nature typically need to go through the proper channels internally. I'd even say it's likely we'll make a change like this in the future, but if and when we do, it will be in consultation with our design team and include a more holistic upgrade to our entire component suite.

Closing for now, but thanks again 👍

Issue #420, opened 2018-01-29
Adding FlowType to the styleguide


tamoore:
Just adding an issue as a catch all discussion area for https://github.com/seek-oss/seek-style-guide/pull/412

The current plan is to use babylon to create an AST of the styleguide and build a library definition from proptypes defined per component.

## Spec

The plan at the moment is just to support a pretty trivial FS whereby one file exports all the components we want in the lib definition (basically how the style guide is currently set up).

It will use a generic placeholder type for any component that hasn't got a leading '@flow' comment at the top of the file.

tamoore:
The current code for this module is located here: https://github.com/tamoore/styleguide-flow-type-generator

tamoore:
Currently some constraints have arisen for an MVP:

- We won't be able to use locally defined types as properties of component 'Props'. We can use globally defined types but a few details with regards to that are TBD.
- We'll need to refer to all React types via the `$` notation `React$Node` etc.

tamoore:
#422

tamoore:
Verify that [flow-copy-source](https://github.com/AgentME/flow-copy-source) definitely won't work for out use case.

I have tried it in the past, but I possibly got a false negative. So let's just revisit it and ensure it won't work (or will, hopefully).

tamoore:
I tried the above solution with both `focus-check` and a usual `check` and it didn't report type errors for me, so I'm going to continue with the approach outlined as currently it's the only approach that reports type errors on a component.

PR #427, opened 2018-02-04
feat(index): Basic flow support

tamoore:
## Purpose

In discussions over the last few months we've come to the conclusion that we want to add flow support to the style-guide primarily for the developer experience benefits.

## Approach

I looked at this from a load of different angles. Ultimately, I realised that a lib def is not required for the styleguide and the reason that I thought it was came down to a config setting in one of the projects here at SEEK internally. Therefore, whilst it's not a bad idea to also have one, we can just go ahead with defining props as flow types and using a babel transformer to convert those to runtime props.

That's the approach I've taken here.

## Notes.

A couple of things to be aware of. Projects, when updating the styleguide, will need to also update their webpack decorator to [v2.2.0](https://github.com/seek-oss/seek-style-guide-webpack/releases/tag/v2.2.0)

I have a PR for sku which also adds flow type stripping and proptype transformation:
https://github.com/seek-oss/sku/pull/122


markdalgleish:
Just as an update, had a chat yesterday with @tamoore so I could better understand how this works from a consumer's perspective, and it sounds pretty good. The plan is to add a section to the readme to help guide Flow users in terms of what configuration is necessary on their end, once we've confirmed what the minimal config requirements are.

tamoore:
Have added a README section for this PR. Let me know if anything is unclear.

tamoore:
Fixes #420

README.md
```diff
@@ -330,6 +330,34 @@ Then, include the following in your Babel config:
...
+[lints]
+
+[options]
+# This is required to prevent errors to less file imports
+module.name_mapper.extension='less' -> '<PROJECT_ROOT>/no-declarations.js.flow'
+
+# Good idea to ignore json too
+module.name_mapper.extension='json' -> '<PROJECT_ROOT>/no-declarations.js.flow'
+```
+> *no-declaratins.js.flow* is just an empty file
```
markdalgleish: Should be `no-declarations`.

PR #428, opened 2018-02-05
fix(transparentBackground): Setting a max width on desktop

tamoore:
## Purpose

On desktop (firefox) the backdrop currently overflows the page as it's a fixed
position.

## Approach

For desktop, setting a max-width to 100vw resolves the issue and has no impact
on functionality.

image

### DIFF
--- a/theme/mixins/backdrop.less
+++ b/theme/mixins/backdrop.less
@@ -7,6 +7,11 @@
   transition: opacity 0.3s;
   z-index: @z-index-backdrop;
   opacity: 0;
+  @media @desktop {
+    // This is required as otherwise on desktop the backdrop
+    // Overflows the page.
+    max-width: 100vw;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


theme/mixins/backdrop.less
```diff
@@ -7,6 +7,11 @@
   transition: opacity 0.3s;
   z-index: @z-index-backdrop;
   opacity: 0;
+  @media @desktop {
+    // This is required as otherwise on desktop the backdrop
+    // Overflows the page.
+    max-width: 100vw;
```
markdalgleish: This mixin also defines `position: fixed` and `left: -999px`. Wouldn't this mean that it doesn't actually cover the full width of the screen on desktop anymore?

tamoore:
The backdrop max-width approach was not quite right. Cleaning up this mixin first.

PR #429, opened 2018-02-05
fix(transparentBackdrop): Removing -999px left, top, right

tamoore:
## Purpose

In trying to debug a firefox issue in discover, we found that the
transparentBackgroup mixin really needed a lil refactor.

## Approach

We are setting just top and left to 0 and then setting the width and height to
the viewport (minus 1 for an issue with firefox see the comments).

This then standardished, for discover at least the placement of the backdrops,
which have an altered stacking context. We can therefore adjust these with JS
for now until we get around updating the searchBar not to use transforms (if
possible).

## Notes
This shouldn't impact other styleguide consumers, but if you do have an issue
whereby your stacking context changes it might be good to double check this does
not break your UI in subtle ways.

### DIFF
--- a/theme/mixins/backdrop.less
+++ b/theme/mixins/backdrop.less
@@ -1,9 +1,9 @@
 .transparentBackdrop() {
   position: fixed;
-  top: -999px;
-  left: -999px;
-  right: -999px;
-  bottom: -9999px;
+  top: 0;
+  left: 0;
+  width: 100vw;
+  height: 100vh;
   transition: opacity 0.3s;
   z-index: @z-index-backdrop;
   opacity: 0;


theme/mixins/backdrop.less
```diff
@@ -1,9 +1,13 @@
...
-  right: -999px;
-  bottom: -9999px;
+  top: 0;
+  left: 0;
+  // calc(100vw -1) is an adjustment for FF because it adheres to w3c spec
+  // in that it includes the scrollbars. This however is problematic
+  // for us as it means there is still a small space that overflows:
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=811403
+  width: calc(100vw - 1);
+  height: 100vh;
```
mikehdt: Given a fixed position element, wouldn't `right: 0; bottom: 0;` give the same result as setting `width` and `height`, but without the need to deal with the viewport units and scrollbars issue? 🤔

I could be wrong though, I'm not so familiar with the styleguide itself.

tamoore: Give it a shot Mike if you got time, be great to simplify this further if possible.

theme/mixins/backdrop.less
```diff
@@ -1,9 +1,13 @@
...
-  left: -999px;
-  right: -999px;
-  bottom: -9999px;
+  top: 0;
+  left: 0;
+  // calc(100vw -1) is an adjustment for FF because it adheres to w3c spec
+  // in that it includes the scrollbars. This however is problematic
+  // for us as it means there is still a small space that overflows:
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=811403
+  width: calc(100vw - 1);
```
georgespyropoulos: Why a unit-less one? Does the adjustment need to be unit-less or could we specify this as a pixel?

tamoore: We may actually take a slightly different approach, but my brain is fried today (I need to verify it integration side e.g. chalice). So will try it out tomorrow morning and see if the approach Mike mentioned above works. It'd be simpler and generally better IMHO.

tamoore:
@georgespyropoulos @mikehdt I tried the idea of not using width and height but the integration on the discover side didn't yield a good outcome. 

I think as it stands it's perhaps not gold plated, but it's better than what it was. Thoughts?

PR #431, opened 2018-02-06
fix(Section): Fix overflow issues on Section components

georgespyropoulos:
Recent changes to the `<Section>` component introduced `overflow-x: hidden` to accomodate the pullout style for the `<Alert>` component. In IE, setting overflow-x creates a greyed out vertical scrollbar (screenshot below). 

screen shot 2018-02-07 at 9 56 00 am

Advertiser also has some issues with overflow being on `<Section>`

screen shot 2018-02-07 at 10 14 03 am

This problem is being fixed by removing overflow and adjusting the negative margin used in the pullout style to be exactly the same width as the gutter. This means overflow is not necessary because the pullout style no longer breaks out of a Section

### DIFF
--- a/react/Section/Section.less
+++ b/react/Section/Section.less
@@ -4,7 +4,7 @@
 
 .root {
   overflow-x: hidden;
-  -ms-overflow-style: -ms-autohiding-scrollbar;
+  -ms-overflow-style: none;
   padding: (@grid-row-height * 2) @grid-gutter-width;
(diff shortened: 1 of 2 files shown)


georgespyropoulos:
Alternatively, this can be fixed by using `overflow: hidden` as opposed to `overflow-x: hidden`. I was trying to avoid setting overflow-y but perhaps it's the cleaner solution. Thoughts?

PR #436, opened 2018-02-09
fix(Alert): Ensure close icon is right aligned

markdalgleish:
If the provided message isn't long enough to wrap onto multiple lines, the close icon is no longer right aligned, instead being rendered immediately after the text:

screen shot 2018-02-09 at 11 41 11 am

I discovered this issue while creating Sketch symbols for the Alert component, which is why they're also included in this PR.

Also, while ensuring this fix also worked in IE11, I discovered a polyfill on the demo site, which I've also included in this PR.

### DIFF
--- a/react/Alert/Alert.less
+++ b/react/Alert/Alert.less
@@ -11,10 +11,6 @@
 .alert {
   display: flex;
   align-items: center;
-
-  &.hideIcon.showCloseButton {
-    justify-content: space-between;
-  }
 }
 
 .positive {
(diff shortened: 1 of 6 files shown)


react/Alert/Alert.less
```diff
@@ -42,6 +42,7 @@
   display: inline-block;
   vertical-align: middle;
   max-width: ~"calc(100% - (@{icon-width} + @{alert-gutter}))";
+  flex-grow: 1;
```
georgespyropoulos: Using `flex-grow` in this rule is a cleaner solution that the various max-width rules we have for `.text`. I think we can remove all the max width rules here?

markdalgleish: Just pushed an update—I was actually able to remove quite a few styles. The close icon moved a few pixels to the right, if that matters, but it looks okay to me.

Issue #312, opened 2017-09-08
RFC: New 6px baseline grid to enable 16px / 24px standard text


markdalgleish:
# Summary

Migrate from a 9px grid to a 6px grid to allow for larger 16px body copy over a 24px line height across the site.

# Motivation

Our current `standard` text size is 14px / 18px on desktop and 18px / 27px on mobile. We also have `substandard` and `superstandard` variants which, respectively, maintain 14px / 18px or 18px / 27px across both desktop and mobile. In many cases, both of these are causing us issues:

- For dense, skimmable content (e.g. search results), 14px is smaller than we would like on mobile, but 18px is much too big. We'd choose a font size between these two, but it doesn't sit correctly on our current 9px grid, generating excess amounts of leading.
- For more long-form content on desktop, 18px seems to be more preferable, but our heading sizes were not designed with 18px body copy in mind. When using `superstandard` text in combination with the full range of headings, it can lead to very unbalanced typography.

Another issue to consider is that, in practice, the 9px grid often creates too much whitespace by forcing elements to either be too close or too far from each other. If we can improve the amount of design flexibility, without undoing all the benefits of having a baseline grid, that would be a positive outcome too.

# Detailed design

If we want to properly solve this issue, we find ourselves naturally gravitating towards a pretty substantial change—discarding the current 9px grid, settling on a larger body copy size that works *everywhere* (without any sub/super variant hacks), and using this to form the basis a new baseline grid.

From a design perspective, we quickly landed on 16px / 24px, which leaves us with a grid system based on either 8px, 6px or 4px. After much experimentation across a wide range of projects, we landed on 6px as it provided the best balance between project-level flexibility and system-level rigidity.

Since the grid variables are used in many calculations throughout the site, changing their values will no doubt cause widespread breakage. As a result, I think we should introduce entirely new grid variables so that consumers have a chance to upgrade without being forced to convert every single grid-based calculation at once, e.g.

```less
@grid-row: 6px; // New
@grid-row-height: 9px; // Old
 ```

Our horizontal grid values shouldn't need to change, but, for consistency, we might introduce new names for these too, e.g.

```less
@grid-gutter: 20px; // New
@grid-gutter-width: 20px; // Old
 ```

So far, the new type hierarchy based on a 6px baseline grid looks like this:

Desktop:
- Substandard/Superstandard: DEPRECATED 🎉
- Standard: 16px / 4 (24px)
- Small: 14px / 3 (18px)
[...]

PR #437, opened 2018-02-13
feat(grid): Migrate to 6px baseline grid with 16px standard type

markdalgleish:
## RFC

https://github.com/seek-oss/seek-style-guide/issues/312

## Reason for Change

In order to support a larger standard body copy size of 16px / 24px (instead of 14px / 18px), we've also had to change the underlying baseline grid from 9px to 6px.

## Breaking Changes

- You will need to perform a design audit of your project, since the underlying grid and typography has changed significantly.
- The `superstandard` and `substandard` type variants have been removed from React components and Less variables/mixins. Consumers should use `standard` type instead.

### Design Audit

During the migration process, and at the end as a final review, you should seek guidance from the designer(s) on your team, or those that you collaborate with most often.

The grid and typography changes affect a large portion of the components, variables and mixins provided by this library. Even in cases where the public API of a component doesn't change, its internal dimensions may have changed, which could have flow-on effects for the design of your product.

### Grid Variables

Use the new `@row-height` variable instead of `@grid-row-height`, but be aware that since the value has changed from `9px` to `6px`, **you'll need to revisit any existing calculations in your project.**

For example:

```diff
-@grid-row-height * 2
+@row-height * 3
```

When changing calculations like this, you're encouraged to visually check the result of each change, but you can use this table as a quick migration reference to speed up the process:

<table>
<tr><th>@grid-row-height * X</th><th>@row-height * X</th></tr>
<tr><td>1</td><td>2</td></tr>
<tr><td>2</td><td>3</td></tr>
<tr><td>3</td><td>5</td></tr>
<tr><td>4</td><td>6</td></tr>
<tr><td>5</td><td>7</td></tr>
<tr><td>6</td><td>9</td></tr>
<tr><td>7</td><td>11</td></tr>
</table>

If you arrive at different values in different scenarios, that's totally fine.

For consistency, all grid variables have been updated to follow the same naming convention, but since their values remain the same, you can safely rename all existing usages:

```diff
-@grid-gutter-width
+@gutter-width

-@grid-column-width
+@column-width

-@grid-container-width
+@container-width
```

To reduce breakage of custom layouts, the old grid variables are still provided with their existing values, but **you're encouraged to migrate away from the old variables as soon as possible.** Until you fully migrate all grid variables, your custom layouts might not necessarily be broken, but they won't be following the new grid.

### Typography

All typography has been adjusted to fit the new grid, with most changes simply trying to match the existing design as closely as possible, with some notable exceptions.

**The `superstandard` and `substandard` variants have been removed.** In most cases, you should use standard type instead.

For example, when using the `Text` component:

```diff
-<Text substandard>...
+<Text>...

-<Text superstandard>...
+<Text>...
```

Or, for older projects, when using the Less mixins:

```diff
-.superstandardText()
+.standardText()

-.superstandardTextResponsive()
+.standardTextResponsive()

-.substandardText()
+.standardText()

-.substandardTextResponsive()
+.standardTextResponsive()
```

Even though we encourage the use of larger body copy, **for some existing usages of `substandard`, you may want to keep the text at `14px`.** In these cases, you should switch to `small` text instead:

```diff
-<Text substandard>...
+<Text small>...
```


react/Text/Text.less
```diff
@@ -15,29 +15,21 @@
...
-      padding: 0 @grid-gutter-width;
+      padding: 0 @gutter-width;
     }
   }
 }
 
 .standard {
   &:not(.baseline) { .standardTextResponsive(@baseline: false); }
   &.baseline { .standardTextResponsive(@baseline: true); }
+  padding-bottom: @row-height;
```
michaeltaranto: This should be added to `small` too i feel

react/Autosuggest/Autosuggest.less
```diff
@@ -39,19 +39,15 @@
 }
 
 .suggestionsContainer_withLabel {
-  @labelRowOffset: 4;
-  top: @grid-row-height * (@suggestionContainerRowsOffset + @labelRowOffset);
-  @media @desktop {
-    @labelRowOffset: 3;
-    top: @grid-row-height * (@suggestionContainerRowsOffset + @labelRowOffset);
-  }
+  @labelRowOffset: 6;
```
michaeltaranto: I feel like this works better as `5`. The spacing between the field (when the field has a label) and autosuggest list. At the moment it gives 2 rows, but i prefer 1. Thoughts? (Also the label is 5 rows high now that we have reduced standard copy to only have one spacing row below)

georgespyropoulos: Yeah I agree, I think 5 is better

markdalgleish: I'm out today, so feel free to update the PR if you like 👌

michaeltaranto: Done

michaeltaranto:
Also pushed the optimisation around the descender scale, made this align more perfectly with Roboto, while not disturbing the rhythm in our fallback font stack

docs/src/components/App/Header/Header.less
```diff
@@ -129,8 +129,10 @@
   color: white;
   text-decoration: none;
   display: inline-block;
-  margin-bottom: @grid-row-height;
   &:hover {
     box-shadow: inset 0 -2px 0 @sk-white;
   }
+  @media @desktop {
+    margin-bottom: (@row-height * 2);
+  }
```
michaeltaranto: This doesnt look right on the docs site, the links in the menu on desktop appear too loose. I'd just remove it.

react/Columns/Columns.less
```diff
@@ -10,15 +10,12 @@
   @media @desktop {
     flex-basis: 0;
     flex-grow: 1;
-
-    min-width: 0;
```
michaeltaranto: This `min-width` looks like it is required to support the flexible `Columns` work implemented in https://github.com/seek-oss/seek-style-guide/pull/350

markdalgleish:
@michaeltaranto Updated 👍

react/Columns/Columns.less
markdalgleish: Nice catch 👏

PR #438, opened 2018-02-16
fix(icons): Updated delete and plus icons to use private icon component

tgdev:
The Delete and Plus icon components were not using the base Icon component and therefore not inheriting the root icon styles. 

This was causing these two components to use the inline width and height attributes defined in the svg files, rendering them larger than the other icons.

The changes were made to bring these icons in line with the other icon components used in seek style guide to ensure consistency in styling and implementation.

MIGRATION GUIDE:

N/A

EXAMPLE USAGE:

N/A


react/DeleteIcon/DeleteIcon.svg
```diff
@@ -1,18 +1 @@
...
-	<path d="M5.962,5.254c-0.414,0-0.749,0.336-0.749,0.75V17.99c0,0.414,0.335,0.75,0.749,0.75s0.75-0.336,0.75-0.75V6.004
-		C6.712,5.59,6.376,5.254,5.962,5.254z"/>
-	<path d="M8.965,5.254c-0.414,0-0.75,0.336-0.75,0.75V17.99c0,0.414,0.336,0.75,0.75,0.75s0.75-0.336,0.75-0.75V6.004
-		C9.715,5.59,9.379,5.254,8.965,5.254z"/>
-	<path d="M11.963,5.254c-0.414,0-0.749,0.336-0.749,0.75V17.99c0,0.414,0.335,0.75,0.749,0.75s0.75-0.336,0.75-0.75V6.004
-		C12.713,5.59,12.377,5.254,11.963,5.254z"/>
-</g>
-</svg>
+<svg xmlns="http://www.w3.org/2000/svg" width="18" height="21.009"><path d="M17.25 2.965h-3.84V.749A.75.75 0 0 0 12.66 0H5.223a.75.75 0 0 0-.75.749v2.216H.75a.75.75 0 0 0 0 1.499h.694V20.26a.75.75 0 0 0 .75.749h13.521a.75.75 0 0 0 .749-.749V4.464h.785a.75.75 0 1 0 .001-1.499zM5.972 1.499h5.939v1.466H5.972V1.499zm8.994 18.011H2.943V4.487h12.023V19.51z"/><path d="M5.962 5.254a.75.75 0 0 0-.749.75V17.99a.75.75 0 1 0 1.499 0V6.004a.75.75 0 0 0-.75-.75zM8.965 5.254a.75.75 0 0 0-.75.75V17.99a.75.75 0 0 0 1.5 0V6.004a.75.75 0 0 0-.75-.75zM11.963 5.254a.75.75 0 0 0-.749.75V17.99a.75.75 0 1 0 1.499 0V6.004a.75.75 0 0 0-.75-.75z"/></svg>
```
georgespyropoulos: Has anything changed in this SVG? Is this just changes to whitespace?

georgespyropoulos: 🤔... not sure if we should version svgs like this given a lot of styleguide consumers would be using svgo as part of their webpack config. Any other opinions on this?

tgdev: @georgespyropoulos RE: Breaking change. The result of this change is as follows:

- Plus icon: 20px x 20px (before) vs. 16px x 16px (after)
- DeleteIcon: 18px x 21px (before) vs. 16px x 16px (after)

So would if be fair to say that it's only a breaking change if consumers have applied CSS rules to adjust the size as this update might caused visual differences?

Happy to be advised on this one as I'm unfamiliar with what constitutes a breaking change in this instance

kosanna: Icons are styles with 16px by default with the Icon component, so there will be no difference for consumers @georgespyropoulos. We should've them compressed I guess from the start I guess

kosanna: Interestingly changes the icon 🤔 Maybe will agree to revert the compression for now

mengtzu: You might need to hand edit the svgo output a little.  We've converted all the icons in `seek-asia-style-guide` but you probably want to go to `viewBox`, see if there's any weird fill rules or masks or whatever that came through.

I don't *really* expect to see that on style guide icons rather than random icons pulled in from specs, but that's the process I've been using this week to convert unoptimised source icons.

tgdev:
Changes to SVG files were reverted due to compression weirdness.

react/DeleteIcon/DeleteIcon.svg
tgdev: This has been reverted along with the compression change to `PlusIcon.svg`

PR #440, opened 2018-02-26
refactor(Icons): Organise icons code in a single place

kosanna:
Reason for change:

Currently we have duplicated components for all of the icons. 
Sometimes they are different, aka some accept className prop, some of them don't etc.
Recently 2 icons were refactored to be the copycats of others due to inconsistencies (https://github.com/seek-oss/seek-style-guide/pull/438)

Other issue: currently it's hard to understand the scope of all of the icons because they are hanging out in the root folder as all other components of a different nature.

Thus we could not easily apply any icon strategy to all of them easily.

This PR adds icons generator and puts all of the icons in a dedicated folder.
Thus all of the icons are not separate components, but just a single codebase.
Hence no need to go and refactor each icon in case we need to apply some change to all of them in the future.

Also kills some of the dupe styles.

P.S.: PR is a lil drafti'sh, will add tests.
Happy to rethink on the folder structure/naming etc. Feels like it might be a private thing.
Direction prop is slightly confusing cause is used only for chevron...perhaps will just write a custom proptype checker.

Comments, complains are welcome ❤️


kosanna:
Closing for now

PR #439, opened 2018-02-25
feat(theme): Add new blues and use internally, darken sk-black

markdalgleish:
This adds the following colours:
- `@sk-blue-darker`
- `@sk-blue-dark`
- `@sk-blue-light`
- `@sk-blue-lighter`

The new blues have been applied to links, buttons and form accents.

For accessibility reasons, `@sk-black` has also been darkened slightly.

While revisiting form elements, some visual styling has been tweaked:

- Button shadows are now inset, ensuring they match the button hue
- SlideToggle animations have been polished up a bit

PR #441, opened 2018-02-27
chore(theme): Use palette blues rather than incorrect use of elements

michaeltaranto:
Since [the introduction of the blue variants](https://github.com/seek-oss/seek-style-guide/pull/439), we can now more semantically reference shades of directly from the palette as opposed to using `sk-link` out of context of an actual link.

Also migrating away from using `sk-highlight` as this will be deprecated in the future in favour of using the palette or element colours specifically.

### DIFF
--- a/react/Header/UserAccountMenu/UserAccountMenu.less
+++ b/react/Header/UserAccountMenu/UserAccountMenu.less
@@ -76,10 +76,10 @@
   }
 }
 
-.icon.jobSearch { color: @sk-highlight; }
+.icon.jobSearch { color: @sk-blue-lighter; }
 .icon.profile { color: @sk-purple; }
 .icon.saveJobs { color: @sk-pink; }
-.icon.saveSearches { color: @sk-highlight; }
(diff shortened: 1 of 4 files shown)


react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -76,10 +76,10 @@
   }
 }
 
-.icon.jobSearch { color: @sk-highlight; }
+.icon.jobSearch { color: @sk-blue-lighter; }
```
markdalgleish: This actually changes the colour—should this be a release?

michaeltaranto: I was happy not to trigger it, but if you want it to roll out we can. Thoughts?

markdalgleish: I was going to suggest releasing it, but I'm probably not so keen on the release noise. We can leave it.

PR #442, opened 2018-03-02
feat(Columns): New reverse prop added

tgdev:
New reverse prop allows consumers to flip the order of the children of the Columns component

When using the Columns component to replicate a fairly common UI pattern of Image on left with content on right, the pattern also offer the reverse of Content on left and Image on right.

Seeing as the Columns component does not accept a classNames prop to control this in the form of a CSS override, after discussing the problem with fellow contributors, the better option seemed to add a new prop that, if passed in, flips the ordering of the items within the columns component.

### Changes
- added new `reverse` prop to Columns component
- updated existing styles to enable reversal of columns' children
- added unit test for reverse prop
- updated demo to include missing prop options with new prop option (flexible, tight & reverse)
- added missing default props to `Columns` component

EXAMPLE USAGE:

```js
<Columns reverse>
    <Card>Item 1</Card>
    <Card>Item 2</Card>
</Columns>
```

### DIFF
--- a/react/Columns/Columns.less
+++ b/react/Columns/Columns.less
@@ -3,6 +3,9 @@
 .columns {
   @media @desktop {
     display: flex;
+    &.columns_reverse {
+      flex-direction: row-reverse;
+    }
   }
 }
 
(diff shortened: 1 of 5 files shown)


react/Columns/Columns.less
```diff
@@ -20,11 +23,17 @@
     .columns_tight & {
       padding: 0 (@gutter-width / 4);
     }
-    &:first-child {
+    .columns:not(.columns_reverse) &:first-child {
```
jahredhope: Could the duplicate statements be combined? Using `,` to combine them.

tgdev: Done

PR #444, opened 2018-03-14
fix(Checkbox): Avoid shrinking checkbox in limited space

jahredhope:
When the checkbox text becomes too long for the space available it may attempt to wrap.
This causes the checkbox to shrink as it is a peer to the text content.

This change stops the checkbox 'box' area from shrinking or growing with the space available.

In addition this change will ensure the outer hit-box that handles the on-click will always encompass the text inside it, currently if text overflows it wont increase the hit-box.

### DIFF
--- a/react/Autosuggest/__snapshots__/Autosuggest.test.js.snap
+++ b/react/Autosuggest/__snapshots__/Autosuggest.test.js.snap
@@ -28,7 +28,6 @@ exports[`Autosuggest should render with label 1`] = `
       <input
         aria-autocomplete="list"
         aria-expanded="false"
-        aria-haspopup="false"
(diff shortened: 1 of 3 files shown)


react/Autosuggest/__snapshots__/Autosuggest.test.js.snap
```diff
@@ -139,7 +137,6 @@ exports[`Autosuggest should render with simple props 1`] = `
       <input
         aria-autocomplete="list"
         aria-expanded="false"
-        aria-haspopup="false"
```
jahredhope: These snapshot changes only occurred when I updated to the latest npm flexible dependency versions. They don't appear to be related to the change.

PR #445, opened 2018-03-16
feat(BulletList): Add BulletList and Bullet components

jahredhope:
# Change
A Bullet point list is a common text layout pattern used across many sites.
This change provides access to a basic implementation of a bullet point list.

### Indent
The default implementation includes a 15px indent.

### Bottom margin
The default implementation provides a small amount of bottom margin that is common amongst the majority of bullet point lists implementations. An optional property to remove this bottom margin may be required in the future.

### SlideToggle
A missing required property in <SlideToggle /> was causing console errors during development and has been added.

# Example usage

```js
    <BulletPointList>
      <Bullet>Bullet point 1</Bullet>
      <Bullet>Bullet point 2</Bullet>
      <Bullet>Bullet point 3</Bullet>
    </BulletPointList>
```
screen shot 2018-03-16 at 5 16 07 pm


markdalgleish:
Just a few thoughts.

- Do you think we could shorten the component names to `BulletList` and `Bullet`?
- I wonder how this would play out with different font sizes? Might be worth experimenting to make sure the API can handle it. I can imagine us needing larger bullet lists in more promotional material.
- I'm not sure whether it's a good idea to abstract away the `Bullet` components or not. Did you have any thoughts on this?

michaeltaranto:
I had similar thoughts around a `List` component with some sort of `type` interface. Allows us to centralise any type of list into here.

jahredhope:
Shorten naming:
I preferred the verbose name as I think it's more findable.

Abstract away Bullet:
Happy to pull it in if it's not going to be useful.

Font-sizes:
Currently not supported in this implementation as I wasn't able to find an example where we have non-body font sizes with bullet points. Happy to add support now, but would prefer to wait for an example need.

markdalgleish:
Regarding font sizes, I'm more interested in it as a thought experiment to help guide us towards a better API—one that plays better with the other typographical components. I'm concerned the current API is too inflexible, being limited to standard type.

react/index.js
```diff
@@ -22,6 +22,8 @@ export { default as Secondary } from './Secondary/Secondary';
 export { default as Strong } from './Strong/Strong';
 export { default as Text } from './Text/Text';
 export { default as TextLink } from './TextLink/TextLink';
+export { default as BulletPoint } from './BulletPoint/BulletPoint';
```
michaeltaranto: What scenario should we need this exposed? If it's not required at the moment, i would limit increasing the api footprint to just the list for now.

jahredhope:
Made some updates based on feedback:
- Renamed BulletPointList to BulletPoint
- Removed the double applying of baseline transform (due to standardText mixin)
- Bullet was initially removed, but with the changes to API to support the above it's been added back in

Keen to get feedback on this new API

react/BulletList/BulletList.sketch.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { BulletList, Bullet } from 'seek-style-guide/react';
+export const symbols = {
```
markdalgleish: Thanks for adding some Sketch symbols! 💎👏👏👏

react/BulletList/BulletList.sketch.js
```diff
@@ -0,0 +1,25 @@
...
+  'BulletPoint/Small': (
```
markdalgleish: These symbol names need to be updated to match the new component name.

react/BulletList/BulletList.test.js
```diff
@@ -0,0 +1,18 @@
...
+
+import Text from '../Text/Text';
+
+import BulletList from './BulletList';
+
+describe('BulletList', () => {
+  it('renders with standard text', () => {
+    const wrapper = shallow(
+      <BulletList>
+        <Text>Foo</Text>
```
markdalgleish: Does this test need to be updated to use `Bullet` components?

react/SlideToggle/SlideToggle.sketch.js
```diff
@@ -4,6 +4,6 @@ import SlideToggle from './SlideToggle';
 const noop = () => {};
 
 export const symbols = {
-  'Slide Toggle/Unchecked': <SlideToggle label="Slide toggle" onChange={noop} checked={false} />,
-  'Slide Toggle/Checked': <SlideToggle label="Slide toggle" onChange={noop} checked={true} />
+  'Slide Toggle/Unchecked': <SlideToggle id="myToggle" label="Slide toggle" onChange={noop} checked={false} />,
+  'Slide Toggle/Checked': <SlideToggle id="myToggle" label="Slide toggle" onChange={noop} checked={true} />
```
markdalgleish: This should be a different ID?

react/Text/Text.less
```diff
@@ -59,3 +59,9 @@
 .raw {
   padding-bottom: 0;
 }
+
+.bullet {
+  display: list-item;
+  list-style-type: disc;
+  margin-left: 1.5 * @gutter-width;
+}
```
markdalgleish: So clean! 😍

markdalgleish:
Just a few relatively minor comments—this is looking really slick 👍

jahredhope:
Thanks @markdalgleish 
Updated to the above comments.

react/Text/Text.js
```diff
@@ -13,6 +13,8 @@ import stylesRegular from '../Regular/Regular.less';
 import withTextProps, { SizePropTypes } from '../private/withTextProps';
 
 const Text = ({
+  component,
```
michaeltaranto: Can we add `component` to the proptypes definition below (along with the default)

react/Text/Text.js
```diff
@@ -25,35 +27,40 @@ const Text = ({
...
-    className={classnames({
-      [styles.root]: true,
-      [className]: className,
-      [styles[size]]: size,
-      [styles.raw]: raw,
-      [styles.baseline]: baseline
-    })}>
-    <span
+}) => {
+  const Component = component || (bullet ? 'li' : 'span');
```
michaeltaranto: If we add `component` to the props list with a default of `span` this can be simplified to: `bullet ? 'li' : component`

react/Text/Text.less
```diff
@@ -59,3 +59,9 @@
 .raw {
   padding-bottom: 0;
 }
+
+.bullet {
+  display: list-item;
+  list-style-type: disc;
+  margin-left: 1.5 * @gutter-width;
```
michaeltaranto: I dont like what im about to suggest, but, if we pull this out to a mixin we can use it in places where we cannot use the style guide components themselves, not dissimilar to the text mixins. Thoughts?

react/Text/Text.js
jahredhope: This would change the behaviour.
The current behaviour priorises a consumer defined component over a bullet's defaut component.
This was chosen to keep things be a little less magical for the consumer, sometimes honouring a custom component and sometimes not may lead to confusion while debugging.

react/Text/Text.less
markdalgleish: YAGNI (IMO) 😄

michaeltaranto: We already do, i'll take it offline.

jahredhope: I'd rather avoid supporting consumers using mixins without an known consumer.
We could pull this out to a mixin if and when it is required.

Thats just a preference, if you want to support mixin consumers from day one I will.

michaeltaranto: There's already a need for it, it's not hypothetical. Check slack 😄

react/Text/Text.js
jahredhope: Added component proptype definition. Thanks

PR #447, opened 2018-03-20
feat(Radio): Add Radio button component

dborysiewicz:
## New Feature

We have a need for a styled Radio button and this is just bringing it into the style guide for all to use , Api followed is similar to the Checkbox component 🔘

## Usage

```js
<Radio
 id="uploadCoverLetter"
 inputProps={{
  name: "coverLetter",
  value: "upload",
  checked: coverLetterOption === 'upload',
  onChange: this.handleCoverLetterChange
 }}
 label="Upload a cover letter"
/>
```


react/Radio/Radio.less
```diff
@@ -0,0 +1,52 @@
...
+  margin-right: @gutter-width / 2;
+
+  .input:not([readonly]):not([disabled]):focus + .label & {
+    .focus();
+    border-radius: 100%;
+  }
+}
+
+.circle {
+  fill: @sk-blue-lighter;
```
markdalgleish: This should be `@sk-form-accent`

react/Radio/Radio.less
```diff
@@ -0,0 +1,52 @@
...
+@circumference: @row-height * 5;
+
+.root {
+  display: inline-block;
+}
+
+.input {
+  opacity: 0;
+  position: absolute;
+  width: @circumference;
```
markdalgleish: Currently, Radio renders as only `30px` high, which is below our `48px` touchable size. It also means that Radio and Checkbox are different heights. I'm guessing we'd need to fix this up?

react/Radio/Radio.less
michaeltaranto: I believe this should be using `@sk-form-accent` instead

react/Radio/Radio.js
```diff
@@ -0,0 +1,81 @@
...
+
+    return (
+      <label {...allLabelProps}>
+        <svg
+          className={styles.svg}
+          viewBox="0 0 200 200"
+          xmlns="http://www.w3.org/2000/svg"
+          focusable="false" >
+          <circle className={styles.circle} cx="100" cy="100" r="100" />
+        </svg>
```
michaeltaranto: Curious about why we have used an inline svg versus a background using css?

react/Radio/Radio.sketch.js
```diff
@@ -0,0 +1,24 @@
...
+import styles from './Radio.demo.less';
+
+const Container = ({ children }) => <div className={styles.root}>{ children }</div>;
+Container.propTypes = {
+  children: PropTypes.node.isRequired
+};
+
+const noop = () => {};
+
+export const symbols = {
```
markdalgleish: Thanks for adding Sketch symbols! 💎🙌🏻

react/Radio/Radio.demo.less
```diff
@@ -0,0 +1,3 @@
+.root > :global(*) {
+  min-width: 150px; // :(
```
markdalgleish: Obviously don't want to block this PR, but this component is continuing the trend of form components being a bit awkward to use without custom styling—which is not your fault. It'd be great to clean up our form components at some point.

react/Radio/Radio.js
dborysiewicz: This is/how it was being done in profile as really I'm just duplicating what they did there and pulling it into the style guide, if it's a issue I can change it?

michaeltaranto: Not an "issue", just not consistent with what i've seen here and felt solveable with fewer classes in css. I'll leave it up to you.

react/Radio/Radio.less
```diff
@@ -11,7 +11,7 @@
   width: @circumference;
   height: @circumference;
   border: none;
-  margin: 0;
+  margin: @row-height 0 0;
```
michaeltaranto: Should it just be `margin-top` now?

michaeltaranto:
I would be good if there were a way to deselect the radio in the demo page. Appreciate it's not a toggle, but you can't really play with the component without this. Maybe a second radio button in the group even?

react/Radio/Radio.less
dborysiewicz: I am overriding any browser defaults, in this case Chrome is `margin: 3px 0.5ex;`

michaeltaranto:
Also it'd be great to breathe some life into the this component with some subtle animation, i'm having a play with it now.

react/Radio/Radio.less
dborysiewicz: Added padding which addresses this 👍

michaeltaranto: Arent all the styles redundant as its an invisible checkbox?

dborysiewicz: OSX voiceover overrides the focus state and uses the hidden input size and position so this just makes it a bit nicer for that use case.

theme/animations/animations.less
```diff
@@ -0,0 +1,2 @@
+@spring-transition: transform 0.2s cubic-bezier(0.02, 1.505, 0.745, 1.235);
```
markdalgleish: 👏

react/Radio/Radio.less
michaeltaranto: Amazing! Love it 👏

react/Radio/Radio.less
```diff
@@ -0,0 +1,73 @@
...
+  border: none;
+  margin: @row-height 0 0;
+}
+
+.label {
+  display: flex;
+  cursor: pointer;
+  position: relative;
+  align-items: center;
+  padding: (@row-height * 1.5) 0;
```
michaeltaranto: Rather than use partial row values, I would suggest setting the height and vertically aligning to the center with flex box as partial row values are asking for trouble down the track.

PR #448, opened 2018-03-25
fix(MonthPicker): Accessibility

dborysiewicz:
Add field set and legend along with aria labelling so that user's of voice over finds the MonthPicker more intuitive to use.

### DIFF
--- a/react/MonthPicker/MonthPicker.js
+++ b/react/MonthPicker/MonthPicker.js
@@ -113,13 +113,17 @@ export default class MonthPicker extends Component {
 
     return (
       <div className={classNames}>
-        <FieldLabel
-          {...{ id, label, labelProps, secondaryLabel, tertiaryLabel }}
-        />
-        {this.renderInput()}
-        <FieldMessage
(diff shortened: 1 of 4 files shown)


react/MonthPicker/MonthPicker.js
```diff
@@ -113,13 +113,17 @@ export default class MonthPicker extends Component {
...
     return (
       <div className={classNames}>
-        <FieldLabel
-          {...{ id, label, labelProps, secondaryLabel, tertiaryLabel }}
-        />
-        {this.renderInput()}
-        <FieldMessage
-          {...{ invalid, help, helpProps, valid, message, messageProps }}
-        />
+        <fieldset>
```
michaeltaranto: Rather than nest the `fieldset` in the `div` you can change the `div` to a `fieldset` for the same result right?

dborysiewicz: Having a re-think about this one I'm not a fan as nested fieldset has problems, I'll close for now and work on a better solution.

PR #449, opened 2018-03-26
fix(SlideToggle): Reorder html for VoiceOver & style to match other components

dborysiewicz:
The current implementation of the input nested inside the label breaks accessibility with voice over. This is just reordering the markup so the input is now outside the label along with the css changes so it looks the same. I have also made the hidden input match the visual component so osx voice over focus state is a nicer.


react/SlideToggle/SlideToggle.js
```diff
@@ -49,29 +49,31 @@ export default class SlideToggle extends Component {
...
+        ref={this.storeInputRef}
+        type="checkbox"
+        id={id}
+        aria-label={label}
+        className={inputStyles}
+        checked={checked}
+        onChange={onChange}
+      />
+        <label htmlFor={id}>
+          <div className={styles.switch}>
```
michaeltaranto: This `div` can go now as we can use the `label` as the parent element (trying to minimise the `div` soup 🍜 )

react/SlideToggle/SlideToggle.less
```diff
@@ -14,43 +18,45 @@
...
   position: absolute;
-  left: 10px;
-  top: 5px;
+  left: 1px;
   margin: 0;
+  width: @toggleWidth;
+  height: @toggleHeight;
 
-  &:focus ~ .slider .slideButton::before {
+  &:focus ~ label .slider .slideButton::before {
```
michaeltaranto: If we [do this](https://github.com/seek-oss/seek-style-guide/pull/449/files#r176975817) we can move away from using the global element selectors

react/SlideToggle/SlideToggle.less
```diff
@@ -110,12 +115,12 @@
   opacity: 0;
   transition: 0.3s;
   position: relative;
-  top: 5px;
+  top: -4px;
```
michaeltaranto: This looks off to me, i would actually suggest removing the relative positioning and apply flex centering in the `icon` class above.

dborysiewicz: Hrmmm I'll give it ago, be right back 🚪

react/SlideToggle/SlideToggle.less
```diff
@@ -1,60 +1,59 @@
...
-  left: 10px;
-  top: 5px;
+  left: 1px;
   margin: 0;
+  width: @toggleWidth;
+  height: @toggleHeight;
+  cursor: pointer;
 
-  &:focus ~ .slider .slideButton::before {
+  &:focus ~ .switch .slider .slideButton::before {
```
michaeltaranto: The focus style has gone missing, you probably need to remove this `::before` selector to be consistent with the rest of this PR.

react/SlideToggle/SlideToggle.less
```diff
@@ -1,60 +1,48 @@
 @import (reference) "~seek-style-guide/theme";
 
+@toggle-height: (@row-height * 5);
+@toggle-slider-width: 48px;
+@toggle-slider-height: (@row-height * 4);
+@toggle-button-size: 30px;
+@toggle-button-border: 1px;
+@toggle-radius: (@row-height * 2.5);
```
georgespyropoulos: Perhaps this would read better as `(@toggle-height / 2)`?

dborysiewicz:
@georgespyropoulos 
1. I do see what you mean however I'm torn as currently Radio buttons do this but Checkboxes do not. I'm keen to make them all work the same so I'll see what @markdalgleish & @michaeltaranto think in regards to this way we are going to go.
2. There was a deliberate removal of the animation on the colour as it look a a bit weird when it was opening with it as it clashed with the button 'rolling' to reveal the background.

michaeltaranto:
Handful of fixes for lesser browsers such as IE11 and Safari

PR #453, opened 2018-03-29
fix(MonthPicker): Improve labels for screen readers, make full width

dborysiewicz:
BREAKING CHANGE:
MonthPicker now has a default width: 100%, this is inline with the default options of other inputs.
CustomMonthPicker is now correctly styled with the correct 10px padding between inputs and 50% widths.

FIX: 
Add hidden labels to aid with VoiceOver.


react/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -133,46 +136,60 @@ export default class CustomMonthPicker extends Component {
 
   render() {
     const { value, className, valid, id } = this.props;
+    // eslint-disable-next-line react/prop-types
+    const { label, labelProps, secondaryLabel, tertiaryLabel } = this.props;
```
michaeltaranto: Any reason you are destructuring these props separately to the line above? That coupled with the eslint ignore feels like something isnt playing nice

react/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -133,46 +136,60 @@ export default class CustomMonthPicker extends Component {
...
-            onChange: this.handleYearChange,
-            value: yearValue,
-            className: styles.dropdownInput,
-            ref: this.storeYearReference
-          }}
-        />
+        <div className={styles.monthWrapper}>
+          <FieldLabel
+            {...{ id: `${id}-month`, label: label ? <span>{label} <ScreenReaderOnly>Month</ScreenReaderOnly></span> : <ScreenReaderOnly>Month</ScreenReaderOnly>, labelProps, secondaryLabel, tertiaryLabel }}
+          />
```
michaeltaranto: For legibility it would be good to pull these out to internal rendering functions, and maybe shared between both fields if the code remains clear

react/MonthPicker/NativeMonthPicker/NativeMonthPicker.js
```diff
@@ -81,18 +86,22 @@ export default class NativeMonthPicker extends Component {
...
+          {...{ id: `${id}`, label: label ? <span>{label} <ScreenReaderOnly>Month Year</ScreenReaderOnly></span> : <ScreenReaderOnly>Month Year</ScreenReaderOnly>, labelProps, secondaryLabel, tertiaryLabel }}
+        />
         <ChevronIcon
           className={styles.chevron}
           svgClassName={styles.chevronSvg}
           direction="down"
         />
         <input
-          {...(id ? { id } : {})}
+          id={id}
```
michaeltaranto: Is `id` now a required property?

react/MonthPicker/NativeMonthPicker/NativeMonthPicker.js
```diff
@@ -81,18 +86,22 @@ export default class NativeMonthPicker extends Component {
 
     return (
       <div className={rootClasses}>
+        <FieldLabel
+          {...{ id: `${id}`, label: label ? <span>{label} <ScreenReaderOnly>Month Year</ScreenReaderOnly></span> : <ScreenReaderOnly>Month Year</ScreenReaderOnly>, labelProps, secondaryLabel, tertiaryLabel }}
```
michaeltaranto: No need to use a template string for `id` here. And as above, it'd be great to format this a little friendlier

react/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
dborysiewicz: This is how we manage having a central propsType check inside FieldMessage for these props - this is done is all other Components which use FieldMessage.

react/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
dborysiewicz: It's different between each field as we only want to pass labelProps, secondaryLabel, tertiaryLabel to the visible label but can pull out to internal functions if you think it's worth it 👍

react/MonthPicker/NativeMonthPicker/NativeMonthPicker.js
dborysiewicz: Arrrr it would be nice but not sure I want to make that change - will revert.

markdalgleish:
Secondary and tertiary labels are broken:

screen shot 2018-04-04 at 8 54 21 am

dborysiewicz:
@markdalgleish only broken as the space is too small to support that length of text and the options in the demo are coming from the imported `fieldMessageOptions`. Easiest solution would be to just remove the demo of secondary & tertiary labels for this component?

react/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -34,16 +35,14 @@ export default class CustomMonthPicker extends Component {
...
-    className: PropTypes.string,
     id: PropTypes.string,
     minYear: PropTypes.number.isRequired,
     maxYear: PropTypes.number.isRequired,
     ascendingYears: PropTypes.bool.isRequired
   };
 
   static defaultProps = {
-    value: {},
-    className: ''
```
markdalgleish: This is technically a breaking change. Why is this being removed?

markdalgleish: Sorry, I missed that this change isn't actually part of the public API. All good 👍

PR #455, opened 2018-03-29
docs(site): Inline critical CSS with purify-css

markdalgleish:
giphy-downsized

EDIT: To see this in action, view source on the preview deployment: https://seek-style-guide--61f35957bc11ee9b6bd3ed05310de503dc73b4bd.surge.sh


docs/src/server-render.js
```diff
@@ -1,17 +1,25 @@
...
       <App />
     </StaticRouter>
   );
 
-  return template({ html, baseHref });
-};
+  const purifyOptions = {
+    minify: true,
+    whitelist: ['*:not*'] // https://github.com/purifycss/purifycss/issues/161
+  };
```
kosanna: Might we want to log unused selectors?

markdalgleish: The log would be pretty noisy—I don't imagine it would help much.

kosanna: Might help in dev 🤔 But yeah, happy with either

PR #456, opened 2018-04-03
fix(Radio): Fix dropped click events on elements inside label in IE11

markdalgleish:
Right now, in IE11, clicks on the radio button often fail. Turns out it's because the click event isn't bubbling up to the label correctly. This PR fixes it by disabling pointer events on all children of the label, ensuring that clicks are always registered against the label directly.

### DIFF
--- a/docs/webpack.dev.config.js
+++ b/docs/webpack.dev.config.js
@@ -16,6 +16,7 @@ const appPaths = [
 
 const config = decorateClientConfig({
   entry: [
+    'core-js/modules/es6.symbol',
     'react-hot-loader/patch',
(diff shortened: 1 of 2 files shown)


docs/webpack.dev.config.js
```diff
@@ -16,6 +16,7 @@ const appPaths = [
 
 const config = decorateClientConfig({
   entry: [
+    'core-js/modules/es6.symbol',
```
markdalgleish: This was required for the site works in IE. It needed to be added before `react-hot-loader`, which is why it's in the `entries` array rather than `client-render.js` with the other polyfills.

PR #461, opened 2018-04-09
fix(FieldMessage): Make messages describe its relevant elements for VoiceOver

dborysiewicz:
Pass/create known id's for FieldMessage and add aria-describedby to form elements so that messages are correctly read out for voice over.


react/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -170,7 +171,8 @@ export default class CustomMonthPicker extends Component {
               onBlur: this.handleBlur,
               onChange: this.handleMonthChange,
               value: monthValue,
-              ref: this.storeMonthReference
+              ref: this.storeMonthReference,
+              'aria-describedby': fieldMessageId // Is is important as these fields are using a shared FieldMessage
```
kosanna: Could we then just define it as a shared prop object before the return?

kosanna:
```
const sharedInputProps = {
    onBlur: this.handleBlur,
    'aria-describedby': fieldMessageId
};
```

Then we don't need to worry about someone unintentionally breaking it

PR #466, opened 2018-04-13
feat(Pill): Pill component

georgespyropoulos:
Creating a pill component for the Styleguide based on pills used in Profile, Incubator and Talent search.

Credit for this work goes to @tamoore, we paired on this yesterday.

Demo page http://seek-style-guide--66a8928d97f5d0428635b7657def1ad25f67f343.surge.sh/pill

screen shot 2018-04-13 at 10 36 00 am
screen shot 2018-04-13 at 10 36 08 am

EXAMPLE USAGE:

```js
<Pill
  text="I am a pill"
  onClose={() => {}}
/>
```


react/Pill/Pill.sketch.js
```diff
@@ -0,0 +1,18 @@
...
+    <SketchFieldContainer>
+      <Pill text="Pill text" />
+    </SketchFieldContainer>
+  ),
+  'Pill/Interactive': (
+    <SketchFieldContainer>
+      <Pill text="Pill text" onClose={noop} />
+    </SketchFieldContainer>
+  )
+};
```
markdalgleish: 👏👏👏👏👏

tamoore:
@georgespyropoulos you're too kind. I just pointed at the screen a few times, you did most of the work.

georgespyropoulos:
Demo page http://seek-style-guide--59ae9bdcb30f953505d9214ee4593f03eeba9a1d.surge.sh/pill

markdalgleish:
I think the focus state needs some work, and we probably need hover/active states too. What do you think? Should we do this now or later?

georgespyropoulos:
The code was mostly taken from the pill in profile and it doesn't have any hover or active states either. That's not to say it doesn't need those but if we go out with this, we're not making anything worse. But it shouldn't take long to sort out those states so I am happy to do it now

tamoore:
I'm not sure I agree that front loading the component with as yet defined states is a good idea. Isn't that the point of iteration?

georgespyropoulos:
Yeah I guess that's the question - what do those states look like? Do we want designer input on that?

markdalgleish:
I feel like the middle ground is to spike it out with a few people and see what we can do in a short period of time. I agree we should iterate, but this is more about setting a baseline level of quality for style guide components.

georgespyropoulos:
Surge link with the latest design changes seek-style-guide--66a8928d97f5d0428635b7657def1ad25f67f343.surge.sh

react/Critical/Critical.less
```diff
@@ -1,5 +1,5 @@
-@import (reference) "~seek-style-guide/theme";
+@import (reference) '~seek-style-guide/theme';
```
markdalgleish: Is this another Prettier update? Do we need to bite the bullet and add Prettier to this repo? 😄

georgespyropoulos: Oh yeah it is - I missed that change when I pushed it up. I have it turned on in my editor. I would love to have prettier on in this repo

mengtzu: Please do bite that bullet.

react/Pill/Pill.less
```diff
@@ -0,0 +1,87 @@
...
+  height: @staticTagHeight;
+  line-height: @staticTagHeight;
+  padding: 0 15px;
+  margin: @row-height 10px @row-height 0;
+}
+
+.interactivePill {
+  align-items: center;
+  display: inline-flex;
+  font-size: 18px;
```
michaeltaranto: We shouldn't have explicit font sizes specified like this, happy to pair with you if there is a problem that you're working around

react/Pill/Pill.less
```diff
@@ -0,0 +1,87 @@
...
+  background: @sk-gray-light;
+  border-radius: 24px;
```
michaeltaranto: Should this be more dynamic? Maybe the same value as `@removeCircleDiameter`? or is that a coincidence?

react/Pill/Pill.less
```diff
@@ -0,0 +1,87 @@
...
+  margin: @row-height 10px @row-height 0;
+}
+
+.removeButton {
+  margin: 0;
+  padding: 0;
+  height: @interactiveTagHeight - 2;
+  width: @interactiveTagHeight - 2;
+  color: @sk-white;
+  border-radius: 0;
```
michaeltaranto: This is the default right? Im not sure its needed.

react/Pill/Pill.less
```diff
@@ -0,0 +1,87 @@
...
+  width: @removeCircleDiameter;
+  border-radius: @removeCircleDiameter / 2;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  transition: @spring-transition;
+}
+
+.removeIcon {
+  height: 12px;
```
michaeltaranto: Curious if this needs to be exact or a percentage of the container (`.removeCircle`) as that has an enforced width

theme/mixins/focus.less
```diff
@@ -1,7 +1,7 @@
-.focus() {
+.focus(@focus-color: @sk-focus) {
```
michaeltaranto: This parameter doesnt appear to be used, did it become unnecessary after a refactor?

react/Pill/Pill.less
```diff
@@ -0,0 +1,87 @@
...
+  .rawText(@interaction-type-scale);
+
+  align-items: center;
+  display: inline-flex;
+  color: @sk-charcoal;
+  background: @sk-gray-light;
+  border-radius: @staticTagHeight / 2;
+  height: @staticTagHeight;
+  line-height: @staticTagHeight;
+  padding: 0 15px;
```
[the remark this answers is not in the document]
georgespyropoulos: I didn't know we had that - I'll use it

react/Pill/Pill.less
georgespyropoulos: Turns out that's not needed at all, it was left over from some CSS that was copy / pasted from profile. `<Text>` is controlling the font size

react/Pill/Pill.less
georgespyropoulos: Good catch, the radius should be half the height of the interactive pill so I've changed it to `@interactiveTagHeight / 2`

react/Pill/Pill.less
georgespyropoulos: Good catch, it's not needed. But in the process of double checking that I stumbled across this https://github.com/necolas/normalize.css/issues/706

react/Pill/Pill.less
georgespyropoulos: It should be half the height of `.removeCircle` so I've changed it to `@removeCircleDiameter / 2`

theme/mixins/focus.less
georgespyropoulos: Yes it became unnecessary after a refactor. I decided to leave it in because it seems like something we will need at some point. But I'm happy to take it out - it's easy to add back in the future

react/Pill/Pill.less
michaeltaranto: Interesting, we could always add that into the reset sheet if needed

tamoore:
Is there any reason why this is still in review? It would be awesome to start using it. Could someone please give this a second approval?

PR #467, opened 2018-04-13
feat(StarIcon): Add 'half' prop, larger stars for Rating

vgoulko:
This change will update the Stars size (Empty, Halfull, Full) so they can look better, e.g. on Company Reviews pages.


react/StarIcon/StarIcon.less
```diff
@@ -9,3 +12,36 @@
   50% { transform: scale(1.1); }
   100% { transform: scale(1); }
 }
+
+.svg {
+  position: relative;
+  top: 0.05em;
+}
+
+.heroSvg {
```
michaeltaranto: It feels like we need an interface to set the `icon-scale` from the component rather than having to duplicate/override all these mixins. Is that something you looked into already?

react/StarIcon/StarIcon.less
```diff
@@ -9,3 +12,36 @@
   50% { transform: scale(1.1); }
   100% { transform: scale(1); }
 }
+
+.svg {
+  position: relative;
+  top: 0.05em;
```
michaeltaranto: Does this position work for non-rating usages of the `StarIcon`?

react/StarIcon/StarIcon.less
vgoulko: We were trying to resize the stars without changing the size of other icons and make it uncoupled.

markdalgleish: For me, I think the main thing was to make it really explicit so it's easy for people to make tweaks to individual sizes, if necessary.

react/StarIcon/StarIcon.less
vgoulko: Yes, it does, check it out:
Stars in playground

react/StarIcon/StarIcon.less
markdalgleish: I'm open to the mixin idea, I'm just a little concerned that it hides too much away and makes it hard to adjust the output.

react/StarIcon/StarIcon.less
markdalgleish: It's been adjusted to sit better alongside text in general, optically sitting on the baseline, so that's the assumption. Might be worth trying it out on your app if you're concerned.

michaeltaranto:
The `StarIcon` seems to be broken on [the preview site](http://seek-style-guide--61ecc5d563de44cd8223b619ad548d2b87e97003.surge.sh/)
screen shot 2018-04-16 at 9 45 21 am

react/StarIcon/StarIcon.less
michaeltaranto: I believe this is a breaking change, as consumers are likely to already have position adjustments that now need to be updated/removed. Also looking at the playground before and after, it's off grid after the icons and ratings sections.

markdalgleish:
@michaeltaranto just paired with @vgoulko to update the PR—we've reverted the StarIcon size changes and hidden them behind a private `_large` prop for now, which only `Rating` uses. Long term, I think we need to look at icon sizing across the board, but this fixes the immediate issue. Let me know what you think.

react/StarIcon/StarIcon.less
```diff
@@ -9,3 +19,59 @@
...
 }
+
+._large {
+  .svg {
+    position: relative;
+    top: 0.05em;
+  }
+
+  &.hero {
+    .iconHeight(@hero-type-row-span, @hero-type-row-span-mobile);
```
markdalgleish: @michaeltaranto just a heads up, we had to explicitly set the height of the icon container so that it follows the grid, rather than relying on child nodes. Ideally, this would happen in `Icon.less`, but wanted to limit the blast radius of this PR—we'll revisit this when we revisit icons as a whole.

PR #468, opened 2018-04-16
feat(AsidedLayout): Support left aligned aside

jahredhope:
# Reason for Change
We are looking to support layouts where an aside is provided to the left of the main content.

Support a left aside with the `position` prop.
Default behaviour remains right aligned.
Right aligned behaviour is unchanged.
'Reverse on mobile' behaviour is maintained for right and left aligned positions.

# How it works
Currently the position of the aside node moves depending on the `reverse` prop.
CSS is then used to reverse the display order using `flex-direction: row-reversed` on desktop.
This change reuses this same functionality but applies XOR behaviour to `position` and `reverse` as follows.

Props | Desktop Position | Mobile Position | DOM Position
------ | ---------------- | --------------- | --------------
| `position=right` | Right | After | After |
| `position=right` `reverse` | Right | Before | Before |
| `position=left` | Left | Before | Before |
| `position=left` `reverse` | Left | After | After |

# Example usage

```js
<AsidedLayout position="left" renderAside={...} >
...
</AsidedLayout>
```

```js
<AsidedLayout position="right" renderAside={...} >
...
</AsidedLayout>
```

### DIFF
--- a/react/AsidedLayout/AsidedLayout.js
+++ b/react/AsidedLayout/AsidedLayout.js
@@ -3,13 +3,18 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
+const LEFT = 'left';
+const RIGHT = 'right';
+
 const defaultRenderAside = () => null;
 
-const conditionallyRenderAside = (condition, renderAside, classNameAside, size) => (
+const conditionallyRenderAside = (condition, renderAside, classNameAside, size, position) => (
   condition ?
     <div
       className={classnames({
         [classNameAside]: classNameAside,
+        [styles.leftAside]: position === LEFT,
+        [styles.rightAside]: position !== LEFT,
         [styles.aside]: true
       })}
       style={{ flexBasis: size }}>
@@ -18,7 +23,8 @@ const conditionallyRenderAside = (condition, renderAside, classNameAside, size) 
     null
 );
 
-export default function AsidedLayout({ className, children, renderAside = defaultRenderAside, classNameAside, size, reverse, ...restProps }) {
+export default function AsidedLayout({ className, children, position, renderAside = defaultRenderAside, classNameAside, size, reverse, ...restProps }) {
+  const renderAsideBeforeContent = position === LEFT ? !reverse : reverse;
   return (
     <div
       {...restProps}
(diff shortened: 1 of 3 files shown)


kosanna:
I feel that it might be more clear to do a breaking change but have desktopPosition and mobilePosition now, otherwise it's confusing with the reverse prop

react/AsidedLayout/AsidedLayout.js
```diff
@@ -3,13 +3,18 @@ import React from 'react';
...
+
 const defaultRenderAside = () => null;
 
-const conditionallyRenderAside = (condition, renderAside, classNameAside, size) => (
+const conditionallyRenderAside = (condition, renderAside, classNameAside, size, position) => (
   condition ?
     <div
       className={classnames({
         [classNameAside]: classNameAside,
+        [styles.leftAside]: position === LEFT,
```
kosanna: could just userd poisition here like [styles[`${position}Aside`]]

kosanna:
I'm also not quite sure that we actively use different actual dom positioning while displaying in reverse, so might be good to align too, thoughts?

react/AsidedLayout/AsidedLayout.js
jahredhope: I probably prefer the implemented solution as it's lint-able and correlates with the dom position behavior when it errors:
e.g. if the prop `lefft` was used React will show a console error but still render, and rendered output would be the same as if the prop was undefined.

jahredhope:
@kosanna yeah the dom re-positioning was a surprise to me, I'm not aware of the requirements behind doing it that way. It seems to work well in practice though.

kosanna:
@jahredhope I feel like it's a nice opportunity to simplify things. I think we might want to align dom positioning here as well, because you might not want accessibility tools to read out the side-content first. So to me it feels like what we should be aiming for is the following

Props | Desktop Position | Mobile Position | DOM Position
------ | ---------------- | --------------- | --------------
| `position=right` | Right | Below | After |
| `position=right` `reverse` | Right | Above | Before |
| `position=left` | Left | Below | After |
| `position=left` `reverse` | Left | Above | Before |

Really not into `reverse` prop naming, it is not descriptive enough to understand what it is actually doing

markdalgleish:
I feel like mobile order and screen reader order should be the same, since they're both defining which content should be first if content is arranged linearly.

markdalgleish:
In addition to the previous point, how about we alias `reverse` as `wrapReverse`?

kosanna:
What wrapReverse means and why that would be better? 🤔 Still kinda think of something like isOnTopOnMobile or whatever

jahredhope:
I'm happy bike-shed ideas to improve the way `reverse` prop works at another time.

react/AsidedLayout/AsidedLayout.less
```diff
@@ -1,4 +1,4 @@
-@import (reference) "~seek-style-guide/theme";
+@import (reference) '~seek-style-guide/theme';
```
markdalgleish: Why the change in quotes?

jahredhope: Ah... didn't realise we didn't `prettier` this repo. Will reverse it.

kosanna:
Ok, hands up. Bike-shed later

PR #469, opened 2018-04-16
fix(FormElements): Make focus state accessible

dborysiewicz:
Currently focus states on form elements are not accessible as it is only relying on a change of colour, this change make it so that focus also increases the border width.

screen shot 2018-04-16 at 2 10 03 pm


theme/mixins/focus.less
```diff
@@ -1,7 +1,7 @@
 .focus() {
   outline: none;
   z-index: 1;
-  box-shadow: 0 0 0 @field-border-width @sk-focus;
+  box-shadow: 0 0 0 2px fade(@sk-focus, 75%);
```
markdalgleish: Shouldn't we be changing `@sk-focus` directly, rather than changing the usage of it? i.e. `@sk-focus: fade(#1e90ff, 75%);`

PR #471, opened 2018-04-16
feat(Info): Add new component and prop for Text

dborysiewicz:
Inline with the other text short cut components such as `<Positive />` and `<Critical />` this adds `<Info />` also have added this as a prop to `<Text />`

EXAMPLE USAGE:

```js
This text is <Info>info</Info>
```

```js
This text is <Text info>info</Text>
```


markdalgleish:
Could you also add this to the typography guide? You can basically just copy and paste an existing example.

PR #474, opened 2018-04-19
feat(Rating): Add ratingComment prop

kosanna:
Add ability to append rating with custom text if rating is displayed with `ratingComment` prop which might be string/component.

```js
<Rating
  rating={5.0}
  description="overall from 86 employee reviews"
/>
```

screen shot 2018-04-19 at 3 01 11 pm


markdalgleish:
Could we drop the `rating` prefix from the prop name, since it's a bit redundant? Not sure if `comment` is a good name by itself, though.

react/Rating/Rating.js
```diff
@@ -61,7 +62,9 @@ const Rating = ({
           );
         })}
         {showTextRating &&
-          <span className={styles.textRating}>{rating.toFixed(1)}</span>
+          <span className={styles.textRating}>
```
shawn-dsz: I don't think it will be obvious that `textRating` is required in order to have the comment show up. We probably should avoid having a dependency here.

react/Rating/Rating.js
```diff
@@ -33,6 +33,7 @@ const getStar = (
 };
 
 const Rating = ({
+  ratingComment,
```
shawn-dsz: I feel `ratingComment` is superfluous considering we are in the context of `<Rating>` how about just `comment`?

PR #477, opened 2018-04-24
fix(Section): Change side padding breakpoint to be desktop one

kosanna:
Currently Section component has a breakpoint close to the gutters of the website where it drops the side padding from 30px to 20px, which makes desktop experience inconsistent across different device widths.

This PR changes the breakpoint of the gutter change to be simply mobile/desktop.

### DIFF
--- a/react/Critical/Critical.less
+++ b/react/Critical/Critical.less
@@ -1,5 +1,5 @@
 @import (reference) "~seek-style-guide/theme";
 
 .root {
-  color: @sk-pink;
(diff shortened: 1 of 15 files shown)


kosanna:
That's a good point, but I'd leave it as not a breaking change 🤔 cc @markdalgleish 👆 thoughts?

kosanna:
@dborysiewicz this changed has been proposed on the design group meeting, so all the designers are across

PR #478, opened 2018-04-30
docs(sandbox): Add sandbox page

markdalgleish:
screen shot 2018-04-30 at 2 29 29 pm

This adds a sandbox page to the documentation site with the following features:

- JSX code is parsed and rendered directly in the browser, allowing you to experiment with style guide code without downloading or installing anything.
- All style guide components are available without having to explicitly import anything.
- The preview pane consists of multiple iframes, to show how the code is rendered on different devices.
- The preview pane is updated on every keystroke (with a slight debounce).
- Code is parsed before rendering, to ensure that invalid code doesn't clear the preview pane.
- Syntax errors are displayed in the gutter, via a circle with a `title` attribute.
- To encourage experimentation, clicking on the syntax error icon in the gutter will revert the code to the last valid state.

Please note that functions and dynamic expressions aren't currently supported, which is definitely a *massive* omission. If this tool gets enough usage, I'll look to add support for these in the future.


docs/index.ejs
```diff
@@ -13,6 +13,7 @@
 <body>
   <div id="app"><%- html %></div>
   <link rel="stylesheet" href="app.css" />
+  <div data-style-loader></div>
```
markdalgleish: Now that we're lazy loading CSS, we need to ensure that styles are injected into the DOM *after* the style sheet.

docs/src/components/Sandbox/Sandbox.less
```diff
@@ -0,0 +1,86 @@
+@import (reference) "~seek-style-guide/theme";
+
+@sandbox-background: darken(@sk-blue, 13);
+@sandbox-preview-background: darken(@sandbox-background, 5);
+@sandbox-line-height: (@row-height * 5);
+@sandbox-marker-size: (@gutter-width / 2);
+@sandbox-marker-color: fade(lighten(@sk-blue, 50), 50);
+@sandbox-editor-height: (@row-height * 47);
```
michaeltaranto: What are your thoughts about making this a percentage? Reason is i tried to zoom out to get more code space and see more previews at once and the preview area grows but the editor area is locked to 47 grids rows.

docs/src/components/Sandbox/Preview/CollectStyles/CollectStyles.js
```diff
@@ -0,0 +1,62 @@
...
+
+  const sheetPromises = Array.from(styleNodes).map(el => {
+    if (el.nodeName === 'STYLE') {
+      return Promise.resolve(el.innerHTML);
+    }
+
+    const href = el.getAttribute('href');
+    const url = urlJoin(baseHref, href);
+
+    return axios.get(url)
```
mattsjones: I'd just use the fetch api here instead of introducing a new dependency. For your use case, it's almost the same code.

docs/src/components/Sandbox/Sandbox.less
markdalgleish: Yeah, good idea. I'll look at it.

docs/src/components/Sandbox/Preview/CollectStyles/CollectStyles.js
tamoore: Does the project already have the pollyfill?

docs/src/components/App/Header/Header.less
```diff
@@ -4,6 +4,10 @@
   background: @sk-blue;
 }
 
+.fullWidth > :global(*) {
```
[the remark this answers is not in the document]
markdalgleish: Not my proudest moment.

docs/src/components/Sandbox/Preview/CollectStyles/CollectStyles.js
mattsjones: Why would you need polyfill? This only runs in the browser right?

tamoore: More reason why axios would be a good choice no? As fetch still has gaps in support right?

mattsjones: Only IE11 is out. As this is a docs page I think we don't need to worry about that.

tamoore: IE11 is definitely a crap browser granted, but I'd only consider dropping it if there was a deficit due to the support cost. Otherwise, I wouldn't intentionally limit my audience for the cost of a simple polyfill / alternative.

mattsjones: Just checked. This page is completely broken in IE11. No surprise lol

docs/src/components/Sandbox/Sandbox.less
markdalgleish: Done 👌

docs/src/components/Sandbox/Preview/CollectStyles/CollectStyles.js
markdalgleish: Just swapped it out for native fetch. We can revisit later if we need to.

markdalgleish:
Just updated the PR to add an error boundary within the preview frames. This solves an issue where erroneous code could lead to an unrecoverable error that kills the whole page.

PR #479, opened 2018-05-03
feat(Checkbox): Adding field message support

conshow:
Reason for change:
Allowing checkbox to have a message such as a validation error associated with it similar to Textfield component.

BREAKING CHANGE:
Checkbox component now has some reserves space below for field messages. To maintain existing functionality you can set `message` is explicitly to `false` as per our other form components.

To retain existing behaviour you can do the following:
```diff
- <Checkbox id="id" label="Label" />
+ <Checkbox id="id" label="Label" message={false} />
```

Now supports the standard field props, for example:
```diff
- <Checkbox id="id" label="Label" />
+ <Checkbox id="id" label="Label" valid={false} message="Invalid"  />
```


markdalgleish:
Would you be able to add the message options to the demo page too, like we've done with the other form components?

tamoore:
This seems sensible although from the diff it didn't appear to be a breaking API change. Was there some discussion around flagging it as a breaking change?

conshow:
@tamoore The breaking change is that there is now always some extra white space for the field message at the bottom except if we expressly set message to false. This is so the form does not move up and down on showing/hiding the message.

PR #484, opened 2018-05-17
feat(forms): Add top-level value, checked, type, onChange/Focus/Blur

markdalgleish:
This non-breaking API change affects the following form components:
- **Autosuggest**
- **Checkbox**
- **Dropdown**
- **EmailField**
- **Radio**
- **TextField**
- **Textarea**

The following props can now be provided via the top-level component API, rather than via an `inputProps` object:
- **value**
- **checked** (for Checkbox and Radio)
- **type** (for Autosuggest and TextField)
- **onChange**
- **onFocus**
- **onBlur**

Currently, `value`, `checked`, `type`, `onChange`, `onFocus` and `onBlur` need to be provided via the `inputProps` object. Since these props are heavily used across all form fields, it adds a lot of friction to our component APIs. It also hinders our ability to create and use form libraries, since this is a custom, non-standard API.

**NOTE: Since `inputProps` is still supported, this API change is backwards compatible.**

## Migration Guide

For example, when migrating `TextField`:

```diff
-<TextField {...} inputProps={{ type: 'search', value, onChange: handleChange, onFocus: handleFocus, onBlur: handleBlur />
+<TextField {...} type="search" value={value} onChange={handleChange} onFocus={handleFocus} onBlur={handleBlur} />
```

As another example, when migrating `Checkbox`:

```diff
-<Checkbox {...} inputProps={{ value, checked: true, onChange: handleChange, onFocus: handleFocus, onBlur: handleBlur />
+<Checkbox {...} value={value} checked={true} onChange={handleChange} onFocus={handleFocus} onBlur={handleBlur} />
```

### DIFF
--- a/react/Autosuggest/Autosuggest.js
+++ b/react/Autosuggest/Autosuggest.js
@@ -1,24 +1,26 @@
 import styles from './Autosuggest.less';
-
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import ReactAutosuggest from 'react-autosuggest';
 import IsolatedScroll from 'react-isolated-scroll';
-
 import omit from 'lodash/omit';
-
 import TextField from '../TextField/TextField';
-
 import smoothScroll from '../private/smoothScroll';
 import smallDeviceOnly from '../private/smallDeviceOnly';
 
+/* eslint-disable react/no-deprecated */
 export default class Autosuggest extends Component {
   static displayName = 'Autosuggest';
 
   static propTypes = {
     id: PropTypes.string.isRequired,
-    inputProps: PropTypes.object.isRequired,
+    value: PropTypes.string.isRequired,
+    onChange: PropTypes.func.isRequired,
+    onFocus: PropTypes.func,
+    onBlur: PropTypes.func,
+    type: PropTypes.string,
+    inputProps: PropTypes.object,
     label: PropTypes.string,
     labelProps: PropTypes.object,
     className: PropTypes.string,
@@ -43,9 +45,11 @@ export default class Autosuggest extends Component {
   };
 
   static defaultProps = {
+    type: 'text',
     className: '',
     label: '',
     labelProps: {},
+    inputProps: {},
     showMobileBackdrop: false
   };
 
@@ -113,7 +117,18 @@ export default class Autosuggest extends Component {
   }
 
   render() {
-    const { inputProps, label, autosuggestProps, suggestionsContainerClassName, showMobileBackdrop } = this.props;
+    const {
+      value,
+      onChange,
+      onFocus,
+      onBlur,
+      type,
+      inputProps,
+      label,
+      autosuggestProps,
+      suggestionsContainerClassName,
+      showMobileBackdrop
+    } = this.props;
     const { theme = {} } = autosuggestProps;
 
     const allAutosuggestProps = {
@@ -134,7 +149,7 @@ export default class Autosuggest extends Component {
     return (
       <div>
         <ReactAutosuggest
-          inputProps={inputProps}
+          inputProps={{ value, onChange, onFocus, onBlur, type, ...inputProps }}
           ref={this.storeInputReference}
           {...allAutosuggestProps}
         />
--- a/react/Checkbox/Checkbox.js
+++ b/react/Checkbox/Checkbox.js
@@ -1,6 +1,5 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-
 import styles from './Checkbox.less';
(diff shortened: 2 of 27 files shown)


react/Autosuggest/Autosuggest.js
```diff
@@ -1,24 +1,26 @@
...
 import IsolatedScroll from 'react-isolated-scroll';
-
 import omit from 'lodash/omit';
-
 import TextField from '../TextField/TextField';
-
 import smoothScroll from '../private/smoothScroll';
 import smallDeviceOnly from '../private/smallDeviceOnly';
 
+/* eslint-disable react/no-deprecated */
```
markdalgleish: I had to add this comment here (and in `CustomMonthPicker.js`) because of an eslint-plugin-react update, due to our use of `componentWillUpdate`. To maintain backwards compatibility, I've left it alone for now and ignored the ESLint rule, but we'll have to rename/remove it when we upgrade to React v17.

react/Checkbox/Checkbox.js
```diff
@@ -24,11 +23,13 @@ export default class Checkbox extends Component {
   static propTypes = {
     id: PropTypes.string.isRequired,
     label: PropTypes.node.isRequired,
+    value: PropTypes.string,
```
mattsjones: Not sure we should expose `value` from Checkbox. I don't think it's technically valid.

markdalgleish: I didn't originally add it, but realised as I was preparing the PR that it's actually valid: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

Same for radio buttons too: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio

markdalgleish: As an example of where this is useful, you might have a large checklist and want to share a single event handler across every checkbox instance, checking `event.target.value` in the callback.

react/Textarea/Textarea.js
```diff
@@ -1,9 +1,7 @@
 import styles from './Textarea.less';
-
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-
```
mattsjones: Interesting that your getting rid of the newlines between import types. I thought this was to make a distinction between them (global/local/css-module). I find it makes the file easier to read. Would love a lint rule for it personally.

react/Autosuggest/Autosuggest.js
michaeltaranto: How about marking it as a warning so it's visible?

react/Checkbox/__snapshots__/Checkbox.test.js.snap
```diff
@@ -54,57 +55,11 @@ exports[`Checkbox inputProps should pass through other props to the input 1`] =
   className="root"
 >
   <input
+    checked={false}
     className="input"
     data-automation="first-name-field"
     id="testCheckbox"
-    type="checkbox"
-  />
-  <label
```
markdalgleish: Trying to figure out why this disappeared from the snapshot 🤔

markdalgleish: Ohhhh, it's just the weird way the diff is rendering. It's just because I removed a snapshot test.

react/Dropdown/Dropdown.js
```diff
@@ -26,24 +22,11 @@ export default class Dropdown extends Component {
...
-
-      if (typeof value !== 'string') {
-        return new Error(`Invalid prop \`inputProps.value\` of type \`${typeof value}\` supplied to \`${componentName}\`, expected \`string\`.`);
-      }
-
-      if (inputId && id) {
-        return new Error(`\`inputProps.id\` will be overridden by \`id\` in ${componentName}. Please remove it.`);
-      }
-    },
-    /* eslint-enable consistent-return */
```
markdalgleish: So glad to get rid of this 😅

react/Dropdown/Dropdown.js
```diff
@@ -79,14 +63,19 @@ export default class Dropdown extends Component {
...
     </option>);
   }
+
   renderSelect() {
-    const { id, inputProps, options, placeholder } = this.props;
+    const { id, value, onChange, onFocus, onBlur, inputProps, options, placeholder } = this.props;
     const inputStyles = classnames({
       [styles.dropdown]: true,
-      [styles.placeholderSelected]: !inputProps.value
+      [styles.placeholderSelected]: !value && !inputProps.value
```
markdalgleish: Almost missed this one 😅

markdalgleish:
Just gave the PR another thorough look, just to be sure. Looking good to go.

Issue #485, opened 2018-05-22
Large body text size


jahredhope:
# Motivation

Currently there are two supported body text sizes. small and standard.
There is a need to create important, standout or lead information in designs.

We've looked at using use-case based naming however these often don't support the variety of situations where it might be used:
- lead
- lede
- important
- intro

Currently work-arounds appear to be created by using subheading in areas where the text is not a heading. This causes semantic issues but also creates undesirable additional bottom margins that are supplied for headings.

# Proposed solution
`<Text large>`

A Text node with large will use the same standard size line-height, font-weight and bottom margin.
A Text node with large will have 18px font-size.
A Text node with large will not have the additional bottom margin supplied to heading sizes.

`large` will be added to the Text props interface making it available to components that use the current text props.

jahredhope:
The question of `line-height` was brought up in offline discussions. Whether the `large` body text should be over the standard 4 rows, or over 5 rows.

4 rows may look tight.
5 rows would result in differences in vertical space between body text which may be undesirable, and may make some text look too floaty.

PR #486, opened 2018-05-23
feat(Text): Add large body Text

jahredhope:
Details listed in issue https://github.com/seek-oss/seek-style-guide/issues/485

This change adds `large` Text size, exposing it to `<Text>` and all other components that uses `withTextProps`.

Large has a text size of 18 on both mobile and desktop, but in all other ways works as standard text.

EXAMPLE USAGE:

```js
<Text large>foo </Text>
```

### DIFF
--- a/react/Text/Text.sketch.js
+++ b/react/Text/Text.sketch.js
@@ -10,6 +10,7 @@ const sizes = [
   'headline',
   'heading',
   'subheading',
+  'large',
   'standard',
   'small'
 ];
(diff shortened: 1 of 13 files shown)


theme/type/type.less
```diff
@@ -36,6 +36,13 @@
 @subheading-type-row-span-mobile: 4;
 @subheading-type-weight-mobile: @sk-bold;
 
+@large-type-scale: 1.8;
+@large-type-row-span: 4;
```
markdalgleish: As discussed, can we change the row span to `5`?

react/Text/Text.sketch.js
```diff
@@ -33,11 +37,24 @@ export const text = {
...
+  ),
+  [`${symbolNames.standard}/5. Critical`]: (
+    <Text critical>Standard critical text</Text>
+  ),
+  [`${symbolNames.standard}/6. Link`]: (
+    <Text>
+      <TextLink href="#">Standard link text</TextLink>
+    </Text>
+  ),
+  [`${symbolNames.large}`]: <Text large>Large text</Text>,
```
markdalgleish: This line needs to be moved above the `standard` definitions, or the numbers are out of order, since `large` appears before `standard` in the `sizes` array defined above.

theme/type/type.less
```diff
@@ -36,6 +36,13 @@
...
+@large-type-weight: @sk-regular;
+@large-type-scale-mobile: 1.8;
+@large-type-row-span-mobile: 4;
```
markdalgleish: Same as above, obviously.

tgdev:
Any reason why this is being held up?

PR #490, opened 2018-05-28
fix(MonthPicker): Fix props passed to Dropdown

georgespyropoulos:
Monthpicker consumes dropdown and is currently passing it props using the deprecated inputProps
object. Converting it here to the new structure


react/MonthPicker/CustomMonthPicker/CustomMonthPicker.js
```diff
@@ -141,8 +141,10 @@ export default class CustomMonthPicker extends Component {
     const monthValue = String(month || '');
     const yearValue = String(year || '');
 
+    const sharedProps = {
+      onBlur: this.handleBlur
```
michaeltaranto: There are quite a few shared props that aren't apart of this object, would it make sense to add them here? I think it might be more performant to just add these directly to the `Dropdown`s below.

PR #501, opened 2018-06-14
feat(Footer): Update Saved Searches link to require auth

adenj:


michaeltaranto:
Curious as to why the autosuggest snapshot was updated? If there is a reason it'd be better as a separate PR, otherwise if you can refine this PR to be solely the `Footer` that would be preferred.

adenj:
@michaeltaranto Yep that was a brainfart. I've removed it

dborysiewicz:
Nudge on /sign-in page (https://www.seek.com.au/sign-in?returnUrl=/myactivity) is done so this is ready to go from my end 👍

PR #504, opened 2018-06-19
feat: Make FieldLabel & FieldMessage public

dborysiewicz:
We have use cases in which we need to use `FieldMessage` & `FieldLabel` on custom components and when reviewing these components they have been pretty stable and makes sense to make them public now.

Their should be no BREAKING CHANGE unless you where naughtily pulling in private components in which it will fail now. 

EXAMPLE USAGE:

Example of where we want to have a list of radio buttons and use the standard `FieldMessage`.

```js
import { FieldMessage } from 'seek-style-guide/react';

<div>
  {radios.map((radio, index) => (
    <div
      key={`${radio.id}${name}`}
      className={index === radios.length - 1 ? rowStyle : ''}
    >
      {renderRadio(radio, value, name, onChange, onBlur, !valid)}
    </div>
  ))}

  <FieldMessage
    {...{ id: `${name}-message`, valid, message, messageProps }}
  />
</div>
```

Example of adding `FieldLabel` to a custom ui component.

```js
import { FieldLabel } from 'seek-style-guide/react';

<FieldLabel
  {...{
    id,
    label: (
      <span>
        {label}
        {value ? (
          <ScreenReaderOnly>
            {' '}
            , currently selected: {value}
          </ScreenReaderOnly>
        ) : null}
      </span>
    ),
    labelProps,
    secondaryLabel,
    tertiaryLabel
  }}
/>
```


react/Autosuggest/Autosuggest.demo.js
```diff
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import Autosuggest from './Autosuggest';
 import styles from '../TextField/TextField.less';
 import classnames from 'classnames';
-import fieldMessageOptions from '../private/FieldMessage/FieldMessage.demo';
-import fieldLabelOptions from '../private/FieldLabel/FieldLabel.demo';
+import fieldMessageOptions from '../FieldMessage/FieldMessage.privateDemo';
```
markdalgleish: Is there a reason this can't just be `FieldMessage.demo`, like the other components?

dborysiewicz: Because get's auto imported with `https://github.com/seek-oss/seek-style-guide/blob/2962cce3fb8fc6e9ba853c2cfedc94c3ef6a1a34/docs/src/components/App/App.js#L13` and as it's not a normal demo (just imported into other demos) it fails.

dborysiewicz: Open to other suggestions/names?

markdalgleish: Ah, yeah, this is a tricky one to name. Not sure that this is "private" compared to the others. How about `demoFragment`?

dborysiewicz: Sounds good to me, I'll switch it over 👍

markdalgleish: Should this PR also add a proper demo for these components, rather than just fragments?

PR #506, opened 2018-06-21
feat(Breakpoints): Extra breakpoint

tgdev:
## Reason / Scope

While migrating pages from a legacy system to a new tech stack built on SKU and SEEK Style Guide there are often times where the legacy pages will have bespoke components that require different styles applied past the `@desktop` breakpoint.

Some in-browser prototyping suggests that the value of the existing `@desktop` variable does not need to be changed and that simply adding another, higher value can give us greater styling options across varying screen widths.

What this PR is putting forward is a new set of breakpoint variables that can be used alongside the existing ones.

### Breakpoint sizes

**740px** - existing (currently assigned to `@desktop`)

**910px** - new and may be altered as part of feedback for this PR but will likely land somewhere between 910px - 990px

### Variable names

`@small` === `@mobile`

`@medium` === `@desktop`

`@mediumMax` = new size - 1

`@large`= new size

## BREAKING CHANGE:

NOT a breaking API change as the existing variables `@mobile` and `@desktop` will remain to avoid such an issue.

## EXAMPLE USAGE:

Existing:

```css
.someClass {
  // base styles

  @media @mobile {
    // styles applied at 739px max
  }
  
  @media @desktop {
    // styles applied at 740px and up
  }
}
```

New:

```css
.someClass {
  // base styles

 @media @small {
    // styles applied up to 739px max
  }
  
  @media @medium {
    // styles applied at 740px and up
  }

  @media @mediumMax {
    // styles applied up to 909px max
  }

  @media @large {
    // styles applied at 910px and up
  }
}
```

### DIFF
--- a/theme/layout/breakpoints.less
+++ b/theme/layout/breakpoints.less
@@ -1,6 +1,19 @@
 // Breakpoint
 @responsive-breakpoint: 740px;
+@responsive-breakpoint-lg: 910px;
 
+@small: %(~"only screen and (max-width: %s)", (@responsive-breakpoint - 1));
+
+@medium: %(~"only screen and (min-width: %s)", @responsive-breakpoint);
+
+@mediumMax: %(
+  ~"only screen and (max-width: %s)",
+  @responsive-breakpoint-lg - 1
+);
+
+@large: %(~"only screen and (min-width: %s)", @responsive-breakpoint-lg);
+
+// Legacy variables to avoid breaking changes
 @mobile: %(~"only screen and (max-width: %s)", (@responsive-breakpoint - 1));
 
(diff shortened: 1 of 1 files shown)


kosanna:
@mediumMax seems to be too bespoke, that might be either gutter width or at least tablet. However, it feels that introducing a new breakpoint is not something that our new approach needs, feels that the problem should be solves from the other end with either adopting and aligning all the 'bespoke' things to use either mobile or desktop endpoints, if not possible, the seek some ux advise on how criteria could be met

tgdev:
Closing this PR for now as further discussion is required between internal teams at SEEK.

PR #508, opened 2018-06-22
feat(PasswordField): Add basic password field component

tgdev:
Similar to the existing `EmailField` component, this new component simply exports a `TextField` component with an input type of `password`.

The need for this component arose when working with the `@seek/forms-ui` lib where I had to create a wrapping component in my project to connect the `forms-ui` Field with the style guide's `TextField` which had a type of `password`.

NOTE: The more complex password component with Show/Hide functionality could be added to this basic component in the future.

EXAMPLE USAGE:

```js
import { PasswordField } from 'seek-style-guide/react';

<PasswordField />
```

### DIFF
--- a/react/PasswordField/PasswordField.js
+++ b/react/PasswordField/PasswordField.js
@@ -1,0 +1,43 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import TextField from '../TextField/TextField';
+
+export default class PasswordField extends Component {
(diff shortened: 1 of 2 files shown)


kosanna:
Why do we need a component for that while we could just pass a prop directly to the text field?

tgdev:
Closing as it's been pointed out that this component adds little value.

PR #509, opened 2018-06-25
feat(Pill): Accept content as children, support all valid node types

apoorvmalhotra:
Pills currently only support rendering strings as content. In order to support the full range of typographical components, you can now render arbitrary nodes inside of pills.

To make this clearer, we now support rendering pill content as children. Note that, while this is a backwards compatible change, you're strongly encouraged to migrate to the new API.

EXAMPLE USAGE:

```js
<Pill>The last word is <Strong>strong.</Strong></Pill>
```

MIGRATION GUIDE:

```diff
-<Pill text="Hello world" />
+<Pill>Hello world</Pill>
```

### DIFF
--- a/react/Pill/Pill.js
+++ b/react/Pill/Pill.js
@@ -1,6 +1,5 @@
 // @flow
 import styles from './Pill.less';
-import PropTypes from 'prop-types';
 
 import React, { Component } from 'react';
 import classnames from 'classnames';
@@ -10,7 +9,7 @@ import CrossIcon from '../CrossIcon/CrossIcon';
 import ScreenReaderOnly from '../ScreenReaderOnly/ScreenReaderOnly';
 
 type Props = {
-  text: PropTypes.node,
+  text: React$Node,
   onClose?: Function,
   className?: string
 };
--- a/react/Pill/Pill.js
+++ b/react/Pill/Pill.js
@@ -1,15 +1,15 @@
 // @flow
 import styles from './Pill.less';
-
 import React, { Component } from 'react';
+import type { Node } from 'react';
 import classnames from 'classnames';
-
 import Text from '../Text/Text';
 import CrossIcon from '../CrossIcon/CrossIcon';
 import ScreenReaderOnly from '../ScreenReaderOnly/ScreenReaderOnly';
 
 type Props = {
-  text: React$Node,
+  children?: Node,
+  text?: Node,
   onClose?: Function,
   className?: string
 };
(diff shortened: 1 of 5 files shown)


react/Pill/Pill.js
```diff
@@ -9,7 +10,7 @@ import CrossIcon from '../CrossIcon/CrossIcon';
 import ScreenReaderOnly from '../ScreenReaderOnly/ScreenReaderOnly';
 
 type Props = {
-  text: string,
+  text: PropTypes.node,
```
michaeltaranto: This component is using flow type definitions that compile to `PropTypes` rather than setting prop types directly. In this case we should use `React$Element` here.

michaeltaranto: Actually look at the [usage below](https://github.com/seek-oss/seek-style-guide/pull/509/files#diff-e73fc4fe5c618a913a9593e8926bf867R34) this api should probably not be changed as the `text` prop is rendered in a `Text` component which should not be the case for arbitrary children. We should probably have a quick chat about how best to accommodate this

michaeltaranto: Ignore me, use `React$Node` to support both strings and node https://github.com/facebook/flow/blob/master/lib/react.js#L12

react/Pill/Pill.test.js
```diff
@@ -15,6 +15,11 @@ describe('Pill:', () => {
     expect(pill).toMatchSnapshot();
   });
 
+  it('should render a child component', () => {
+    const ChildComponent = () => (<div>Hello</div>);
+    expect(render(<Pill text={<ChildComponent />} />)).toMatchSnapshot();
```
kosanna: You can use shallow here as well, because in this case we do not care what are actual html elements of a child

georgespyropoulos: Can you use the exisiting `renderPill` function to keep this test consistent with the others in this file?

react/Pill/Pill.js
```diff
@@ -9,7 +9,7 @@ import CrossIcon from '../CrossIcon/CrossIcon';
 import ScreenReaderOnly from '../ScreenReaderOnly/ScreenReaderOnly';
 
 type Props = {
-  text: string,
+  text: React$Node,
```
tgdev: This seems confusing to keep this prop as `text` if it's no longer only rendering text.

Is a valid alternative to remove the prop (breaking change) and allow child nodes be rendered instead?

eg:

```
Props = {
  children: ReactNode,
  ...otherProps
}

render: <Pill {...props}>{this.props.children}</Pill>

// usage
<Pill><Text>Some text</Text></Pill>
<Pill><div>Some child node</div></Pill>
```

or swap text for `render` prop:

```
Props = {
  render: ReactNode,
  ...otherProps
}

<Pill render={<Text>Some text</Text>} />
<Pill render={<div>Some child node</div>} />
```

markdalgleish: I've actually been wanting to make this change anyway—I think you're right that now is a good time to do it.

I've taken the reigns of this PR for now, so I'll add this.

markdalgleish: To be clear, I'm going to add support for `<Pill>Hello world</Pill>` in addition to `<Pill text="Hello world" />`, to maintain backwards compatibility.

markdalgleish: Done. Updated the PR title and release notes.

react/Pill/Pill.js
```diff
@@ -27,25 +27,29 @@ export default class Pill extends Component<Props> {
...
 
   renderStaticPill() {
-    const { text, className, ...restProps } = this.props;
+    const { children, text, className, ...restProps } = this.props;
+    const content = children || text;
+
     return (
       <span className={classnames(className, styles.staticPill)} {...restProps}>
-        <Text baseline={false} raw>{text}</Text>
+        <Text baseline={false} raw>{content}</Text>
```
michaeltaranto: Sorry if i missed this, but should `children` be rendered inside `<Text>` or instead of?

markdalgleish: Inside, since we don't want arbitrary text styles within a pill.

michaeltaranto: Since we're turning `baseline` off that should be fine

PR #511, opened 2018-06-26
fix(reset): Remove browser default styling of mark elements

michaeltaranto:
The introduction of the `<Highlight>` component using the `<mark>` element has exposed that we previously had no reset for the default browser styles of this element — namely the `background-color` and `color`.

Anyone using the `<mark>` element elsewhere in their apps and expecting the default browser styling (extremely unlikely) will have to manually reinstate these rules.

### DIFF
--- a/react/StyleGuideProvider/reset.less
+++ b/react/StyleGuideProvider/reset.less
@@ -47,8 +47,11 @@ table {
   border-collapse: collapse;
   border-spacing: 0;
 }
+
+/* Custom reset rules */
 mark {
   background-color: transparent;
   color: inherit;
 }
+
 /* stylelint-enable */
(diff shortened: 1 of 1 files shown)


markdalgleish:
Would it be worth explicitly flagging this as a custom reset rule? Right now, it looks like it's part of the Meyer reset.

michaeltaranto:
@markdalgleish would you prefer moving it to the `StyleGuideProvider.less`?

markdalgleish:
@michaeltaranto Yeah, that could work.

markdalgleish:
@michaeltaranto Sorry, actually—it won't work, because the reset is used in `StandaloneProvider.less`. Just adding a comment like `// Custom reset rules` should be enough, I think.

michaeltaranto:
Done

PR #512, opened 2018-06-29
feat(SeeMore): Add the see more component

mattsjones:
`SeeMore` is used to hide excessive content but still show a smaller view of it. If the content is not larger then the set amount then it will be rendered unaltered. 

EXAMPLE USAGE:

```js
<SeeMore maxRows={20}>
  <Text>{someLongText}</Text>
</SeeMore>
```


mattsjones:
Closing because Mark doesn't like me

PR #513, opened 2018-07-02
feat(ReadMore): Add 'ReadMore' component

mattsjones:
The `ReadMore` component will truncate text if it is longer then the specified number of max lines of text. If it is equal or shorter it will render as normal. Currently, the component assumes your text is standard size. We could open up other sizes in future but opted to keep the component simple for now. 

EXAMPLE USAGE:

```js
<ReadMore maxLines={5}>
  <Text>
    {longText}
  </Text>
</ReadMore>
```

### DIFF
--- a/react/ReadMore/ReadMore.demo.js
+++ b/react/ReadMore/ReadMore.demo.js
@@ -49,8 +49,8 @@ ReadMoreContainer.propTypes = {
 };
 
 export default {
-  route: '/readmore',
-  title: 'ReadMore',
+  route: '/read-more',
+  title: 'Read More',
   category: 'Typography',
   component: ReadMore,
   container: ReadMoreContainer,
--- a/react/ReadMore/ReadMore.demo.js
+++ b/react/ReadMore/ReadMore.demo.js
@@ -52,7 +52,7 @@ ReadMoreContainer.propTypes = {
 export default {
   route: '/read-more',
   title: 'Read More',
-  category: 'Typography',
+  category: 'Layout',
   component: ReadMore,
   container: ReadMoreContainer,
   initialProps: {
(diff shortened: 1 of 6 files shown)


react/private/themeVars.json
```diff
@@ -0,0 +1,5 @@
+{
```
mattsjones: This file was added as a short-term solution to using theme variables within javascript.

markdalgleish: To simplify the build tool requirements (e.g. to maintain webpack v1 support), this might be better off as a JS module rather than JSON.

markdalgleish:
This is missing an export in `react/index.js` — luckily, noticed this when I went to try it out in the sandbox and got an error that `ReadMore is not defined`.

react/ReadMore/ReadMore.demo.js
```diff
@@ -0,0 +1,110 @@
...
+);
+
+ReadMoreContainer.propTypes = {
+  component: PropTypes.any,
+  componentProps: PropTypes.object.isRequired
+};
+
+export default {
+  route: '/readmore',
+  title: 'ReadMore',
```
markdalgleish: For consistency, should be `route: '/read-more', title: 'Read More'`.

Sorry, this should probably be automated somehow.

react/ReadMore/ReadMore.less
```diff
@@ -0,0 +1,20 @@
...
+.fadeOut {
+  background: linear-gradient(to bottom, rgba(255, 255, 255, 0) 0%, white 100%);
+  position: absolute;
+  height: (@row-height * 4);
+  width: 100%;
+  bottom: 0;
+}
+
+.chevron {
+  margin-left: 5px;
```
markdalgleish: Don't want to block this now, but we should figure out a way to remove the need for custom horizontal icon spacing.

PR #515, opened 2018-07-03
test(Alert): Update test with correct propType

tgdev:
Alert component was reporting an invalid prop type on one of it's test for passing in a boolean instead of a function.

### DIFF
--- a/react/Alert/Alert.test.js
+++ b/react/Alert/Alert.test.js
@@ -63,7 +63,8 @@ describe('Alert:', () => {
   });
 
   it('should render a close button', () => {
-    const alert = renderAlert({ onClose: true });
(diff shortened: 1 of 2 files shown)


georgespyropoulos:
@tgdev The build is failing on snapshot tests. Is this build running a different version of `react-autowhatever` to the one you used to generate the snapshots?

tgdev:
Hmmm this is weird... Seems like @markdalgleish did an update of these same snapshots as part of #502 which this PR is undoing...

tgdev:
@markdalgleish and @georgespyropoulos seems like the `Autosuggest` snapshot issue was a local thing for me. Have updated this PR accordingly with new Title, description and changes.

PR #516, opened 2018-07-04
feat(PartnerSites): Move to standalone component

tgdev:
With multiple product sites like Candidate, Advertiser, SEEK Learning, etc all rolling their own versions of the partner sites bar at the top of their site, it seemed like a good opportunity to take the one out from the style guide header and modify it a little to be consumable on it's own as well as within the style guide header.

[Please see the included Demo page to see these modifications.](http://seek-style-guide--e88ae9f2ad3f87a393d59a3637346f9a09fa97a1.surge.sh/partner-sites)

For new consumers of this component: the existing analytics attached to each link are included so as to avoid any breaking changes with the Header. 

You'll likely need to discuss with your tracking expert as to if/how this might impact you.

EXAMPLE USAGE:

```js
import { PartnerSites } from 'seek-style-guide/react';

const linkRenderer = props => (<a {...props} />);

// No active site
<PartnerSites locale='AU' linkRenderer={linkRenderer} />

// Jobs set as active site
<PartnerSites locale='AU' linkRenderer={linkRenderer} activeSite='Jobs' />

// NZ label shown first in list (when switched to NZ locale)
<PartnerSites locale='NZ' linkRenderer={linkRenderer} />
```


react/Header/Header.js
```diff
@@ -122,14 +121,7 @@ export default function Header({
             divider={divider}
           />
         </Hidden>
-        <Hidden print mobile className={styles.topBanner}>
```
michaeltaranto: These `topBanner` styles should be split up, `Header` should still be responsible for the flex ordering as that is specific to this usage of PartnerSites. possible worth keeping the `Hidden` concept external as well? Thoughts @markdalgleish?

markdalgleish: Yeah, agreed—I think the flex ordering needs to be hoisted to a wrapping div in `Header`.

I think the `Hidden` concept belongs inside `PartnerSites` though, since the component itself isn't responsive, and is designed to never be printed globally either.

react/PartnerSites/PartnerSites.js
```diff
@@ -0,0 +1,30 @@
...
+      </div>
+    </div>
+  </Hidden>);
+}
+
+PartnerSites.displayName = 'PartnerSites';
+
+PartnerSites.propTypes = {
+  locale: PropTypes.oneOf(['AU', 'NZ']).isRequired,
+  linkRenderer: PropTypes.func.isRequired,
```
markdalgleish: Like `Header`, this prop should be optional.

react/PartnerSites/Products/Products.js
```diff
@@ -79,7 +98,12 @@ export default function PartnerSites({ locale, linkRenderer }) {
...
-  locale: PropTypes.string.isRequired,
-  linkRenderer: PropTypes.func.isRequired
+Products.propTypes = {
+  locale: PropTypes.oneOf(['AU', 'NZ']).isRequired,
+  linkRenderer: PropTypes.func.isRequired,
+  activeProduct: PropTypes.oneOf(['Jobs', 'Courses', 'Businesses for sale', 'Volunteering'])
+};
+
+Products.defaultProps = {
+  activeProduct: null
```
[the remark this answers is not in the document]
tgdev: @markdalgleish It's currently set to null because we don't set an active one on https://talent.seek.com.au.

If this is just an over-site and `activeProduct` should be set to `Jobs` regardless whether you're on the Candidate or Employer "Jobs" site, then I'm happy to set default to `Jobs`.

Also, I'm not sure if you're aware but there are talks going on at the moment to include Employer in this list as it's own nav item. Not sure if that impacts setting a default `activeProduct` now or if we just refactor this Component if/when this new item comes into the mix.

react/PartnerSites/PartnerSites.js
tgdev: Will update

react/PartnerSites/PartnerSites.js
```diff
@@ -0,0 +1,30 @@
...
+        <Locales locale={locale} linkRenderer={linkRenderer} />
+      </div>
+    </div>
+  </Hidden>);
+}
+
+PartnerSites.displayName = 'PartnerSites';
+
+PartnerSites.propTypes = {
+  locale: PropTypes.oneOf(['AU', 'NZ']).isRequired,
```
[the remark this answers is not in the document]
tgdev: Will update

react/Header/Header.js
markdalgleish: Yep, just the ordering, since that's the only styling opinion that's specific to its usage within `Header`.

react/PartnerSites/Products/Products.js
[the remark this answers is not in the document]
markdalgleish: Yep, makes sense—let's not include a default 👍

react/Header/Header.js
tgdev: 👍

react/Header/Header.js
```diff
@@ -122,14 +121,9 @@ export default function Header({
...
-          <div className={styles.topBannerContent}>
-            <PartnerSites locale={locale} linkRenderer={linkRenderer} />
-            <div className={styles.locale}>
-              <Locales locale={locale} linkRenderer={linkRenderer} />
-            </div>
-          </div>
-        </Hidden>
+        <div className={styles.topBanner}>
+          <PartnerSites locale={locale} linkRenderer={linkRenderer} activeProduct='Jobs' />
```
markdalgleish: Does this need to be `activeSite` now?

react/PartnerSites/Sites/Sites.js
```diff
@@ -79,7 +98,12 @@ export default function PartnerSites({ locale, linkRenderer }) {
   );
 }
 
-PartnerSites.propTypes = {
-  locale: PropTypes.string.isRequired,
-  linkRenderer: PropTypes.func.isRequired
+Sites.propTypes = {
+  locale: PropTypes.oneOf(['AU', 'NZ']).isRequired,
+  linkRenderer: PropTypes.func.isRequired,
```
markdalgleish: This prop should be optional, I think.

markdalgleish: We just need to provide a default `linkRenderer`, like we do with `Header`.

tgdev: This is a child/private component so it should be required from my understanding. I believe the PartnerSites props are optional as per convention.

react/Header/Header.js
tgdev: Whoops. Will update

markdalgleish (approved):

react/PartnerSites/PartnerSites.js
```diff
@@ -0,0 +1,35 @@
...
+      <div className={styles.locale}>
+        <Locales locale={locale} linkRenderer={linkRenderer} />
+      </div>
+    </div>
+  </Hidden>);
+}
+
+PartnerSites.displayName = 'PartnerSites';
+
+PartnerSites.propTypes = {
```
kosanna: shall we flow type "new" components?

tgdev: @kosanna I've not used flow before and have already spent a considerable amount of time on this PR in between project work. Is this something that could be picked up as a separate PR or would this block this PR from being merged?

kosanna (approved):

react/PartnerSites/PartnerSites.js
kosanna: go for it, we'll address it separately together 👍

tgdev: Thanks @kosanna. That sounds good to me 👍

PR #517, opened 2018-07-17
test(css): Parse CSS class names in unit tests

markdalgleish:
I split this PR out from another piece of work where I've started programatically iterating over the class names exported by a Less file. This clashes with our usage of `identity-obj-proxy` in unit tests, since `Object.keys` returns `[ ]` when called on a proxy object. Instead, we now have a custom Jest transformer that quickly extracts class names via a simple regex.

Not surprisingly, this change has highlighted differences in our snapshots where the proxy object didn't behave like a real CSS module.

### DIFF
--- a/react/PartnerSites/Locales/__snapshots__/Locales.test.js.snap
+++ b/react/PartnerSites/Locales/__snapshots__/Locales.test.js.snap
@@ -3,7 +3,6 @@
 exports[`Locales: should render with locale of AU first and active 1`] = `
 <nav
   aria-labelledby="Locales"
-  class="root"
   role="navigation"
 >
   <span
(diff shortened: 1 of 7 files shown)


test/cssModules.transform.js
```diff
@@ -0,0 +1,21 @@
+// Really naive, regex-based class name extractor
+
+module.exports = {
+  process: src => {
+    const classNames = [];
+
+    const classNameRegEx = /\.([_a-zA-Z0-9]+)(?=[\.:\[\s{])/gm;
```
jahredhope: I don't think we have any classes that use `-` but it might be good to support it to avoid later confusion:
Just replacing `[_a-zA-Z0-9]` to `[_a-zA-Z0-9-]+` would work. (though even that is not technically correct, but only because it's too broad)

markdalgleish: Done.

react/PartnerSites/Locales/__snapshots__/Locales.test.js.snap
```diff
@@ -3,7 +3,6 @@
 exports[`Locales: should render with locale of AU first and active 1`] = `
 <nav
   aria-labelledby="Locales"
-  class="root"
```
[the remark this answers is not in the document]
markdalgleish: It's because this is a non-existent class—a good example of why this is a big improvement.

I opted to leave this in for now, rather than changing production code, since this is supposed to be scoped to tests only.

PR #521, opened 2018-07-27
test(css): Prefix mock class names with file name to avoid conflicts

markdalgleish:
This issue was discovered when trying to write tests for `Text.js` where it imported a `secondary` class from two different CSS modules.

As a bonus, I feel like this actually makes the snapshots easier to read since it makes it explicit where each class name is coming from.

### DIFF
--- a/test/cssModules.transform.js
+++ b/test/cssModules.transform.js
@@ -1,7 +1,8 @@
-// Really naive, regex-based class name extractor
+const path = require('path');
 
 module.exports = {
-  process: src => {
+  process: (src, srcPath) => {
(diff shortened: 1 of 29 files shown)


test/cssModules.transform.js
```diff
@@ -13,7 +14,7 @@ module.exports = {
 
     const cssModule = {};
     classNames.forEach(className => {
-      cssModule[className] = className;
+      cssModule[className] = `${fileName}__${className}`;
```
markdalgleish: Just to make the review process a little easier, this file is the core of the change. The rest of the PR is just updated test code and snapshots to accomodate it.

PR #524, opened 2018-07-31
feat(FieldLabel): Decouple FieldLabel styling with valid prop

georgespyropoulos:
Styling of the field label has previously been controlled by the valid prop. This can lead to unwanted styling as the validity of the form does not always dictate the desired styling. For example, a field message for a valid field should not always be green because this text could be used as grey help text to guide the user instead.

A new `tone` prop has been added to the `FieldLabel`, `Checkbox`, `Dropdown`, `MonthPicker`, `Textarea` and `TextField` components. This can be set to 'positive', 'critical' or 'help' which will style the FieldLabel text as green, pink or grey.

MIGRATION GUIDE:

This is not a breaking change however the valid prop should no longer be used to control styling of fields.

Before:

```js
<Textfield valid={false} message="Required" />
```

After:

```js
<Textfield tone="critical" message="Required" />
```

Before

```js
<Textfield valid={true} message="Looks good" />
```

After:

```js
<Textfield tone="positive" message="Looks good" />
```


kosanna:
Isn't that actually a breaking change? Api is changed 🤔

react/Checkbox/Checkbox.js
```diff
@@ -110,19 +112,19 @@ export default class Checkbox extends Component {
   }
 
   render() {
-    const { className, id, valid, message, messageProps } = this.props;
+    const { className, id, valid, message, messageProps, tone } = this.props;
 
     const rootClassNames = classnames({
       [styles.root]: true,
-      [styles.invalid]: valid === false,
+      [styles.invalid]: typeof tone !== 'undefined' ? tone === TONE.CRITICAL : valid === false,
```
kosanna: Probably could specifically check for oneOf (in values) the types, otherwise anything passed will break it

react/Checkbox/Checkbox.test.js
```diff
@@ -33,6 +33,18 @@ describe('Checkbox', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('should render the invalid style for tone="critical" even if the field is valid', () => {
```
kosanna: To the same point, need a test case
```
  it('should render the invalid style for tone="someNonExistingTone" even if the field is valid', () => {
    expect(...)...;
  });```

File: react/FieldMessage/FieldMessage.js [hunk 3]
```diff
@@ -1,31 +1,34 @@
...
 import ErrorIcon from '../ErrorIcon/ErrorIcon';
 import TickCircleIcon from '../TickCircleIcon/TickCircleIcon';
 
 import Text from '../Text/Text';
-
-export default class FieldMessage extends Component {
+import { TONE } from '../private/tone';
+
+type Props = {
+  id: string,
```
kosanna: 👏

react/FieldMessage/FieldMessage.js
```diff
@@ -1,31 +1,34 @@
...
+  invalid?: boolean,
+  help?: string,
+  helpProps?: Object,
+  valid?: boolean,
+  message: false | Node | string,
```
kosanna: keep forgetting it, but Node should already include 2 other cases 🤔

react/FieldMessage/FieldMessage.js
```diff
@@ -66,10 +67,13 @@ export default class FieldMessage extends Component {
     return null;
   }
 
-  renderMessageIcon() {
-    const { valid } = this.props;
+  renderMessageIcon = () => {
+    const { tone, valid } = this.props;
+    const toneDefined = typeof tone !== 'undefined';
+    const showErrorIcon = toneDefined ? tone === TONE.CRITICAL : valid === false;
```
kosanna: Could we reuse showCritical here?

react/FieldMessage/FieldMessage.js
```diff
@@ -35,27 +38,25 @@ export default class FieldMessage extends Component {
...
+
       return (
         <Text
           id={id}
           small
           {...restMessageProps}
-          critical={(valid === false && !secondary) || critical}
-          positive={(valid === true && !secondary) || positive}
-          secondary={typeof valid === 'undefined' || secondary}
+          critical={(showCritical && !secondary) || critical}
```
georgespyropoulos: Suggestions from @conshow - could `critical`, `positive` and `secondary` override both tone and valid here to reduce complexity?

react/Section/Section.js
```diff
@@ -1,32 +1,30 @@
...
-  SECONDARY: 'secondary',
-  TERTIARY: 'tertiary'
-};
+type Props = {
+  children: any,
+  className?: string,
+  header?: boolean,
+  pullout?: boolean,
+  slim?: boolean,
+  tone?: 'positive' | 'critical' | 'neutral',
```
markdalgleish: Is this missing 'help'?

react/Checkbox/Checkbox.js
georgespyropoulos: Yes there's a prop type check for `tone` in this component so console warnings will show up if an invalid value is passed in

react/Checkbox/Checkbox.test.js
georgespyropoulos: Are we testing anything else like this? If we were to write a test like that, what kind of value do you think it would provide us? We throw a prop type warning for an invalid value and I'm not sure we need to go any further than that

react/FieldMessage/FieldMessage.js
georgespyropoulos: It seems that it's not quite the same, I changed it to `message: Node` and got this error.

screen shot 2018-09-25 at 5 02 33 pm

georgespyropoulos:
> Isn't that actually a breaking change? Api is changed 🤔

No, this isn't a breaking change. You can continue to use the `valid` prop and everything will continue to work. Switching over to Tone is encourage because ideally we'd like to move away from `valid` having any control of the style of field message

PR #525, opened 2018-07-31
feat(ReadMore): Adds backgroundColor prop

bjervisSEEK:
There is a need to use the ReadMore component on backgrounds that are not white (like a card is). This PR adds a `backgroundColor` prop, which takes a string describing the context the component is used in (currently "card" or "body"), and sets the colour of the fade element to match. Defaults to "card".

EXAMPLE USAGE:

```js
<ReadMore backgroundColor="body" maxLines={3}>
  <Text>Content</Text>
</ReadMore>
```

### DIFF
--- a/react/ReadMore/ReadMore.demo.js
+++ b/react/ReadMore/ReadMore.demo.js
@@ -7,6 +7,8 @@ import Text from '../Text/Text';
 import Card from '../Card/Card';
 import Section from '../Section/Section';
 
+import styles from './ReadMore.demo.less';
+
 const longText = (
   <Text>
     Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus vitae nulla
@@ -37,7 +39,11 @@ const shortText = (
 );
 
 const ReadMoreContainer = ({ component: DemoComponent, componentProps }) => (
-  <Card style={{ width: 500 }}>
+  <Card
+    style={{ width: 500 }}
+    className={
+      componentProps.backgroundColor === 'body' ? styles.grayBackground : ''
+    }>
     <Section>
       <DemoComponent {...componentProps} />
     </Section>
@@ -121,6 +127,32 @@ export default {
           transformProps: props => ({
             ...omit(props, 'maxLines'),
             maxRows: 20
+          })
+        }
+      ]
+    },
+    {
+      label: 'Color',
+      type: 'radio',
+      states: [
+        {
+          label: 'Default Background',
+          transformProps: props => ({
+            ...omit(props, 'backgroundColor')
+          })
+        },
+        {
+          label: 'Card',
+          transformProps: props => ({
+            ...props,
+            backgroundColor: 'card'
+          })
+        },
+        {
+          label: 'Body',
+          transformProps: props => ({
+            ...props,
+            backgroundColor: 'body'
           })
         }
       ]
--- a/react/ReadMore/ReadMore.js
+++ b/react/ReadMore/ReadMore.js
@@ -27,7 +27,8 @@ type Props = {|
   maxLines?: number,
   maxRows?: number,
   moreLabel: string,
-  lessLabel: string
+  lessLabel: string,
+  backgroundColor: 'card' | 'body'
 |};
 type State = {|
   showMore: boolean,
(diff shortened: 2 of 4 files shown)


michaeltaranto:
As a general comment, I feel the api doesn't feel right — `backgroundColor` doesn't actually accept a colour. Have you discussed any alternatives? Happy to chat offline if its easier

tgdev:
@michaeltaranto / @bjervisSEEK could you use a `tone` prop which is similar to other existing style guide components. The values could `light` and `dark` or something along those lines (just ideas really, the `tone` prop is the main suggestion)

bjervisSEEK:
After a brainstorm in feed yesterday, we're feeling that `backgroundComponentName` would fit a little better

theme/palette/elements.less
```diff
@@ -5,6 +5,7 @@
 @sk-highlight: #1d9ede;
 @sk-footer: @sk-mid-gray-light;
 @sk-background: @sk-gray-light;
+@sk-card-bg: @sk-white;
```
kosanna: should we then replace it in a card component as well? Feels very disconnected though, because card might be transparent as well 🤔

react/ReadMore/ReadMore.js
```diff
@@ -27,7 +27,8 @@ type Props = {|
   maxLines?: number,
   maxRows?: number,
   moreLabel: string,
-  lessLabel: string
+  lessLabel: string,
+  backgroundColor: 'card' | 'body'
```
kosanna: Then I believe we should have a variable for body background as well, so in case we decide to change the body background it stays connected

react/ReadMore/ReadMore.js
```diff
@@ -86,11 +87,19 @@ class ReadMore extends PureComponent<Props, State> {
...
 
   render() {
-    const { children, maxLines, maxRows, moreLabel, lessLabel } = this.props;
+    const {
+      children,
+      maxLines,
+      maxRows,
+      moreLabel,
+      lessLabel,
+      backgroundColor
```
kosanna: Agree about api, Name would be better, however why don't we just supply colour so it's more scalable?

theme/palette/elements.less
bjervisSEEK: Most of the colour organisation work is waiting for proper CSS-in-JS. `@sk-card-bg` was an interim measure suggestion from @michaeltaranto. We also want to add `@sk-body` but that can wait for the colour re-org

react/ReadMore/ReadMore.js
bjervisSEEK: Because the less implementation of this uses `@sk-background`, it will stay connected.

react/ReadMore/ReadMore.demo.js
```diff
@@ -125,6 +131,32 @@ export default {
...
+          label: 'Default Background',
+          transformProps: props => ({
+            ...omit(props, 'backgroundColor')
+          })
+        },
+        {
+          label: 'Card',
+          transformProps: props => ({
+            ...props,
+            backgroundColor: 'card'
```
kosanna: should it be backgroundComponentName? Or I'm tired?

georgespyropoulos: Oh yeah good catch, I think it should be

react/ReadMore/ReadMore.js
```diff
@@ -86,11 +87,19 @@ class ReadMore extends PureComponent<Props, State> {
...
+      maxRows,
+      moreLabel,
+      lessLabel,
+      backgroundComponentName
+    } = this.props;
     const { tooLong, showMore, mounted } = this.state;
 
     const truncate = mounted ? !showMore && tooLong : true;
     const showFade = truncate && mounted;
+    const fadeColor = styles[backgroundComponentName || 'card'];
```
kosanna: Will it be better to declare it in default props? Leaving it up to you

react/ReadMore/ReadMore.js
```diff
@@ -102,7 +111,9 @@ class ReadMore extends PureComponent<Props, State> {
       <div>
         <div className={styles.content} style={contentStyle}>
           <div ref={this.setTextRef}>{children}</div>
-          {showFade ? <div className={styles.fadeOut} /> : null}
+          {showFade ? (
+            <div className={[styles.fadeOut, fadeColor].join(' ')} />
```
kosanna: why not just classnames?

react/ReadMore/ReadMore.less
```diff
@@ -8,13 +8,30 @@
...
+      fade(@sk-card-bg, 0) 0%,
+      @sk-card-bg 100%
+    );
+}
+
+.body {
+  background:
+    linear-gradient(
+      to bottom,
+      fade(@sk-background, 0) 0%,
```
kosanna: should we do a background mixin with color argument here so we do not duplicate it?

react/ReadMore/ReadMore.demo.js
```diff
@@ -37,7 +39,11 @@ const shortText = (
 );
 
 const ReadMoreContainer = ({ component: DemoComponent, componentProps }) => (
-  <Card style={{ width: 500 }}>
+  <Card
+    style={{ width: 500 }}
+    className={
+      componentProps.backgroundColor === 'body' ? styles.grayBackground : ''
```
conshow: Is that meant to be updated as well?

PR #526, opened 2018-08-03
fix(Autosuggest): Fix accessibility bug for screen readers

kosanna:
Current version of Autosuggest is unusable in with screen readers, this will reduce accessibility
in OSX, but it's still usable regardless.


kosanna:
Thanks @dborysiewicz and @conshow 👏

PR #528, opened 2018-08-03
fix(Autosuggest): Add resolutions to fix yarn install for pinned package

dborysiewicz:
Yarn install was not de-duping the dependency correctly and still installing the incorrect version of react-autowhatever.

This is a non issue with npm and also will be removed when we rebuild Autosuggest.


kosanna:
I think we can even remove the direct dependency now. Do you want to give it a go?

PR #529, opened 2018-08-06
fix(Autosuggest): Fork react-autosuggest by copying code into style guide

dborysiewicz:
Switched to loading a internal version of `react-autosuggest@9.3.4`, the reason for this fork is to pin one of it’s dependencies `react-autowhatever` to a previous version `10.1.1` due to accessibility issues in the latest version. This was not possible via explicitly adding the wanted version of autowhatever as a dependency as there as issues getting this supported with both npm & yarn.

This should be temporary as our long term solution would be to build our own in house Autosuggest as currently there are still issues (but at least useable for now) with the current Autosuggest with certain screen readers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,17 +49,13 @@
   },
   "homepage": "https://github.com/seek-oss/seek-style-guide#readme",
   "dependencies": {
+    "@seek/react-autosuggest": "9.3.5",
     "basekick": "^3.0.0",
     "classnames": "^2.2.5",
     "lodash": "^4.17.4",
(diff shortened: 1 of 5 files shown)


kosanna:
wtf,  404 Not Found: @seek/react-autosuggest@9.3.5

kosanna:
Are you working on the same branch or we can close this one? @dborysiewicz

dborysiewicz:
This is the copy to internally @kosanna 😄

PR #532, opened 2018-08-10
feat(Footer): Added 'new' badge to 'Work For SEEK' footer link

lgshaw:
Helping to expose new or updated pages via the footer, a 'new' badge (similar to the one on the main nav) has been proposed for footer links.
Initially added for the updated 'Work for SEEK' pages.

BREAKING CHANGE:
none

EXAMPLE USAGE:

```js
{
    name: 'Work for SEEK',
    href: '/work-for-seek',
    analytics: 'toolbar:work+for+seek',
    secondary: true,
    newBadge: true
},
```

### DIFF
--- a/react/Footer/Footer.js
+++ b/react/Footer/Footer.js
@@ -50,7 +50,7 @@ export default class Footer extends Component {
     const { locale, linkRenderer, authenticationStatus } = this.props;
 
     return (!specificLocale || specificLocale === locale) ?
-      <div key={i}><FooterLink secondary={secondary} linkRenderer={linkRenderer} authenticationStatus={authenticationStatus} {...restProps}>{name}</FooterLink></div> :
+      <FooterLink secondary={secondary} linkRenderer={linkRenderer} authenticationStatus={authenticationStatus} key={i} {...restProps}>{name}</FooterLink> :
       null;
   }
 
--- a/react/Footer/Footer.js
+++ b/react/Footer/Footer.js
@@ -50,7 +50,7 @@ export default class Footer extends Component {
     const { locale, linkRenderer, authenticationStatus } = this.props;
 
     return (!specificLocale || specificLocale === locale) ?
-      <div><FooterLink secondary={secondary} linkRenderer={linkRenderer} authenticationStatus={authenticationStatus} key={i} {...restProps}>{name}</FooterLink></div> :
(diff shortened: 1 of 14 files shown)


react/Pill/Pill.js
```diff
@@ -28,10 +29,11 @@ export default class Pill extends Component<Props> {
 
   renderStaticPill() {
     const { children, text, className, ...restProps } = this.props;
+    const footerStyle = this.props.footer ? styles.footerBadge : null;
```
markdalgleish: The header/footer badge styles are non-standard compared to the rest of the system, so I don't think we should make this an option of `Pill`. If you hard-code this style inside of `Footer`, that should be fine.

react/Footer/data/connect.js
```diff
@@ -8,7 +8,8 @@ export default [
     name: 'Work for SEEK',
     href: '/work-for-seek',
     analytics: 'toolbar:work+for+seek',
-    secondary: true
+    secondary: true,
+    badge: 'new'
```
markdalgleish: Are we supporting multiple badge types? Would this be better off as `new: true`?

kosanna:
Might we ever have not NEW badges? If not I don't see the point of the prop not being a boolean like `isNew` instead

react/Footer/Footer.js
```diff
@@ -50,7 +50,7 @@ export default class Footer extends Component {
     const { locale, linkRenderer, authenticationStatus } = this.props;
 
     return (!specificLocale || specificLocale === locale) ?
-      <FooterLink secondary={secondary} linkRenderer={linkRenderer} authenticationStatus={authenticationStatus} key={i} {...restProps}>{name}</FooterLink> :
+      <div key={i}><FooterLink secondary={secondary} linkRenderer={linkRenderer} authenticationStatus={authenticationStatus} {...restProps}>{name}</FooterLink></div> :
```
kosanna: why do we need an extra wrapper here?

kosanna: Can use Fragment here

lgshaw: You are correct @kosanna , made obsolete with my latest changes. Removed!

kosanna:
With minimum code looks like that, don't feel it's bad 
screen shot 2018-08-16 at 10 29 41 am

PR #533, opened 2018-08-17
feat(Textarea): Add validation highlighting to textarea

georgespyropoulos:
Demo: http://seek-style-guide--8ca44530df8619f8b8fc61e036522e3e80d4abcf.surge.sh/textarea/

Background: We're pre-populating form fields with imported data that may not conform to our validation rules. In these instances, it's helpful to be able to show users which part of their text is invalid so they can easily identify and fix problems. I'm hoping this solution will make things easier for users and increase conversions on forms.

I decided to move away from using contenteditable once I did a bit of cross browser testing. I'm now using a regular textarea with a div that shadows the textarea and displays the highlighting. Here's an article that explains the approach in detail https://codersblock.com/blog/highlight-text-inside-a-textarea/

EXAMPLE USAGE:

Invalid word (matches all occurences)

```js
<Textarea
  id="description"
  label="Description"
  invalidText="bad"
/>
```

Invalid range of characters

```js
<Textarea
  id="description"
  label="Description"
  invalidText={{
    end: 20,
    start: 10
  }}
/>
```

Multiple invalid ranges

```js
<Textarea
  id="description"
  label="Description"
  invalidText={[
    {
      end: 20,
      start: 10
    },
    {
      end: 40,
      start: 30
    }
  ]}
/>
```


conshow:
Would it make sense to have a more self-descriptive property such as highlightText so that it's clearer what it does and so that it may be used not just for invalid text?

georgespyropoulos:
@conshow That's an interesting question... I just created this PR https://github.com/seek-oss/seek-style-guide/pull/534 so perhaps we could also allow tone to be set for the text you wish to highlight? Can you think of a place where we would highlight text for a reason other than it being invalid?

conshow:
I don't really have a use case. But having invalidText property highlight text feels a bit similar to the valid property setting the field message to green or red. I thought a clearer name might help avoid this type of confusion.

react/Textarea/Textarea.js
```diff
@@ -85,21 +102,46 @@ export default class Textarea extends Component {
...
+    const { id, value, invalidText, onChange, onFocus, onBlur, inputProps } = this.props;
+    let html;
+    const highlightErrors = typeof invalidText !== 'undefined';
+    if (highlightErrors) {
+      html = formatInvalidText(value, invalidText, styles.invalidTextChunk);
+    }
 
-    return (
-      <textarea {...allInputProps} />
+    const textarea = (props, classname) => (
```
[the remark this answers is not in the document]
mattsjones: Render textarea

react/Textarea/Textarea.js
```diff
@@ -15,6 +16,7 @@ function combineClassNames(props = {}, ...classNames) {
   };
 }
 
+/* eslint-disable react/no-deprecated, react/no-danger */
```
kosanna: Not sure there's anything to ignore here 🤔 no-danger could be moved to respective line

react/Textarea/Textarea.js
```diff
@@ -40,7 +42,10 @@ export default class Textarea extends Component {
         return new Error(`\`value\` must be supplied if \`${propName}\` is set`);
       }
     },
-    secondaryLabel: PropTypes.string
+    secondaryLabel: PropTypes.string,
+    invalidText: PropTypes.oneOfType([
+      PropTypes.string, PropTypes.object, PropTypes.arrayOf(PropTypes.object)
+    ])
     /* eslint-enable consistent-return */
```
kosanna: This one could be removed, just return null earlier

react/Textarea/Textarea.less
```diff
@@ -60,5 +61,37 @@
...
+}
+
+.highlightTextareaWrapper {
+  position: relative;
+}
+
+.highlightTextarea {
+  background-color: transparent;
+  position: relative;
+  z-index: 2;
```
kosanna: Could we please use variables here?

PR #539, opened 2018-09-18
chore(deps): Upgrade eslint deps

michaeltaranto:
Upgrade all eslint related dependencies.


tgdev:
@michaeltaranto is it worth adding to `renovate.json`?

PR #543, opened 2018-09-27
fix(Textarea): Fix highlighting ranges API

georgespyropoulos:
The API for setting highlighting ranges was not very intuitive. Updating to make it more human usable rather than machine usable.

Note that this is technically a breaking change but it's not being marked as one given the text highlighting feature was released yesterday


kosanna:
Do you want to release it as a breaking change?

PR #546, opened 2018-10-02
feat(Header): Update icons

kosanna:
- Update icons colours on mobile menu
- Add company reviews icon

Screenshot:
https://user-images.githubusercontent.com/15971854/46346720-5f379780-c68c-11e8-946b-7f4450ef5579.png

Extras:
- Add flow to Header component
- Convert propTypes to flow types
- Use Hidden component where appropriate
- Extract Badge to a separate function so it's not copy-pasted with all props every time


react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -76,25 +76,33 @@
...
+.icon.saveSearches {
+  color: @sk-blue-lighter;
+  @media @mobile {
+    color: @sk-highlight;
+  }
+}
+
 .icon.recommendedJobs { color: @sk-green-light; }
 
+.crSvg {
```
mbfisher: I can't see where this is used anywhere - did I miss it?

react/Header/UserAccountMenu/UserAccountMenu.js
```diff
@@ -17,267 +19,278 @@ import { AUTHENTICATED, UNAUTHENTICATED, AUTH_PENDING } from '../../private/auth
...
+const PROFILE = 'Profile';
+const SAVED_SEARCHES = 'Saved Searches';
+const SAVED_AND_APPLIED = 'Saved & Applied Jobs';
+const APPLIED_JOBS = 'Applied Jobs';
+const RECOMMENDED_JOBS = 'Recommended Jobs';
+const SETTINGS = 'Settings';
+const CAREER_ADVICE = 'Career Advice';
+const COMPANY_REVIEWS = 'Company Reviews';
+
+type TabsType = typeof JOB_SEARCH |
```
georgespyropoulos: 👏

react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -76,25 +76,33 @@
...
-.icon.jobSearch { color: @sk-blue-lighter; }
+.icon.jobSearch { color: @sk-teal; }
 .icon.profile { color: @sk-purple; }
 .icon.saveJobs { color: @sk-pink; }
-.icon.saveSearches { color: @sk-blue-lighter; }
+
+.icon.saveSearches {
+  color: @sk-blue-lighter;
+  @media @mobile {
+    color: @sk-highlight;
```
michaeltaranto: Curious why this colour is different on mobile? Also it would be preferred to use a blue from the palette rather than cross use the `highlight` colour if possible.

kosanna: I think Carmelia was saying the there's no blue that she'd prefer here instead, but I'll doublecheck. Difference is because works differently on dark and light background, this is the only responsive blue icon

PR #547, opened 2018-10-08
feat(Badge): Add Badge

shawn-dsz:
## Description
Badges are used to label status such as 'new', 'expired' and 'follower'. Use the primary for when you need the badge to be loud, and secondary if you need it to be quieter. Tones should be used to make sure the colours meaning is consistent.
<details>
 <summary>Design Specifications</summary>

</details>

## Usage

```js
import {  Badge } from 'seek-style-guide/react';

<Badge strong tone="positive">13 new</Badge>
```

### DIFF
--- a/react/Badge/Badge.js
+++ b/react/Badge/Badge.js
@@ -42,8 +42,7 @@ export default class Badge extends PureComponent<Props> {
           [styles.positive]: positive,
           [styles.strong]: strong
         })}
-        {...this.props}
-      >
+        {...this.props}>
         {children}
       </div>
     );
--- a/react/Badge/Badge.js
+++ b/react/Badge/Badge.js
@@ -5,15 +5,18 @@ import classnames from 'classnames';
 
 import styles from './Badge.less';
 
+type Tone =
+  | 'accent'
+  | 'critical'
+  | 'info'
+  | 'neutral'
+  | 'positive'
+  | 'secondary';
+
 type Props = {
   children: React$Node,
-  strong?: boolean,
-  accent?: boolean,
-  critical?: boolean,
-  info?: boolean,
-  neutral?: boolean,
-  positive?: boolean,
-  secondary?: boolean
+  tone?: Tone,
+  primary?: boolean
 };
 
 export default class Badge extends PureComponent<Props> {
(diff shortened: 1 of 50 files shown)


react/Badge/Badge.less
```diff
@@ -0,0 +1,48 @@
...
+  font-weight: @sk-medium;
+  height: @height;
+  display: inline-flex;
+  align-items: center;
+  border-radius: 3px;
+  padding: 0 10px;
+  .__style(@sk-black);
+}
+
+.__style(@color) {
```
georgespyropoulos: What about calling this something more descriptive like `badgeColour`?

react/Badge/Badge.js
```diff
@@ -0,0 +1,53 @@
+// @flow
+
+import React, { PureComponent } from 'react';
+import classnames from 'classnames';
+
+import styles from './Badge.less';
+
+type Props = {
```
georgespyropoulos: We've used a `tone` prop in `Section` for critical, info, positive etc. https://github.com/seek-oss/seek-style-guide/blob/master/react/Section/Section.js#L15. Not saying this is wrong but we should try and have the same API for both components.

react/Badge/Badge.js
```diff
@@ -0,0 +1,42 @@
...
+
+  render() {
+    const { children, tone, primary } = this.props;
+
+    return (
+      <div
+        className={classnames({
+          [styles.root]: true,
+          [styles[tone]]: tone,
+          [styles.primary]: primary
```
michaeltaranto: The previous implementation deliberately stayed away from using `primary`/`secondary` as the relationship is not necessarily hierarchical, rather more to do with visual weight. @stephaniemoss?

theme/palette/brand.less
```diff
@@ -5,6 +5,7 @@
 @sk-blue-light: #184da8;
 @sk-blue-lighter: #2765cf;
 @sk-pink: #e60278;
+@sk-pink-dark: #cc026a;
```
michaeltaranto: I believe this is less of a brand colour and more of a darker critical variant as you've done with `positive` and `info`. What are your thoughts?

shawn-dsz: Yes, this is not brand colour,  I put it here as I thought this was a variation on `@sk-pink` similar to `@sk-blue-dark`. What do you think about moving it to `elements` as `@sk-critical-dark` rather than `@sk-pink-dark`.

react/Badge/Badge.js
stephaniemoss: Yeah, this is a problem. We need to get together to discuss what the logic is behind the visual weight as it applies to things like buttons and alerts.

shawn-dsz: How about `strong`?

theme/palette/brand.less
michaeltaranto: Thats the point i was hoping to make, its not a variant on `sk-pink`, its a darker variant of `criticial`. The fact that critical is `sk-pink` is implementation detail.

PR #549, opened 2018-10-08
feat(accessibility): Accessibility and SEO updates

chardos:
This is part of our initiative to increase accessibility across the site. These changes are the result of a discussion with both people from accessbility, and people from SEO (to make sure accessibility concerns did not affect our SEO).

- `H1`'s in the footer reduced to `H4`'s.
- Removal of unnecessary hidden `H1`'s
- `aria-labelled-by` changed to `aria-label`s, to facilitate above point.


react/Footer/Footer.js
```diff
@@ -59,72 +58,62 @@ export default class Footer extends Component {
...
+                  { social.map(this.renderLink) }
+                </ToggleContainer>
+              </FooterNav>
+
+              <FooterNav secondary label="Employers">
+                { employers.map(this.renderLink) }
+              </FooterNav>
+            </div>
+
+            <div className={styles.copyright}>
```
michaeltaranto: Can we confirm that this was correctly changed from a `<nav>` to a `<div>`?

chardos: Yeap, have confirmed with Adem this does not need to be a `<nav>`.

PR #552, opened 2018-10-11
fix(Flow): Fix flow errors with Badge and Section

georgespyropoulos:
Not defaulting some props in `Badge` and `Section` has caused a few flow errors in apps that consume the styleguide (see screenshot). The issue is that level and tone can be undefined when used to look up a style

screen shot 2018-10-11 at 1 55 59 pm


react/Badge/Badge.js
```diff
@@ -25,7 +25,7 @@ export default class Badge extends PureComponent<Props> {
   props: Props;
 
   render() {
-    const { children, tone, strong } = this.props;
+    const { children, tone = '', strong } = this.props;
```
michaeltaranto: Should this be defaulting to `neutral`?

georgespyropoulos: Yeah could do. My aim was for it to not default to undefined. Happy to change it to neutral instead

michaeltaranto: Well really, we could say that `tone` is not optional and default it, that might be a more accurate contract. Does that make sense?

kosanna: We've already had it with Highlight, seems sensible to default to neutral

georgespyropoulos: Yeah that's probably more accurate here. Away from my laptop but I'll make tone required later this arvo.

kosanna: https://github.com/seek-oss/seek-style-guide/blob/master/react/Highlight/Highlight.js#L15

craigcollie:
Hey @georgespyropoulos, this might be outside of the scope of this PR, but we've found that using `$Values<typeof TONE>` a simple way of declaring enum types from constants without having to declare types separately.

georgespyropoulos:
Hey @craigcollie that's a good tip! In this case, not all tones are the same in all components. The idea is to make a type local to the component with the tone values that the component accepts. But definitely handy bit of syntax if your component accepts all of them

react/Badge/Badge.js
```diff
@@ -4,13 +4,14 @@ import React, { PureComponent } from 'react';
...
   | 'accent'
-  | 'critical'
-  | 'info'
-  | 'neutral'
-  | 'positive'
+  | typeof TONE.CRITICAL
+  | typeof TONE.INFO
+  | typeof TONE.NEUTRAL
+  | typeof TONE.POSITIVE
   | 'secondary';
```
kosanna: Are accent and secondary only specific for this component? Should we move them to tones?

react/Section/Section.js
```diff
@@ -22,8 +22,8 @@ export default function Section({
   header,
   pullout,
   slim,
-  tone,
-  level,
+  tone = '',
```
kosanna: we don't have neutral style for the section, but it's declared in prop types. Should we remove it from there?

react/Badge/Badge.js
```diff
@@ -4,13 +4,14 @@ import React, { PureComponent } from 'react';
...
 import styles from './Badge.less';
+import { TONE } from '../private/tone';
 
 type Tone =
   | 'accent'
-  | 'critical'
-  | 'info'
-  | 'neutral'
-  | 'positive'
+  | typeof TONE.CRITICAL
```
kosanna: If we start using it across multiple components probably it make sense to create a shared TONE type 🤔 Not necessarily in this PR, just thinking

react/Section/Section.js
georgespyropoulos: Ah yes, I will update that. We have an info style but it was missing from the prop type so I will add that one in

react/Section/Section.js
```diff
@@ -22,8 +22,8 @@ export default function Section({
...
+  level = '',
```
kosanna: Is defaulting here helps achieve anything?

react/Badge/Badge.js
georgespyropoulos: We have one but not all components accept the same values for tone https://github.com/seek-oss/seek-style-guide/blob/master/react/private/tone.js#L2

react/Section/Section.js
georgespyropoulos: Yeah I get flow errors in my app without defaulting to an empty string here

craigcollie: If you're setting a default prop for `tone`, then you shouldn't have to declare it as an optional type. Might be why flow is throwing an error.

kosanna: This is not a proper fix because it will be converted to a required prop and will be throwing console warnings. We've found the issue in the flow config in profile ui repo

craigcollie: I usually only make props optional if I haven't declared a default prop. Otherwise it's technically not really required since it has a fallback.

georgespyropoulos: I'm not seeing any console warnings...

georgespyropoulos: @craigcollie I agree - it's not really required. You don't need to pass anything to satisfy the contract with the component

kosanna: then it's an optional, right?

react/Section/Section.js
```diff
@@ -47,5 +47,7 @@ Section.defaultProps = {
   className: '',
   header: false,
   pullout: false,
-  slim: false
+  slim: false,
+  tone: '',
```
kosanna: But there's no empty string declared as a proptype? 🤔

kosanna: I think it would be nice if we group all the defaults in one place 
https://github.com/seek-oss/seek-style-guide/pull/552/files#diff-04146d12928269aed074c3da275eed92R21
https://github.com/seek-oss/seek-style-guide/pull/552/files#diff-04146d12928269aed074c3da275eed92R47

georgespyropoulos: No 🤔... flow doesn't seem to care about this

kosanna: I tried removing defaulting, seems flow doesn't care at all

kosanna: React warning do not care at all 🤔 Any thoughts on that?

As a short summary, I feel this is not ideal to default to an empty string because it does not mean anything from the component point of view

georgespyropoulos: I think from the component point of view, you don't need to pass in any of these props because a default is provided for you. I wouldn't want to be warned about `tone` here because it's valid to not provide one. Most apps don't use the `tone` prop with `Section`

kosanna:
If anyone else's happy to look at the PR? I feel like people now wait on my approval here. Officially stating - I can't make my mind on that one so would like to abstain from approval because I think the proper fix is to fix flow config on the app itself, not changing style-guide component this way.

react/Section/Section.js
kosanna: if it were proptypes what would you put there. `tone: PropTypes.string.isRequired` or `tone: PropTypes.string`?

georgespyropoulos: I think it should be `tone: PropTypes.string.isRequired` because the Section component requires a tone prop. The default props provide one if it isn't passed in. I realise this isn't how prop types have been used throughout the styleguide but I think this is more correct. 

Our flow types are describing the component and stating which props is requires. Tone is required for the component to render in this case so I think the type should mark tone as required.

PR #554, opened 2018-10-11
feat(Navigation): Update Header and Footer to use Badge component

kosanna:
Currently custom component is used to render badges on Header and Footer.
Changing it to use new Badge component.


react/Footer/FooterLink/FooterLink.less
```diff
@@ -15,15 +15,5 @@
...
-  font-size: 12px;
-  font-weight: 500;
-  line-height: 18px;
-  height: 18px;
-  color: @sk-white;
-  background-color: @sk-purple;
-  padding: 0 @row-height;
-  border-radius: 9px;
-  margin-left: 8px;
+  margin-left: @gutter-width / 4;
```
michaeltaranto: This should not proportionate to the gutter, its more accurately a non-breaking space character.

react/Header/Navigation/Navigation.less
```diff
@@ -64,10 +64,13 @@
   margin-left: 5px;
   @media only screen and (max-width: @container-width) {
     position: absolute;
-    top: -4px;
-    right: (@gutter-width / 2);
+    right: -(@gutter-width / 4);
```
michaeltaranto: As above.

react/Header/Navigation/Navigation.less
```diff
@@ -64,10 +64,13 @@
...
+    top: -3.5 * @row-height;
```
michaeltaranto: Again not certain this is a multiple of row height. We should take care to use these principals only where it makes sense, not for all x and y values.

michaeltaranto:
This PR will need to be updated once [this PR](https://github.com/seek-oss/seek-style-guide/pull/555) is merged.

kosanna:
This PR had the latest of the mentioned PR, so just merge is required

react/Badge/Badge.js
```diff
@@ -17,7 +17,8 @@ type Tone =
 type Props = {
   children: React$Node,
   tone?: Tone,
-  strong?: boolean
+  strong?: boolean,
+  className?: string
```
kosanna: will just clean this up as no longer valid

react/Header/UserAccountMenu/UserAccountMenu.js
```diff
@@ -54,7 +54,7 @@ const clearLocalStorage = () => {
   }
 };
 
-const BadgeComponent = () => <Hidden desktop component={NewBadge} key="new" className={styles.newBadge} />;
+const BadgeComponent = () => <Hidden desktop className={styles.newBadge}><Badge strong tone="info" key="new">New</Badge></Hidden>;
```
michaeltaranto: The `key` needs to be hoisted to the `Hidden` component as that is the component requiring an identifier

kosanna:
Is here anything outstanding? 🙏

react/Header/Navigation/Navigation.js
```diff
@@ -51,24 +51,28 @@ export default function Navigation({ locale, linkRenderer, activeTab, newBadgeTa
...
-                            className={badgeStyles}
-                          />
+                          <span
+                            className={classnames({
+                              [styles.newBadge]: true,
+                              [styles.newBadge_isShort]: isShort,
+                              [styles.newBadge_isLast]: items.length - 1 === key
+                            })}>&nbsp;
+                            <Badge
+                              key={name}
```
michaeltaranto: The `key` needs to be hoisted to the `span` now

react/Header/Navigation/Navigation.less
```diff
@@ -61,13 +61,15 @@
...
-  margin-left: 5px;
   @media only screen and (max-width: @container-width) {
     position: absolute;
-    top: -4px;
-    right: (@gutter-width / 2);
+    right: 0;
+    top: -@touchableTextHeight / 2 + 2px;
     &.newBadge_isShort {
-      right: @gutter-width * 1.5;
+      right: 17%;
```
michaeltaranto: Is this correct as a percentage? Seems an odd number thats all

kosanna: Yeah, it's more correct as a percentage, it's weird because it's custom and depends on the width of the tab and copy on the badge. So there can't be anything not odd

react/Header/Navigation/Navigation.js
```diff
@@ -51,24 +51,28 @@ export default function Navigation({ locale, linkRenderer, activeTab, newBadgeTa
...
+                          <span
                             key={name}
-                            className={badgeStyles}
-                          />
+                            className={classnames({
+                              [styles.newBadge]: true,
+                              [styles.newBadge_isShort]: isShort,
+                              [styles.newBadge_isLast]: items.length - 1 === key
+                            })}>&nbsp;
+                            <Badge
```
georgespyropoulos: Any reason you've gone with `<Badge children="New" />` over `<Badge>New</Badge>`? I think the rest of the styleguide uses the later so perhaps that would keep things more consistent?

kosanna: Don't see any difference really...

PR #555, opened 2018-10-12
fix(Badge): Prevent className overrides

kosanna:
Currently internal classes will be overwritten with a className passed onto the badge.

```diff
<Badge className={styles.foo}>something</Badge> // will overwrite Badge styling
```

Also fixes console warning:

image

### DIFF
--- a/react/Badge/Badge.js
+++ b/react/Badge/Badge.js
@@ -2,7 +2,7 @@
 
 import React, { PureComponent } from 'react';
 import classnames from 'classnames';
-import { omit } from 'lodash';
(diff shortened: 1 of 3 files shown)


react/Badge/Badge.js
```diff
@@ -2,6 +2,7 @@
 
 import React, { PureComponent } from 'react';
 import classnames from 'classnames';
+import { omit } from 'lodash';
```
michaeltaranto: Can you import this by path rather than destructure, ie. `lodash/omit`

PR #557, opened 2018-10-16
fix(webpack): Find an open port on start, if port busy

chardos:
Currently, seek-style-guide is running on port 3000, as is many other projects. Right now, the easiest way to run 2 projects simultaneously is to change the port temporarily on one of them.

This PR updates the style guide to check if port 3000 is open on `npm start`, and then increment to the next available port if it is busy.

### DIFF
--- a/docs/server.js
+++ b/docs/server.js
@@ -1,19 +1,27 @@
 const webpack = require('webpack');
 const WebpackDevServer = require('webpack-dev-server');
-const config = require('./webpack.dev.config');
+const createConfigWithPort = require('./webpack.dev.config');
 const opn = require('opn');
-const port = 3000;
+const portfinder = require('portfinder');
+portfinder.basePort = 3000;
 
-new WebpackDevServer(webpack(config), {
-  publicPath: config.output.publicPath,
-  contentBase: __dirname,
-  hot: true,
-  historyApiFallback: true,
-  disableHostCheck: true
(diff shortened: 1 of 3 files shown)


docs/webpack.dev.config.js
```diff
@@ -14,113 +14,116 @@ const appPaths = [
...
 
-const config = decorateClientConfig({
-  entry: [
-    'core-js/modules/es6.symbol',
-    'react-hot-loader/patch',
-    'webpack-dev-server/client?http://localhost:3000',
-    'webpack/hot/only-dev-server',
-    path.resolve(__dirname, 'src/client-render')
-  ],
+const createConfigWithPort = (port) => {
```
chardos: No actual changes to the config, other than wrapping it in a function that receives a port, then passes it to line 22:

`webpack-dev-server/client?http://localhost:${port}`

docs/server.js
```diff
@@ -1,19 +1,27 @@
 const webpack = require('webpack');
 const WebpackDevServer = require('webpack-dev-server');
-const config = require('./webpack.dev.config');
+const createConfigWithPort = require('./webpack.dev.config');
 const opn = require('opn');
-const port = 3000;
+const portfinder = require('portfinder');
```
michaeltaranto: I think you missed saving this to `package.json`

chardos: Good spot!

PR #558, opened 2018-10-17
docs(site): Add search bar

bjervisSEEK:
### Reason
The GitHub pages for the style guide are a great source of information, but can be difficult to navigate and find things quickly. The categories aren't necessarily in order, and it might not be clear where everything lives.

### Change
This PR adds a new search bar to the hamburger menu, to allow for quick and easy navigation to components. Key features:
- Who doesn't love keyboard shortcuts? Press "/" to open the search bar, a la GitHub, and "Esc" to close it. With the search bar open, use the up and down arrow keys to choose a component, and press Enter to go.
- Search is based on component titles (as displayed) but is insensitive to casing or spacing. For example "textf" will match "Text Field". It's a substring search, so "txtf" won't work.
- Search bar will autofocus on open, and select the text, so you can either type over the top, or add to the end and refine the search

### Visual
styleguidesearch

### DIFF
--- a/docs/src/components/App/Header/Header.js
+++ b/docs/src/components/App/Header/Header.js
@@ -52,9 +52,11 @@ class Header extends Component {
   }
 
   componentDidMount = () => {
-    const searchbar = this.searchBarRef.current.input;
     document.addEventListener('keydown', this.handleDocumentKeydown);
-    searchbar.addEventListener('keydown', this.handleSearchbarKeydown);
+  };
+
+  componentWillUnmount = () => {
+    document.removeEventListener(this.handleDocumentKeydown);
   };
 
   handleDocumentKeydown = event => {
@@ -177,6 +179,9 @@ class Header extends Component {
                           label="Search"
                           ref={this.searchBarRef}
                           onChange={this.handleSearch}
+                          inputProps={{
+                            onKeyDown: this.handleSearchbarKeydown
+                          }}
                         />
                       </Card>
 
--- a/docs/src/components/App/Header/Header.js
+++ b/docs/src/components/App/Header/Header.js
@@ -8,14 +8,15 @@ import { Link, withRouter } from 'react-router-dom';
 import {
   PageBlock,
   Card,
-  Highlight,
   Section,
-  ChevronIcon,
   Text,
   TextField,
   ScreenReaderOnly
 } from 'seek-style-guide/react';
 import Logo from './Logo/Logo';
+import WithHighlighting from './WithHighlighting/WithHighlighting';
+import { generateRouteList } from './headerUtils';
+
 import demoSpecExports from '../../../../../react/*/*.demo.js';
 const demoSpecs = demoSpecExports.map(x => x.default);
 const allRoutes = [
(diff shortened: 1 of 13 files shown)


docs/src/components/App/Header/Header.js
```diff
@@ -1,18 +1,81 @@
...
+      </Highlight>
+    );
+  }
+
+  return children;
+};
+
+const buildRoutes = inputRoutes =>
+  inputRoutes.reduce((acc, route) => {
+    const category = route.category ? route.category : 'Other';
```
chardos: If you wanted to, I believe this could be short circuited with:

`const category = route.category || 'Other';`

docs/src/components/App/Header/Header.less
```diff
@@ -140,3 +140,14 @@
...
+
+.highlightWrapper {
+  position: relative;
+}
+
+.highlightedIcon {
+  position: absolute;
+  top: 16px;
+  left: -24px;
+  transform: rotate(-90deg);
```
georgespyropoulos: This style is available via a prop in chevronIcon https://github.com/seek-oss/seek-style-guide/blob/master/react/ChevronIcon/ChevronIcon.js#L23

docs/src/components/App/Header/Header.js
```diff
@@ -1,18 +1,81 @@
...
+  { route: '/typography', title: 'Typography', category: 'Guides' },
+  { route: '/page-layout', title: 'Page Layout', category: 'Guides' },
+  { route: '/sandbox', title: 'Sandbox', category: 'Tools' },
+  ...demoSpecs
+];
 
-export default class Header extends Component {
+const SEARCH_BAR_ID = 'search-bar-field';
+
+const WithHighlighting = ({ highlighted, children }) => {
```
georgespyropoulos: Wonder if it would be worth moving some of these functions into another file. This one has become rather long

docs/src/components/App/Header/Header.js
```diff
@@ -1,18 +1,81 @@
...
+  }
+
+  return 1;
+};
+
+class Header extends Component {
   static propTypes = {
-    fullWidth: PropTypes.bool
+    fullWidth: PropTypes.bool,
+    history: PropTypes.any
```
georgespyropoulos: History is an object right?

docs/src/components/App/Header/Header.js
bjervisSEEK: I've split out a bit of the sorting logic into a new file, as well as the WithHighlighting component

docs/src/components/App/Header/headerUtils.js
```diff
@@ -0,0 +1,48 @@
+import { pipe } from 'lodash/fp';
+
+export const buildRoutes = inputRoutes =>
+  inputRoutes.reduce((acc, route) => {
```
trevorgk: solid use of `reduce`

docs/src/components/App/Header/headerUtils.js
```diff
@@ -0,0 +1,48 @@
...
+    const category = route.category || 'Other';
+    const initialArrayVal = acc[category] || [];
+    return {
+      ...acc,
+      [category]: [...initialArrayVal, route]
+    };
+  }, {});
+
+export const sortRoutes = (routeA, routeB) => {
```
trevorgk: I would suggest renaming this to `compareRoutes`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Syntax

PR #560, opened 2018-10-18
feat(Section): Fix pullout section causing page overflow

kosanna:
Reset side padding instead of compensating it to 0 with negative margins.

<details>
 <summary>Screenshot</summary>

image
</details>


kosanna:
Right, good point, let me think more on that 🤔

PR #564, opened 2018-10-22
fix(ReadMore): Change show more to use TextLink instead of a button

kosanna:
Currently Show more uses a transparrent button component which introduces redundant spacing around the link.

As agreed with @stephaniemoss moving to use TextLink with 3 rows below.

Also this fixes blue highlight border introduced by button focus state when it's clicked.

Reuse TextLink chevron instead of a custom.


react/ReadMore/ReadMore.js
```diff
@@ -127,16 +127,15 @@ class ReadMore extends PureComponent<Props, State> {
             <div className={classnames(styles.fadeOut, fadeColor)} />
           ) : null}
         </div>
-        {showMoreLessButton ? (
-          <Button color="transparent" onClick={this.handleShowMore}>
+        {showMoreLessButton && (
+          <TextLink
+            chevron={showMore ? 'up' : 'down'}
+            className={styles.showMore}
+            component={props => <Text {...props} strong />}
```
michaeltaranto: Did you consider making the `Text` component a `button`?

kosanna: Happy to do <a /> tag (shouldn't it be a part of TextLink btw?), button is doing weird things in terms of styles

kosanna: Ah right, it's already in textLink. let's just do a?

PR #565, opened 2018-10-24
feat(ReadMore): Add new background colour

georgespyropoulos:
Adding a new colour so I can use the read more component for this design.

I realise that this particular style / colour may be of no use to anyone else so I'm open to changing the API to allow a colour instead of a component name.

screen shot 2018-10-24 at 11 47 29 am


michaeltaranto:
[Closing as per comment](https://github.com/seek-oss/seek-style-guide/pull/565#pullrequestreview-177447620)

PR #566, opened 2018-10-24
feat(AccordionItem): Add Accordion Item

adenj:
## Description
Accordion Items are used to convey large amounts of information to users without presenting them with a wall of text, allowing them to reveal only the content which interests them.

## Usage
```js
import { AccordionItem } from 'seek-style-guide/react';

<AccordionItem
  title={() => <span>Hello world</span>}
  id="abc123"
>
  This is some content inside an AccordionItem
</AccordionItem>
```
This will work out of the box, but also allows for a more manual approach to controlling each items state through the optional `onClick` and `expanded`  props. E.g.
```js
<AccordionItem
  title={() => <span>Hello world</span>}
  id="abc123"
  expanded
  onClick={() => {}}
>
  This is some content inside an AccordionItem
</AccordionItem>
```


react/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,82 @@
...
+      this.setState({
+        expanded: nextProps.expanded
+      });
+    }
+  }
+
+  props: Props;
+
+  handleChange = () => {
+    if (this.props.onClick) {
```
michaeltaranto: I would suggest explicitly checking that this is a function
```suggestion
    if (typeof this.props.onClick === 'function') {
```

react/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,82 @@
...
+  children: React$Node,
+  title: React$Node,
+  id: string | number,
+  expanded?: boolean,
+  onClick?: Function,
+  className?: string
+};
+
+type State = {
+  expanded?: boolean
```
michaeltaranto: This is not optional on `state`,
```suggestion
  expanded: boolean
```

react/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,82 @@
...
+};
+
+export default class AccordionItem extends Component<Props, State> {
+  static defaultProps = {
+    expanded: false
+  }
+
+  constructor(props: Props) {
+    super(props);
+    const { expanded } = this.props;
```
michaeltaranto: This should be destructuring from `props` i believe
```suggestion
    const { expanded } = props;
```

react/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,82 @@
...
+    });
+  }
+
+  render() {
+    const { children, className, id, title } = this.props;
+    const { expanded } = this.state;
+    return (
+      <div
+        className={classnames({
+          ...(className ? { [className]: className } : {})
```
michaeltaranto: The `classnames` library handles the conditional aspect for us. You can simplify this as:
```suggestion
          [className]: className
```

react/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,82 @@
...
+        })}>
+        <input onChange={this.handleChange} id={id} className={styles.toggle} type="checkbox" checked={expanded} />
+        <label
+          htmlFor={id}
+          className={classnames({
+            [styles.toggleLink]: true,
+            [styles.toggleLabel]: true
+          })}>
+          <div>{title}</div>
```
michaeltaranto: Is this `div` needed?

react/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,82 @@
...
+        className={classnames({
+          ...(className ? { [className]: className } : {})
+        })}>
+        <input onChange={this.handleChange} id={id} className={styles.toggle} type="checkbox" checked={expanded} />
+        <label
+          htmlFor={id}
+          className={classnames({
+            [styles.toggleLink]: true,
+            [styles.toggleLabel]: true
+          })}>
```
michaeltaranto: If there are no conditional classes here it may read simpler as:
```diff
-          className={classnames({
-            [styles.toggleLink]: true,
-            [styles.toggleLabel]: true
-          })}>
+          className={classnames(styles.toggleLink, styles.toggleLabel)}>
```

react/AccordionItem/AccordionItem.less
```diff
@@ -0,0 +1,40 @@
...
+  opacity: 0;
+}
+
+.toggleLink {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  color: @sk-link;
+  cursor: pointer;
+  min-height: (@grid-row-height * 4);
```
michaeltaranto: Although the value is the same, please use `@row-height` as `@grid-row-height` is a legacy of the old grid system that we are hoping to retire at some stage.

react/AccordionItem/AccordionItem.test.js
```diff
@@ -0,0 +1,34 @@
...
+});
+
+test('should render expanded when prop changes', () => {
+  jest.spyOn(AccordionItem.prototype, 'componentWillReceiveProps');
+  const accordion = renderAccordion();
+  accordion.setProps({ expanded: true });
+  expect(AccordionItem.prototype.componentWillReceiveProps.mock.calls.length).toBe(1);
+  expect(accordion.find({ type: 'checkbox' }).props().checked).toBe(true);
+  expect(accordion).toMatchSnapshot();
+});
```
michaeltaranto: Be great to add a test to ensure the provided `onClick` handler is called too.

react/AccordionItem/AccordionItem.demo.js
```diff
@@ -0,0 +1,65 @@
...
+
+export default {
+  route: '/accordionitem',
+  title: 'Accordion Item',
+  category: 'Layout',
+  component: AccordionItem,
+  container: AccordionContainer,
+  block: true,
+  initialProps: {
+    title: <Text>Living Style Guide</Text>,
```
michaeltaranto: This should be vertically centred rather than on grid.
```suggestion
    title: <Text baseline={false} raw>Living Style Guide</Text>,
```

PR #570, opened 2018-10-30
fix(Radio): Set min-width to stop svg from being squished

dborysiewicz:
Add min-width to `svg` so that is never goes below @circumference which in turns squishes it and makes it oval.

Before:
screen shot 2018-10-30 at 3 43 27 pm

After:
screen shot 2018-10-30 at 3 46 20 pm

### DIFF
--- a/react/Radio/Radio.less
+++ b/react/Radio/Radio.less
@@ -36,6 +36,7 @@
 .svg {
   background-color: @sk-white;
   width: @circumference;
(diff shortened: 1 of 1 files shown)


michaeltaranto:
Are we deliberately reducing the height of the each radio component? While the touchable area is 5 rows high, the additional height previously maintained the 8 row interaction height.

radio

/cc @stephaniemoss

PR #571, opened 2018-10-31
fix: Change tertiaryLabel prop type to Object

trevorgk:
The `tertiaryLabel` flow type of the `TextField` component was string, but this was not reflecting the use case. 

It is useful for consumers to be able to include a link (or potentially any other content) in this section. In the case of the Advertiser sign-in portal, we are including a forgot password link.
image

I have updated the `TextField` component flow type and also discovered that the prop type was missing in other components and included them in this PR as well.
* After discussion we are not including this change, the reason being:
> there was a consensus at the time to NOT include child component prop checks  in the parent component (like tertiaryLabel).

### DIFF
--- a/react/TextField/TextField.js
+++ b/react/TextField/TextField.js
@@ -42,7 +42,7 @@ type FieldLabelProps = {
   label?: string,
   labelProps?: Object,
   secondaryLabel?: string,
-  tertiaryLabel?: string
+  tertiaryLabel?: React$Node
 };
 
 type FieldMessageProps = {


react/TextField/TextField.js
```diff
@@ -42,7 +42,7 @@ type FieldLabelProps = {
   label?: string,
   labelProps?: Object,
   secondaryLabel?: string,
-  tertiaryLabel?: string
+  tertiaryLabel?: Object
```
michaeltaranto: I would recommend using `React$Node` here

trevorgk: Good plan

PR #572, opened 2018-11-01
fix(Highlight): Fix undefined class being assigned to element

kosanna:
Due to the `[styles[tone]]: true` defaulting tone to `TONE.NEUTRAL`, by default we always get `undefined` class being assigned to the element.

This is caused by the fact that `.neutral` class does not exist in the stylesheet.

Tone is NOT a required prop from the consumer's perspective and also defaulting is a) doing nothing, b) introducing incorrect behaviour of the component.


kosanna:
Thank you nice people!

PR #573, opened 2018-11-13
fix(Header): Disable event listener on input fields

bjervisSEEK:
When adding the searchbar keyboard shortcuts, I broke the sandbox by removing the ability to type `/`. The event listener now checks that the event wasn't called on an `<input>` or `<textarea>` field when deciding whether to open the menu.

### DIFF
--- a/docs/src/components/App/Header/Header.js
+++ b/docs/src/components/App/Header/Header.js
@@ -60,7 +60,7 @@ class Header extends Component {
   };
 
   handleDocumentKeydown = event => {
-    const isInputField = event.target.tagName.match(/INPUT|TEXTAREA/i) !== null;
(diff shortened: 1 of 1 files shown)


docs/src/components/App/Header/Header.js
```diff
@@ -60,7 +60,9 @@ class Header extends Component {
   };
 
   handleDocumentKeydown = event => {
-    if (event.key === '/' && event.target.id !== SEARCH_BAR_ID) {
+    const isInputField =
+      ['INPUT', 'TEXTAREA'].indexOf(event.target.tagName) !== -1;
```
michaeltaranto: Personal preference maybe, but I'd prefer to see this as a case insensitive regex.

PR #577, opened 2018-11-20
fix(Footer): Add in new native app download links

adenj:
Added download links for iOS and Android into footer, and also fixed the Volunteering link on the NZ
site that was pointed to a non-SEEK url.

### DIFF
--- a/react/Footer/Footer.js
+++ b/react/Footer/Footer.js
@@ -64,7 +64,7 @@ export default class Footer extends Component {
             <div className={styles.columns}>
               <FooterNav label="Tools">
                 { tools.map(this.renderLink) }
-                <ToggleContainer name="DownloadAppsToggle" label="Download Apps">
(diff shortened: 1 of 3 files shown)


react/Footer/Footer.js
```diff
@@ -64,6 +64,9 @@ export default class Footer extends Component {
             <div className={styles.columns}>
               <FooterNav label="Tools">
                 { tools.map(this.renderLink) }
+                <ToggleContainer name="DownloadAppsToggle" label="Download Apps">
```
mattsjones: Should the label be `Download apps`?

adenj: Thats the label Socrates requested they be under, happy to change it if you have a more appropriate name though

adenj: Oh never mind. Just saw what you meant 🤦‍♂️

PR #590, opened 2018-12-07
fix(ToggleContent): Make ToggleContent look identical to ReadMore

kosanna:
Currently ToggleContent and ReadMore have inconsistent look. This PR makes them look the same.


react/ToggleContent/ToggleContent.less
```diff
@@ -0,0 +1,5 @@
+@import (reference) '~seek-style-guide/theme';
+
+.button {
+  box-shadow: none !important;
```
georgespyropoulos: Looks like box shadow has already been set to none. Do we need to declare it again?

screen shot 2018-12-17 at 9 47 19 am

kosanna: we need to overwrite focus state coming from the button, I'll just do that directly

georgespyropoulos: I'm confused - why would we want to override that style? If we turn off the box shadow there's no way to tell your focussed on the button when using the page with a keyboard

georgespyropoulos: But that breaks keyboard accessibility of the button. Why do you want to reset it?

PR #594, opened 2019-01-14
feat(ts): Add type definitions for Button and ErrorIcon

chardos:
Going to start the process of creating type definitions for components in preparation for more typescript consumers coming into existence. Also, without definitions, typescript is inferring types from the propTypes (incorrectly) and is causing linting errors.


chardos:
Need to do some testing on this with a flow consumer. Will close for now.

PR #599, opened 2019-02-07
feat(ReadMore): Add ability to access state of more button visiblity

georgespyropoulos:
There's currently no way to know if the more button is being displayed or not. I've got an analytics
requirement to find out if users are reading the text so am adding this prop to expose whether the
button is being displayed

### DIFF
--- a/react/ReadMore/ReadMore.js
+++ b/react/ReadMore/ReadMore.js
@@ -30,7 +30,8 @@ type Props = {|
   moreLabel: string,
   lessLabel: string,
   backgroundComponentName: 'card' | 'body' | 'gray-lightest',
-  onShowMore?: Function
+  onShowMore?: Function,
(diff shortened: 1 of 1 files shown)


react/ReadMore/ReadMore.js
```diff
@@ -30,7 +30,8 @@ type Props = {|
   moreLabel: string,
   lessLabel: string,
   backgroundComponentName: 'card' | 'body' | 'gray-lightest',
-  onShowMore?: Function
+  onShowMore?: Function,
+  onMoreButtonVisibilityChange?: Function
```
chardos: Are there other ways for visibility to change other than clicking the more button? Would `onVisibilityChange` suffice?

chardos: I see what the hook is for now. Disregard this :)

PR #600, opened 2019-02-08
docs: Add palette guide

tgdev:
Whenever working on projects that consume the style guide, it is difficult to remember what colours are available. Up til now the only way to find out is to dig through the source code on github or locally via code editor.

This PR adds a new page to the "Guide" section, outputting every colour from within the LESS files inside `theme/palette` grouped by said files and displaying;
- colour swatch
- hex/value
- variable name

_**NOTE: Some colour values are either other LESS variables or LESS functions (eg: fade) and therefore do not output a colour swatch. This is mainly within `elements.less` but happy to get some help on how this might be rectified**_

### New Nav item
screen shot 2019-02-09 at 10 01 07 am

### New Page
ssg-palette-guide


docs/src/components/Palette/Palette.less
```diff
@@ -0,0 +1,21 @@
+@import (reference) '../../../../theme';
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.item {
+  flex: 0 0 168px;
```
michaeltaranto: Not too fussed about the magic number on the docs site, but have you confirmed this number is ok on mobile?

docs/src/components/App/Header/Header.js
```diff
@@ -20,13 +20,14 @@ import { generateRouteList } from './headerUtils';
...
   { route: '/typography', title: 'Typography', category: 'Guides' },
   { route: '/page-layout', title: 'Page Layout', category: 'Guides' },
   {
     route:
-      process.env.NODE_ENV === 'production' ?
-        `${process.env.BASE_HREF}playroom` :
-        'http://localhost:9000/',
+      process.env.NODE_ENV === 'production'
+        ? `${process.env.BASE_HREF}playroom`
+        : 'http://localhost:9000/',
```
michaeltaranto: Looks like a fight with prettier, maybe revert the formatting changes, and we should prettier-ify this repo in another PR as eslint isn't happy

docs/src/components/Palette/Palette.js
```diff
@@ -0,0 +1,69 @@
...
+  { name: 'accessible-variants.less', variables: lessToJs(accessibleVariants) }
+];
+
+export default () => (
+  <div>
+    <PageBlock>
+      <Section header>
+        <Text hero>Palette</Text>
+      </Section>
+      <Card transparent style={{ maxWidth: 720 }}>
```
georgespyropoulos: Perhaps this style could go in `Palette.less`. Or you could use `AsidedLayout` to wrap the card

docs/src/components/Palette/Palette.js
```diff
@@ -0,0 +1,69 @@
...
+import accessibleVariants from '!!raw-loader!../../../../theme/palette/accessible-variants.less';
+
+import styles from './Palette.less';
+
+const palettes = [
+  { name: 'brand.less', variables: lessToJs(brand) },
+  { name: 'grays.less', variables: lessToJs(grays) },
+  { name: 'elements.less', variables: lessToJs(elements) },
+  { name: 'partners.less', variables: lessToJs(partners) },
+  { name: 'accessible-variants.less', variables: lessToJs(accessibleVariants) }
```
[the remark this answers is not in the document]
tgdev: 👍 Reason for this initial approach was to let consumers know where they could find each variable but I guess there's no reason why they'd need to know.

docs/src/components/Palette/Palette.less
tgdev: Good call. Will double check.

docs/src/components/Palette/Palette.js
```diff
@@ -0,0 +1,69 @@
...
+        </Section>
+      </Card>
+      {palettes.map((palette, index) => (
+        <Card transparent key={`palette-${index}`}>
+          <Section>
+            <Text heading>{palette.name}</Text>
+          </Section>
+          <div className={styles.container}>
+            {Object.keys(palette.variables).map((pvars, i) => {
+              const hexValue = palette.variables[pvars];
```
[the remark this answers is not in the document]
tgdev: Could we possibly pair on this @michaeltaranto? I knew there'd be a way to handle this but not sure on exactly how.

docs/src/components/Palette/Palette.js
tgdev: This inline style was copied from another, existing guide page (Typography). We might need to look at applying a consistent fix across all pages which I'd prefer was done in another PR.

docs/src/components/App/Header/Header.js
tgdev: Going to wait for @jahredhope 's PR to add prettier to SSG.

michaeltaranto:
Just thinking, this should really `docs` not `feat`, consumers dont need a new version from npm. Please update before merging

PR #606, opened 2019-03-12
 feat(Header): Add Recommended Jobs to user account menu

terenceduong:
### Changes:
- Added (un-hide)`Recommended Jobs` link (to desktop).
- Remove the gray background behind Saved Search / Saved Jobs / Applied Jobs.
- Change on-hover colour to @sk-mid-gray-light.
- Add separator above Settings / Sign Out group.

The menu is wider now but saved / applied jobs will become saved & applied jobs (single item) in the future too.

### Before:
Before

### After:
After

### After (on hover):
image

### DIFF
--- a/react/Header/UserAccountMenu/UserAccountMenu.js
+++ b/react/Header/UserAccountMenu/UserAccountMenu.js
@@ -96,7 +96,6 @@ export default ({
         ]
       })}
     </Hidden>
-
     <li className={classnames(activeTab === PROFILE && styles.activeTab)}>
       {linkRenderer({
         'data-analytics': 'header:profile',
@@ -113,13 +112,12 @@ export default ({
         ]
       })}
     </li>
-
     <li
       className={classnames(activeTab === SAVED_SEARCHES && styles.activeTab)}
     >
       {linkRenderer({
         'data-analytics': 'header:saved+searches',
-        className: `${styles.item} ${styles.subItem}`,
+        className: styles.item,
         href: '/my-activity/saved-searches',
         children: [
           <span key="label">{SAVED_SEARCHES}</span>,
@@ -131,7 +129,6 @@ export default ({
         ]
       })}
     </li>
-
     <li
       className={classnames(
         activeTab === SAVED_AND_APPLIED && styles.activeTab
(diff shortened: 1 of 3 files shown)


react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -61,10 +61,11 @@
...
 
-.subItem {
-  @media @desktop {
-    background-color: @sk-gray-light;
-  }
+.menuSeparator {
+  height: 1px;
+  border: 0;
+  background-color: @sk-mid-gray-light;
+  margin: 5px 8px;
```
markdalgleish: This throws it off grid. These should be a multiple of the grid row, minus 1 to account for element the height.

terenceduong: Changed these to comply with the grid

react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -61,10 +61,11 @@
...
 
-.subItem {
-  @media @desktop {
-    background-color: @sk-gray-light;
-  }
+.menuSeparator {
+  height: 1px;
+  border: 0;
+  background-color: @sk-mid-gray-light;
+  margin: @row-height @row-height (@row-height - 1);
```
michaeltaranto: Two minor points here:
1. I suggest extracting the `height` value to something like `@separator-height` and using it in this calculation to make the code more self documenting.
2. The horizontal margin dimension should not be a factor of rows. Typically we use a variation of the `@gutter-width`, eg:
```less
margin: @row-height (@gutter-width / 2) @separator-height
```

react/Header/UserAccountMenu/UserAccountMenu.js
```diff
@@ -190,8 +183,8 @@ export default ({
           />
         ]
       })}
-    </Hidden>
-
+    </li>
+    <hr className={styles.menuSeparator} />
```
michaeltaranto: I think the separator should only be rendered on desktop as the separation to "Company Reviews" looks odd to me. Does that sound right to you?
Screen Shot 2019-03-13 at 2 11 48 pm

terenceduong: Oops forgot to check the mobile menu. This change should definitely only be on desktop, cheers!
#mobilefirst4life

PR #608, opened 2019-03-21
feat(Header): Add icon to Applied Jobs (UserAccountMenu)

terenceduong:
### (Desktop) User Account Menu
**Add icon to Applied Jobs**
We have decided against merging the `Saved Jobs` and `Applied Jobs` user account menu items.
Instead we are fixing the current awkwardness which is `Applied Jobs` not having its own icon.

This change adds a `TickCircleIcon`, with the colour of `@sk-green`.

#### Before
Applied Jobs without an icon

#### After
Applied Jobs with an icon

### DIFF
--- a/react/Header/UserAccountMenu/UserAccountMenu.less
+++ b/react/Header/UserAccountMenu/UserAccountMenu.less
@@ -99,6 +99,10 @@
   color: @sk-green-light;
 }
 
+.icon.appliedJobs {
+  color: @sk-green;
+}
+
 .crSvg {
   width: 14px;
   height: 14px;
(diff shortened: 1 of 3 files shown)


react/Header/UserAccountMenu/UserAccountMenu.less
```diff
@@ -99,6 +99,10 @@
   color: @sk-green-light;
 }
 
+.icon.appliedJobs {
+  color: @sk-green;
```
michaeltaranto: When you have applied for a job and it appears in the search results page, the green used on the applied tick is `@sk-green-light`, feel like we should align these. Thought I should bring that to your attention.

PR #610, opened 2019-03-29
fix(Checkbox): Label over two lines should display like the Radio Button

dejanvasic85:
When a label text in a checkbox spans over multiple lines, there is a massive gap between the lines. See screenshot. This is because of the line-height coming from `touchableText`.

Before Change:
Screen Shot 2019-03-29 at 10 06 19 am

After Change:
Screen Shot 2019-03-29 at 10 06 35 am

### DIFF
--- a/react/Checkbox/Checkbox.js
+++ b/react/Checkbox/Checkbox.js
@@ -73,7 +73,7 @@ export default class Checkbox extends Component {
             className={styles.checkMark}
           />
         </div>
-        <span>{label}</span>
+        <span className={styles.checkboxLabel}>{label}</span>
       </div>
     );
   }
--- a/react/Checkbox/Checkbox.less
+++ b/react/Checkbox/Checkbox.less
@@ -34,6 +34,10 @@
   display: flex;
 }
 
+.checkboxLabel {
+  margin-top: (@row-height / 2);
+}
+
 .checkbox {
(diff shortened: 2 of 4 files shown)


patrickliu-sa:
We tweaked the Checkbox style in seek-asia-style-guide:
image

so the checkbox align to first line of label only 😃

dejanvasic85:
@patrickliu-sa I'm assuming then that you did an override with your own react component?

dejanvasic85:
This should be ready for review again @michaeltaranto. Agree with the consistency to the radio button so I've removed the `align:center` and added margin for the span. 

Screen Shot 2019-04-02 at 2 13 01 pm

patrickliu-sa:
@dejanvasic85 The component was coming from SEEK style guide at the beginning. Here's the PR that we tried to solve the vertical height problem:
https://github.com/seekinternational/seek-asia-style-guide/pull/309

But seems your radio button approach is more elegant :)

michaeltaranto:
Just noticing there are a few pre-existing things about `Checkbox` that are not up to the standard of `Radio`. Talked to @dejanvasic85 and i'll take this PR over and bring this component up to the same level.

michaeltaranto:
Given the scope of changes grew on my branch, I will close this and open a separate PR rather than try and merge the two branches.

PR #612, opened 2019-04-01
feat(Header): Add applied jobs icon (desktop)

terenceduong:
### Changes
User Account Menu looks a bit strange containing most icons, but not the Applied Jobs icon.
We are adding an icon in with a colour value (`00b600`) that matches the Job Applied Success page's tick.

Also, after the addition of the Applied Jobs icon we have realised that this icon looks too similar in colour to the Recommended Jobs icon. 
We are changing the Recommended Jobs icon to `@sk-green` - the same colour as it is displayed on the dashboard panel.

#### Before
Before icon has been added to Applied Jobs

#### After - adding Applied Jobs icon
After icon has been added to Applied Jobs

#### After after - changing Recommended Jobs icon to `@sk-green` (matching its colour on the dashboard)
After Recommended Jobs has its colour changed
image


michaeltaranto:
I would recommend exposing these colours to consumers to be used semantically in order to maintain consistency across the user experience. We already have this concept contained in the [elements](https://github.com/seek-oss/seek-style-guide/blob/master/theme/palette/elements.less) area of our palette. I'll push a minor refactor, that way consumers can cut over the the centralised variables and stay in sync.

PR #613, opened 2019-04-03
fix(Bolding): Change Readmore and ghost button font weight

georgespyropoulos:
The `ghost` variant of the button component is the only one that has bold text. I'm reducing the font weight to` @sk-medium` to make it consistent with the other buttons.

I'm also changing the font weight of the `ReadMore` button to `@sk-medium` so it's consistent with other buttons and the `ToggleContent` component.

### DIFF
--- a/react/ReadMore/ReadMore.js
+++ b/react/ReadMore/ReadMore.js
@@ -5,8 +5,8 @@ import React, { PureComponent } from 'react';
 import classnames from 'classnames';
 
 import themeVars from '../private/themeVars';
+import Text from '../Text/Text';
 import TextLink from '../TextLink/TextLink';
-import Button from '../Button/Button';
 
 const { rowHeight, standardTypeRowSpan, interactionTypeRowSpan } = themeVars;
 const buttonHeight = rowHeight * interactionTypeRowSpan;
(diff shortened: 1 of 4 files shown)


react/ReadMore/ReadMore.js
```diff
@@ -135,15 +135,17 @@ class ReadMore extends PureComponent<Props, State> {
...
         {showMoreLessButton && (
-          <Text
+          <TextLink
             strong
+            color="transparent"
+            type="button"
             chevron={showMore ? 'up' : 'down'}
             className={styles.showMore}
-            component={TextLink}
+            component={Button}
```
michaeltaranto: Confirming that the desired effect is that this render a `button` rather than an `anchor` element? Also surprised that `type="button"` wasn't the default in our `Button` component — learn something new everyday.

georgespyropoulos: So what I didn't realise here was that the hit area spanned the entire width of the text previously. My change here has made the More link button shaped which I thought was an improvement but actually makes the hit area smaller. Going to roll this back and manually apply the font-weight change

Screen Shot 2019-04-04 at 12 02 30 pm

PR #617, opened 2019-04-07
fix(Textarea): Remove max-height

dborysiewicz:
We require a `<Textarea />` with a height greater than 180px. While it is possible to override it using a CSS selector (something like .className textarea) it becomes extremely fragile when having highlighting enabled as it also inherits everything from `styles.textarea`.

### DIFF
--- a/react/Textarea/Textarea.demo.js
+++ b/react/Textarea/Textarea.demo.js
@@ -71,7 +71,7 @@ export default {
           })
         },
         {
-          label: 'Count',
+          label: 'Show Count',
           transformProps: props => ({
             ...props,
             countFeedback: value => ({
@@ -87,10 +87,10 @@ export default {
           })
         },
         {
-          label: 'Inital Rows 20',
+          label: 'Initial Rows 20',
           transformProps: props => ({
             ...props,
-            initalRows: 10
+            initialRows: 20
           })
         },
         {
@@ -113,7 +113,7 @@ export default {
           })
         },
         {
-          label: 'Disallowed words',
+          label: 'Disallowed word(s)',
           transformProps: props => ({
             ...props,
             invalidText: 'bad',
@@ -132,7 +132,7 @@ export default {
           })
         },
         {
-          label: 'Multi invalid range',
+          label: 'Multiple invalid ranges',
           transformProps: props => ({
             ...props,
             invalidText: [
--- a/react/Textarea/Textarea.demo.js
+++ b/react/Textarea/Textarea.demo.js
@@ -71,7 +71,7 @@ export default {
           })
         },
         {
-          label: 'Show Count',
+          label: 'Count',
           transformProps: props => ({
             ...props,
             countFeedback: value => ({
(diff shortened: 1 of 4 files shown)


dborysiewicz:
@georgespyropoulos Cover letter has a limit of 5000 characters and forcing a user to use the scrollbar has been raised as a ux issue. We are adding logic to make the text area autosize in our app to fix this and max-height is blocking it.

georgespyropoulos:
Should it auto-size to any height? I think the max-height was here to stop the box growing indefinitely. Should we pick a higher max height? Or add a prop to make this configureable? This could be considered a breaking change by some people that use the component

chardos:
+1 for using a prop. I'm assuming `TextArea` is used everywhere, so we can't just remove the max-height without potentially breaking a bunch of UI's

react/Textarea/Textarea.js
```diff
@@ -9,6 +9,8 @@ import { TONE } from '../private/tone';
 import { formatInvalidText } from './textAreaUtils';
 import attachRefs from '../private/attachRefs';
 
+const rowHeight = 6;
```
mattsjones: Maybe chuck a comment on this saying why it's being hardcoded?

react/Textarea/Textarea.js
```diff
@@ -32,6 +34,8 @@ export default class Textarea extends Component {
     valid: PropTypes.bool,
     description: PropTypes.string,
     inputProps: PropTypes.object,
+    initalRows: PropTypes.number,
```
mattsjones: `initialRows`

react/Textarea/Textarea.demo.js
```diff
@@ -85,6 +85,20 @@ export default {
             ...props,
             description: 'Describe a descriptive description descriptively'
           })
+        },
+        {
+          label: 'Inital Rows 20',
+          transformProps: props => ({
+            ...props,
+            initalRows: 10
```
mattsjones: Is this supposed to be `20`?

react/Textarea/Textarea.demo.js
```diff
@@ -71,7 +71,7 @@ export default {
           })
         },
         {
-          label: 'Show Count',
+          label: 'Count',
```
mattsjones: Did you mean to change this?

georgespyropoulos: I'm assuming you did this to make everything fit on one row on the demo page? Could we try adding `flex-wrap: wrap` to the container that wraps the options on the demo page instead?

react/Textarea/Textarea.js
```diff
@@ -71,7 +75,9 @@ export default class Textarea extends Component {
   static defaultProps = {
     className: '',
     description: '',
-    inputProps: {}
+    inputProps: {},
+    initalRows: 15,
```
georgespyropoulos: (note there's a typo here as well inital rather than initial)

```suggestion
    initialGridRows: 15,
```

react/Textarea/Textarea.js
```diff
@@ -71,7 +75,9 @@ export default class Textarea extends Component {
...
+    maxRows: 30
```
georgespyropoulos:
```suggestion
    maxGridRows: 30
```

PR #619, opened 2019-04-26
AccordionItem

chardos:
Addition of `AccordionItem` to the style guide.
- Has `onOpen` and `onClose` hooks
- Animates opening and closing
- Accepts title as string (will style the default text), or a react element (no styling)

sdf


dejanvasic85:
I was looking for an accordion! We had to implement our own for talent search:
https://talentsearch.seek.com.au/v2/help

@chardos , I noticed that your chevron is not flipping to the side when open an item. Could you add this to mimic our behavior so that we could replace ours in the future?

react/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,105 @@
...
+        }}
+        {...restProps}
+      >
+        {typeof title === 'string' ? (
+          <span className={styles.titleText}>{title}</span>
+        ) : (
+          <span className="AccordionItem__titleContent">{title}</span>
+        )}
+        <ChevronIcon
+          direction="down"
```
jahredhope: Would it be possible to use the direction prop here to control the rotation?

chardos: @jahredhope (see above animated gif for reference)

chardos: Nvm, @adenj has added a commit, that works well.

react/AccordionItem/AccordionItem.less
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+
+.chevron {
+  margin-left: 9px;
+  transition: transform 0.3s ease;
+  color: @sk-blue-lighter;
+}
+
+.chevronOpen {
```
jahredhope: This can be removed

react/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,102 @@
...
+          if (isOpen && onClose) {
+            onClose();
+          } else if (!isOpen && onOpen) {
+            onOpen();
+          }
+        }}
+        {...restProps}
+      >
+        {typeof title === 'string' ? (
+          <span className={styles.titleText}>{title}</span>
```
jahredhope: Could this leverage the `<Text />` or `<TextLink />` components rather than use it's own styles?

chardos: I did consider that, but I'd prefer not to nest an anchor tag inside of a button?

chardos: As discussed, have also tried replacing the button element with `<Button transparent />`, but it takes on the "bounce" effect, and other styling related to that component. Will be simpler to keep current styling rather than writing overrides.

PR #620, opened 2019-05-01
 feat(Footer): Add RSP link to footer

lgshaw:
A new page promoting our Recruitment Software Partners has launched and needs to be surfaced in the candidate footer.

### DIFF
--- a/react/Footer/data/employers.js
+++ b/react/Footer/data/employers.js
@@ -67,4 +67,10 @@ export default [
     analytics: 'toolbar:integration-partners',
     specificLocale: 'AU'
   },
+  {
+    name: 'Recruitment software partners',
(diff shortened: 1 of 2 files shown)


react/Footer/data/employers.js
```diff
@@ -60,5 +60,17 @@ export default [
...
+    href: 'https://talent.seek.com.au/integration-partners/',
+    analytics: 'toolbar:integration-partners',
+    specificLocale: 'AU'
+  },
+  {
+    name: 'Recruitment software partners',
+    href: 'https://talent.seek.co.nz/integration-partners/',
+    analytics: 'toolbar:integration-partners',
+    specificLocale: 'NZ'
+  },
```
jahredhope: I don't think we use trailing commas in this repo.
Prettier isn't happy.

PR #621, opened 2019-05-09
feat(Header): Add support for notification bell

chardos:
This PR adds a link in the header which links to the new notifications page http://www.seek.com.au/notifications (Please note the API serving this page is currently down and is being fixed at the moment). Using AB crunch the link will display to 10% of users.

The first iteration will always display the link, regardless if the user has new notifications. Later iterations will display a red bubble with the number of new notifications displayed inside.

SCREENSHOTS:

image

image


react/Header/NotificationsLink/NotificationsLink.js
```diff
@@ -0,0 +1,54 @@
...
+          .then((response) => {
+            const isInExperiment = Boolean(response.experiments[EXPERIMENT_ID]);
+
+            if (isInExperiment) {
+              this.setState({isInExperiment: true})
+            }
+          });
+      }
+    } catch (e) {
+      console.log(e)
```
georgespyropoulos: Is there need to log an error? I reckon we can just let this silently fail

react/Header/NotificationsLink/NotificationsLink.js
```diff
@@ -0,0 +1,54 @@
...
+import { getCookie } from './utils';
+import { AUTHENTICATED } from '../../private/authStatusTypes';
+
+class NotificationsLink extends React.Component {
+  state = {
+    isInExperiment: false
+  }
+
+  componentDidMount() {
+    const EXPERIMENT_ID = '58CC4E12-A4E0-4C0B-BD2A-681FD98FDC2E';
```
georgespyropoulos: These consts could be moved outside the component

react/Header/NotificationsLink/NotificationsLink.js
```diff
@@ -0,0 +1,54 @@
+import React, { Fragment } from 'react';
```
georgespyropoulos: What about writing a test for this file?

react/Header/NotificationsLink/utils.js
```diff
@@ -0,0 +1,8 @@
+export function getCookie(name) {
```
georgespyropoulos: Should we add a basic test for this function?

.prettierignore
```diff
@@ -3,4 +3,4 @@ dist
 node_modules
 *.log
 yarn.lock
-flow-typed/*.generated.*
\ No newline at end of file
+flow-typed/*.generated.*
```
mattsjones: Not sure why this changed?

react/Header/NotificationsLink/NotificationsLink.js
```diff
@@ -0,0 +1,54 @@
...
+import { AUTHENTICATED } from '../../private/authStatusTypes';
+
+class NotificationsLink extends React.Component {
+  state = {
+    isInExperiment: false
+  }
+
+  componentDidMount() {
+    const EXPERIMENT_ID = '58CC4E12-A4E0-4C0B-BD2A-681FD98FDC2E';
+    const visitorId = getCookie('JobseekerVisitorId');
```
mattsjones: Can we move the `getCookie` call into the try/catch just in case?

react/Header/NotificationsLink/NotificationsLink.js
```diff
@@ -0,0 +1,54 @@
...
+    const url = `http://experiments-api.candidate.prod.outfra.xyz/participants/${visitorId}`;
```
mattsjones: This URL is not `https`. It'll bork on the site. You'll need to use cloud.seek.com.au one I think.

react/BellIcon/BellIcon.js
```diff
@@ -0,0 +1,11 @@
+import markup from './BellIcon.svg';
+import PropTypes from 'prop-types';
```
michaeltaranto: PropTypes does not seem to be used here.

react/Header/Header.less
```diff
@@ -153,3 +154,9 @@
     display: inline-block;
   }
 }
+
+.bellWrapper {
+  display: inline-block;
+  margin-top: 15px;
```
michaeltaranto: Is this margin still necessary even with the flex centering above?

react/Header/NotificationsLink/NotificationsLink.js
michaeltaranto: Agree the `experimentId` and the `url` can be hoisted out, but the `visitorId` should remain.

react/Header/NotificationsLink/NotificationsLink.less
```diff
@@ -0,0 +1,24 @@
+@import (reference) "~seek-style-guide/theme";
+@import (reference) "../Header.less";
+
+.bell {
+  padding: 15px 0;
+  position: relative;
+  top: 1px;
+  color: @sk-pink;
```
michaeltaranto: How married to the pink are we for this? Apart from the contrast on mobile being a little challenging, I feel like it would look much more refined to follow the text colour, eg. black on desktop and white on mobile. Thoughts?

.prettierignore
chardos: It's a newline, my editor auto adds by default

react/BellIcon/BellIcon.js
chardos: Removed

.prettierignore
chardos: same as above

react/Header/NotificationsLink/NotificationsLink.js
chardos: Done

react/Header/Header.less
chardos: Good point, added that margin before adding the vertical centering. Removed.

react/Header/NotificationsLink/NotificationsLink.less
chardos: 100% agree, this was a design call made by Carmelia. I think since she's not in, let's just flip it to black on desktop, white on mobile.

.prettierignore
chardos: Removed

react/Header/NotificationsLink/utils.test.js
```diff
@@ -0,0 +1,15 @@
...
+    const visitorId = '23fba469f32dfd44602bcda9c520bed7';
+    const visitorIdFromCookie = getCookieFromString(
+      'JobseekerVisitorId',
+      cookieString
+    );
+    console.log('visitorIdFromCookie', visitorIdFromCookie);
```
michaeltaranto: Can you remove this console log please. Also i'd consider cleaning up that cookie as there is a lot of info in there that doesnt need to be in the codebase.

react/Header/NotificationsLink/utils.test.js
```diff
@@ -0,0 +1,15 @@
...
+    expect(visitorId).toEqual(visitorIdFromCookie);
+  });
```
michaeltaranto: Can we also capture the unhappy paths, returning `null` and fewer string parts than expected etc.

react/Header/NotificationsLink/NotificationsLink.less
```diff
@@ -0,0 +1,25 @@
+@import (reference) "~seek-style-guide/theme";
+@import (reference) "../Header.less";
+
+.bell {
+  padding: 15px 0;
```
michaeltaranto: This value seems odd, feels like the bell should be vertically centered or have it's height calculated from rows, eg. touchable height or similar. Thoughts?

react/Header/NotificationsLink/NotificationsLink.less
chardos: Confirmed with Carmelia this change is fine to roll with.

react/Header/NotificationsLink/NotificationsLinkContainer.js
```diff
@@ -0,0 +1,57 @@
...
+        fetch(url)
+          .then(response => response.json())
+          .then(response => {
+            const isInExperiment = Boolean(response.experiments[EXPERIMENT_ID]);
+
+            if (isInExperiment) {
+              this.setState({ isInExperiment: true });
+            }
+          })
+          .catch(() => {});
```
jahredhope: There are two catches here that swallow errors. Why are they necessary? And is there any way to document this in code... perhaps just an inline comment?

react/Header/NotificationsLink/NotificationsLinkContainer.js
```diff
@@ -0,0 +1,57 @@
...
+      }
+      // eslint-disable-next-line no-empty
+    } catch (e) {}
+  }
+
+  render() {
+    const { authenticationStatus, linkRenderer } = this.props;
```
jahredhope: Just a idea: I'm liking hooks for this. But that might be the first hook in seek-style-guide requiring us to up the peer-dependency.

```js
const [isInExperiment, setIsInExperiment] = useState(false);
useEffect(() => {
  fetchIsInExperiment().then(setIsInExperiment)
}, [])
```

Perhaps even a `useIsInExperiment(visitorId, url)` hook.

georgespyropoulos: This is a good point... we will need to mark this as a breaking change if we introduce hooks here because we currently only require react 16.3. And this will slow down some teams in updating their headers

react/Header/NotificationsLink/NotificationsLinkContainer.js
mattsjones: The swallowing is just to hide the errors from product teams as there's nothing to worry about from their perspective. The reason for two is because synchronous try catch don't catch promise based exceptions (without async/await).

react/Header/NotificationsLink/NotificationsLinkContainer.js
chardos: Love the idea. Would prefer to do this at a later stage though, in favour of getting this live quicker. This feature needs a follow up PR anyway, to remove the segmentation/feature toggle call.

PR #625, opened 2019-05-20
fix(Pill): Make close button a type of button

chardos:
Currently, when pills exist inside of a form, clicking on the close button will trigger a submit of the form. 

As I can't think of any scenario where closing a pill should trigger a submit, this PR changes the close buttons `type` to be `button`, rather than the default of `submit`.

image

### DIFF
--- a/react/Pill/Pill.js
+++ b/react/Pill/Pill.js
@@ -8,6 +8,7 @@ import ScreenReaderOnly from '../ScreenReaderOnly/ScreenReaderOnly';
 
 type Props = {
   children?: React$Node,
+  buttonType?: string,
   text?: React$Node,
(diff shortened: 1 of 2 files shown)


askoufis:
The skills component on profile relies on this "feature". We don't have a cancel or undo button, we just submit on every skill add/delete. Thoughts @georgespyropoulos ?

chardos:
> The skills component on profile relies on this "feature". We don't have a cancel or undo button, the submit on skill add/delete kind of prevents it. Thoughts @georgespyropoulos ?

As discussed in person with @askoufis and @georgespyropoulos , this change does not affect their use case, as they rely on an `onClick` handler

PR #627, opened 2019-05-23
feat(AccordionItem): Allow controlled component

chardos:
Update to  the `AccordionItem` to allow external control via the `isOpen` prop. If the `isOpen` prop is passed, internal state will no longer be used to determine open/closed state.

### DIFF
--- a/react/AccordionItem/AccordionItem.js
+++ b/react/AccordionItem/AccordionItem.js
@@ -29,11 +29,10 @@ function AccordionItem({
   const [isOpen, setIsOpen] = useState(open);
   const useInternalState = (externalIsOpen === undefined);
   const finalIsOpen = useInternalState ? isOpen : externalIsOpen;
+  // const 
 
-  const prevIsOpen = usePrevious(finalIsOpen);
   useEffect(() => {
-    const needsUpdate = prevIsOpen !== finalIsOpen;
-    if (!useInternalState && needsUpdate) {
+    if (!useInternalState) {
       toggleContent({
         el: contentEl.current,
         setCurrentHeight,
@@ -45,7 +44,7 @@ function AccordionItem({
         setCssOverflow
       });
     }
-  });
+  }, [finalIsOpen]);
 
   const buttonClasses = classnames(className, styles.title);
   const isAnimating = currentHeight === CLOSED_HEIGHT && isOpen;
@@ -86,11 +85,11 @@ function AccordionItem({
         {...restProps}
       >
         {typeof title === 'string' ? (
-          <TextLink>
-            <Text raw subheading baseline={false} className={styles.titleText}>
+          <Text raw baseline={false} className={styles.titleText}>
+            <TextLink>
               {title}
-            </Text>
-          </TextLink>
+            </TextLink>
+          </Text>
         ) : (
           <span>{title}</span>
         )}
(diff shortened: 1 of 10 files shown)


react/AccordionItem/AccordionItem.test.js
```diff
@@ -62,15 +66,19 @@ describe('AccordionItem', () => {
...
       );
 
       const renderedElement = ReactTestUtils.renderIntoDocument(element);
       const button = renderedElement.querySelector('button');
 
       ReactTestUtils.Simulate.click(button);
-      expect(mockOnClose).toHaveBeenCalledTimes(1);
+
+      // wait for useEffect
+      requestAnimationFrame(() => {
```
grahamgoh: hmm isnt this asynchronous , dont u have to await this?

react/AccordionItem/AccordionItem.test.js
```diff
@@ -53,7 +53,11 @@ describe('AccordionItem', () => {
       const button = renderedElement.querySelector('button');
 
       ReactTestUtils.Simulate.click(button);
-      expect(mockOnOpen).toHaveBeenCalledTimes(1);
+
+      // wait for useEffect
+      requestAnimationFrame(() => {
+        expect(mockOnOpen).toHaveBeenCalledTimes(1);
```
grahamgoh: hmm isnt this asynchronous , dont u have to await this?
same as below

react/AccordionItem/AccordionItem.test.js
grahamgoh: so the test might be finishing before u even run this expect>

react/AccordionItem/AccordionItem.test.js
chardos: Good catch man. Added a `done()` call

grahamgoh: nice!

react/AccordionItem/AccordionItem.js
```diff
@@ -10,59 +10,65 @@ function AccordionItem({
...
+        setTimeoutHandle,
+        setCurrentHeight,
+        setCssOpacity,
+        setCssOverflow,
+        setCssVisibility
+      });
+    }
+
+    isMounting.current = false;
+  }, [finalIsOpen]);
```
grahamgoh: The usage of isMounting to control whether useEffect should be run feels like a code smell to me. Is there a way to do it without using the isMounting check?

grahamgoh: should timeoutHandle be part of the array `[finalIsOpen]` as it is only a state use inside the effect lambda?

chardos: > The usage of isMounting to control whether useEffect should be run feels like a code smell to me. Is there a way to do it without using the isMounting check?

Currently, I don't see a tidier way of doing this. In the future we could encapsulate it in a custom hook, maybe something like :

```
useEffectOnUpdate(() => {
  // ...
});
```

chardos: > should timeoutHandle be part of the array `[finalIsOpen]` as it is only a state use inside the effect lambda?

We only want to the lambda to run when the state of the accordion is changed, e.g. `finalIsOpen`. The timeout handler will indeed change with the accordion state, but adding it to the array is going to trigger a recursive loop.

PR #630, opened 2019-06-03
feat(Header): Remove notification bell

chardos:
This PR removes the notification bell (pictured below) from the Header. The bell which links to the notifications page, was an experiment that has been run for the last 3 weeks, and now will be removed while product decisions are made about the next steps moving forward.

image


react/Header/UserAccount/UserAccount.js
```diff
@@ -107,11 +106,6 @@ export default class UserAccount extends Component {
         </div>
 
         <div className={styles.accountContainer}>
```
georgespyropoulos: This div was added in https://github.com/seek-oss/seek-style-guide/pull/621/files. Should we remove it in this PR? There are also some styling changes in #621 that have not been reversed. Any reason why not?

chardos:
> The bell icon was added in #621. Should we remove it in this PR?

Yep. Good catch.

react/Header/UserAccount/UserAccount.js
chardos: Have removed `accountContainer` div. Styling changes were improvements that were made that don't need to be removed.

react/Header/UserAccount/UserAccount.less
```diff
@@ -107,12 +107,12 @@
   text-overflow: ellipsis;
   display: inline-block;
   white-space: nowrap;
-  max-width: 90px; // ~ 10 characters
+  max-width: 120px; // ~ 12 characters
```
georgespyropoulos: These changes don't quite match what we had in #621. Also there are a bunch of other changes in this file in #621. Any reason why we shouldn't revert them all?

chardos: Yeap, that values are different because the bell is now gone, we can afford to allow a larger space for the names.

PR #631, opened 2019-07-15
fix: Update lodash to prevent security vulnerability

tgdev:
Snyk has reported the following as a high severity security issue:

### Detailed paths and remediation
**Introduced through:** adv-shopfront-ui-sku@0.1.0 › seek-style-guide@40.2.0 › lodash@4.17.11
**Remediation:** Open PR to patch lodash@4.17.11.

**Overview**
lodash is a modern JavaScript utility library delivering modularity, performance, & extras.

Affected versions of this package are vulnerable to Prototype Pollution. The function defaultsDeep could be tricked into adding or modifying properties of Object.prototype using a constructor payload.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   "dependencies": {
     "basekick": "^3.0.0",
     "classnames": "^2.2.5",
-    "lodash": "^4.17.4",
+    "lodash": "^4.17.11",
     "pad-left": "^2.1.0",
     "prop-types": "^15.6.0",
     "react-autowhatever": "10.1.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -58,7 +58,7 @@
   "dependencies": {
     "basekick": "^3.0.0",
     "classnames": "^2.2.5",
-    "lodash": "^4.17.4",
+    "lodash": "^4.17.11",
```
72636c:
```suggestion
    "lodash": "^4.17.14",
```

That said, is this PR necessary? Consumers can independently pin a version >4.17.11 due to the caret range.

tgdev: I figured this would fix it once rather than every consumer (who potentially isn't using lodash in their own project) having to add it to their own package.json file.

Also, if that is the recommended solution does this become a "peer dependency"?

PR #634, opened 2019-07-26
fix(Footer): Fix NZ SEEK Business link

adenj:


react/Footer/data/tools.js
```diff
@@ -115,7 +115,8 @@ export const seekSites = [
   },
   {
     name: 'Business for sale',
-    href: 'https://www.seekbusiness.co.nz/?tracking=sk:main:nz:nav:bus',
+    href:
+      'https://www.seekbusiness.com.au/?site=nz&tracking=sk%3Amain%3Anz%3Anav%3Abus',
```
georgespyropoulos: Should line 112 (AU link) also be encoded?

adenj: Possibly. I've tried to only change what I've been requested to 😅

PR #646, opened 2019-10-29
fix(TextField): Update aria attribute

tgdev:
The attribute `aria-describedby` is currently set to an invalid value of `${id}-message` so any performance tool like webpagetest is calling this out.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute)

"The aria-describedby attribute is used to indicate the IDs of the elements that describe the object. It is used to establish a relationship between widgets or groups and text that described them."

Screen Shot 2019-10-29 at 4 31 45 pm

**_UPDATE: Conditionally render the aria attribute if/when the message prop is passed to the TextField component_**

### DIFF
--- a/react/Autosuggest/Autosuggest.js
+++ b/react/Autosuggest/Autosuggest.js
@@ -61,7 +61,7 @@ export default class Autosuggest extends Component {
     areSuggestionsShown: false
   };
 
-  componentWillUpdate(nextProps, nextState) {
+  UNSAFE_componentWillUpdate(nextProps, nextState) {
     if (
       !this.state.areSuggestionsShown &&
       nextState.areSuggestionsShown &&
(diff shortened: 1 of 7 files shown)


tgdev:
> In it's current implementation this PR breaks the relationship between the `TextField` and it's `FieldMessage`. I think we need to dig deeper on the error as the value seems valid to me. Then we can determine the best path forward.

Doesn't the input's id and label's labelfor provide a native relationship?

michaeltaranto:
Correct, what you have described sets up the relationship for the labelling the field. The `aria-describedby` establishes a relationship with the field message below the field to provide additional context to a screen reader, e.g. announcing error messages etc.

So rather than changing the value, the fix might be more to do with conditionally applying the `aria-describedby` prop only when there is a message (and hence a dom element with that id) to reference.

tgdev:
So according to the WAI-ARIA docs the `aria-describedby`:

"Identifies the element (or elements) that describes the object."
https://www.w3.org/TR/wai-aria-1.1/#aria-describedby

It takes a value of an ID reference list
https://www.w3.org/TR/wai-aria-1.1/#valuetype_idref_list

**ID reference**
Reference to the ID of another element in the same document

**ID reference list**
A list of one or more ID references.

So perhaps the {id}-message on the input doesn't match any other element on the page.
I will look at the Field component and see if it has a aria-labeledby value which should be the same (ie: {id}-message)

michaeltaranto:
Closing in favour of consumers migrating to the `TextField` from [Braid](https://seek-oss.github.io/braid-design-system/components/TextField).

PR #663, opened 2020-04-13
fix(deps): Support react-helmet 6.x

etaoins:
The named export was added in `react-helmet` 5 and the default export was removed in `react-helmet` 6.

`react-helmet` 5 was released in 2017 so most people should already have matching versions.

BREAKING CHANGE: Requires `react-helmet` >= 5.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
   "peerDependencies": {
     "react": "^16.3.0",
     "react-dom": "^16.3.0",
-    "react-helmet": "^3.0.0 || ^4.0.0 || ^5.0.0"
+    "react-helmet": "^5.0.0 || ^6.0.0"
   },
   "devDependencies": {
     "@commitlint/cli": "^6.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -68,7 +68,7 @@
   "peerDependencies": {
     "react": "^16.3.0",
     "react-dom": "^16.3.0",
-    "react-helmet": "^3.0.0 || ^4.0.0 || ^5.0.0"
+    "react-helmet": "^5.0.0 || ^6.0.0"
```
tgdev, line 71:
```suggestion
    "react-helmet": ">=^5.0.0"
```

etaoins: How do we know this will work with `react-helmet` 7?

tgdev: We update to `>=6` or `>=7`? either way you need to update.

etaoins: I mean, how do we know the existing `seek-style-guide` code will work with a future version of `react-helmet`? Listing major versions indicates which versions of `react-table` we‘re known to work with.

PR #668, opened 2020-07-28
feat(Header): Update sign out urls

trevorgk:
SEEK candidate site is in the process of migrating to Auth0 from Layer7. During the migration period (approximately 6 months) the user may be signed in to zero, one or both providers. The signout button in the SEEK style guide currently only signs the user out of Layer7.

This change updates the signout links to take the user to the new `/oauth/logout` route. This route will sign the user out of Auth0 before redirecting the user to the legacy route which will sign them out of Layer7.

### DIFF
--- a/react/Header/UserAccountMenu/UserAccountMenu.js
+++ b/react/Header/UserAccountMenu/UserAccountMenu.js
@@ -279,9 +279,7 @@ export default ({
                   'data-analytics': 'header:sign-out',
                   className: styles.item,
                   onClick: clearLocalStorage,
-                  href: returnUrl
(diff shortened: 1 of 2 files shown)


willhackett:
Don't merge yet as this will expose an outfra domain.

georgespyropoulos:
Not sure if you're aware but the header/footer have recently been rebuilt with braid and have moved over here https://github.com/SEEK-Jobs/metropolis/tree/master/packages/anz-candidate-header-footer. This version is now deprecated

markdalgleish:
Sorry for the confusion on this. Just opened a PR to remove the Header and Footer components and add clearer deprecation notices.

